PR #25, opened 2014-05-29
Speed up tests using Spork gem that keeps Rails env loaded

pjanik:
Currently when I try to run a single test, it is pretty time consuming. I guess that mostly due to warm-up time related to Rails environment loading. It's especially visible when I try to run a single test directly from text editor (e.g. using [this nice plugin](https://github.com/maltize/sublime-text-2-ruby-tests)).

This pull request tries to solve this issue using [spork](https://github.com/sporkrb/spork) gem.

I'm not pushing this into master, as I'm not sure if other developers like it (maybe there are some drawbacks of this approach) and whether my division of RSpec configuration makes sense (however it seems to work). 

I've also noticed that Parker tried to do the same thing in [spork branch](https://github.com/concord-consortium/lara/tree/spork) some time ago. This branch has never been merged, so I don't know if he just couldn't force it to work and gave up or I miss something important. After quick look at his changes, I can see RSpec config that wasn't working for me (I had to move some lines to `Spork.each_run`), so maybe that was the reason.

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,1 +1,2 @@
 --colour
+--drb
--- a/Gemfile
+++ b/Gemfile
@@ -89,6 +89,9 @@ group :test, :development do
   gem 'guard-jasmine'
   gem 'jasminerice', :git => "git://github.com/bradphelan/jasminerice.git" # guard-jasmine uses this
   gem 'jasmine-jquery-rails'
+  gem 'spork-rails'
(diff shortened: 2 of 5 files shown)


pjanik:
Well, it took a while, but there is a new pull request which addresses this issue:
https://github.com/concord-consortium/lara/pull/30

PR #30, opened 2015-01-28
Upgrade to Rspec 3.x, use Spring gem to speedup test execution

pjanik:
Long time ago I tried to speedup tests execution using Spork: https://github.com/concord-consortium/lara/pull/25
However @knowuh suggested using Spring which was included in Rails 4.x by default (what made a lot of sense).

I tried a few times, but without success, so I gave up. However finally I ran out of patience - a single test needed 10-15 sec to start... It was really affecting my productivity and approach to testing, so tried to fix it. My blind guess was to update Rspec and it fortunately worked.

Rspec 3.x has completely different syntax than 2.x though. This pull request mainly includes lots of changes in `spec` dir to handle Rspec 3.x syntax. I used https://github.com/yujinakayama/transpec tool to convert vast majority of our tests.

There was one issue with `user_spec` test that I can't explain (old `should` syntax worked, but new `expect.to` was failing...). But I was able to fix it somehow. It can be related to this test setup itself.

I also replaced dead CanCan gem with CanCanCan (community fork, still maintained) to avoid some warnings related to Rspec 3.x.

And obviously I added Spring gem which works now! Finally I can run single test and get results almost immediately.

@scytacki, @knowuh, any reason not to merge it? We have to use totally new syntax in tests, but I think it's only beneficial for us (the old one is depreciated).


scytacki:
I think you should merge it.

pjmorse:
FYI, if you ever have to do the rspec 2->3 migration again: there is an rspec 2.99 version which exists to run your 2.x-syntax tests and provide a list of changes needed to make the jump to 3.x. Might be helpful in places where Transpec doesn't come through.

Also, related to CanCanCan: we've been opting to move applications over to [Pundit](https://github.com/elabs/pundit), which is a policy-based authorization system. The one which migrated so far was a pretty heavy lift, though, and the first step was ensuring solid test coverage of authorization rules (not sure if LARA still/already has this or not).

pjanik:
Right, I used 2.99 version. Basically I was following this guide:
https://www.relishapp.com/rspec/rspec-rails/docs/upgrade
I had to run Transpec one more time (I guess after I upgraded RSpec to 3.x) to convert `stub` to `allow`, but apart from that it was pretty straightforward.

I switched to CanCanCan only to avoid test warnings, but if we work on authorization some day, I'll know to look at Pundit too. ;)

Thanks!

PR #31, opened 2015-02-06
C-Rater integration

pjanik:
@scytacki, @knowuh, I'm creating a pull request so you can take a look at this code.
It's pretty big, so summary:

- All C-Rater-specific functionality is provided in CRater module.
- I'm quite happy about core functionality related to C-Rater like its API wrapper, FeedbackFunctionality, FeedbackItem, FeedbackSettings etc. To enable C-Rater feedback it's enough to include some mixins. Tests coverage is hopefully quite good.
- I'm less happy about argumentation block and changes to InteractivePage model. I've added "additional section" concept to avoid adding yet another `show_<something>` column to this model and hardcoding things in views. Although it includes some weird meta-programming and this concept pretends to be generic, while it clearly addresses argumentation block needs. 
- Filtering of embeddables by section is also pretty specific. I've only added `section` column to PageItem model. Obviously section could be modelled better (probably it should belong to page and consist of page items etc.), but the current approach doesn't require risky migrations that restructure all activities. It's also easy to remove in the future.

Contact point between existing code and C-Rater code:
- OpenResponse and OpenResponseAnswer include CRater modules (2 lines).
- InteractivePage registers argumentation block section (1 line, but of course first I had to add "additional section" feature).
- A few new routes.

I didn't modify existing code too much, but mostly tried to provide new one that is clearly scoped to C-Rater. Unfortunately this new code code is spread across the whole codebase (models, controllers, views, helpers, lib), but should be easy to identify it due to CRater module and/or directories.

### DIFF
--- a/app/controllers/c_rater/argumentation_blocks_controller.rb
+++ b/app/controllers/c_rater/argumentation_blocks_controller.rb
@@ -1,0 +1,51 @@
+class CRater::ArgumentationBlocksController < ApplicationController
+  before_filter :set_page_and_authorize
+
+  def create_embeddables
+    create_arg_block_embeddables(@page)
+    redirect_to(:back)
+  end
+
+  def remove_embeddables
+    @page.page_items.where(section: CRater::ARG_SECTION_NAME).destroy_all
+    redirect_to(:back)
+  end
+
+  private
+
+  def set_page_and_authorize
+    @page = InteractivePage.find(params[:page_id])
+    @activity = @page.lightweight_activity
+    authorize! :update, @page
+    update_activity_changed_by
+  end
+
+  def create_arg_block_embeddables(page)
+    mc1 = Embeddable::MultipleChoice.create(custom: true)
+    mc1.create_default_choices
+    page.add_embeddable(mc1, 0, CRater::ARG_SECTION_NAME)
+
+    or1 = Embeddable::OpenResponse.create(prompt: 'Explain your answer.')
+    page.add_embeddable(or1, 1, CRater::ARG_SECTION_NAME)
+
+    or1_c_rater_settings = CRater::Settings.new(item_id: 'HENRY001')
+    or1_c_rater_settings.provider = or1
+    or1_c_rater_settings.save!
+
+    mc2 = Embeddable::MultipleChoice.create(prompt: 'How certain are you about your claim based on your explanation?',
+                                            show_as_menu: true)
+    mc2.add_choice('(1) Not at all certain')
+    mc2.add_choice('(2)')
+    mc2.add_choice('(3)')
+    mc2.add_choice('(4)')
+    mc2.add_choice('(5) Very certain')
+    page.add_embeddable(mc2, 2, CRater::ARG_SECTION_NAME)
+
+    or2 = Embeddable::OpenResponse.create(prompt: 'Explain what influenced your certainty rating.')
+    page.add_embeddable(or2, 3, CRater::ARG_SECTION_NAME)
+
+    or2_c_rater_settings = CRater::Settings.create(item_id: 'HENRY001')
+    or2_c_rater_settings.provider = or2
+    or2_c_rater_settings.save!
+  end
+end
--- a/app/views/interactive_pages/edit.html.haml
+++ b/app/views/interactive_pages/edit.html.haml
@@ -45,22 +45,27 @@
     %span.preview
       = link_to "Preview", preview_activity_page_path(@activity, @page), :target => 'new'
   %div.edit_interactive_page
-    = form_for @page, :remote => true do |f|
+    = form_for @page do |f|
       = f.select :layout, options_for_select(InteractivePage::LAYOUT_OPTIONS.map { |l| [l[:name], l[:class_val]] }, @page.layout), {},  { :onchange => 'this.form.submit();', :title => 'Page layout' }
       = f.select :embeddable_display_mode, InteractivePage::EMBEDDABLE_DISPLAY_OPTIONS, {}, { :onchange => 'this.form.submit();', :title => 'Display of embeddables: stacked for scrolling, or carousel for sequential display' }
-      %span.page-section-controls#page-section-controls     
-        %label{:id => 'page_intro',:for => "interactive_page_show_introduction"}    
+      %p.page-section-controls#page-section-controls
+        %label
           = f.check_box :show_introduction, :onchange => 'this.form.submit();' 
           Page introduction
-        %label{:id => 'info_block',:for => "interactive_page_show_info_assessment"}
+        %label
           = f.check_box :show_info_assessment, :onchange => 'this.form.submit();'
           Info/Assessment block
-        %label{:id => 'interactive_box',:for => "interactive_page_show_interactive"}
+        %label
           = f.check_box :show_interactive, :onchange => 'this.form.submit();'
           Interactive box
-        %label{:id => 'page_sidebar',:for => "interactive_page_show_sidebar"}
+        %label
           = f.check_box :show_sidebar, :onchange => 'this.form.submit();'
-          Page sidebar text  
+          Page sidebar text
+        - InteractivePage.registered_additional_sections.each do |s|
+          %label
+            = f.check_box "show_#{s[:name]}", :onchange => 'this.form.submit();'
+            = s[:label]
+
   - if @page.show_introduction
     .instructions{:class => @page.text.blank? ? 'block-empty' : '' }
       %h2 Introduction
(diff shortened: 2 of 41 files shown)


app/controllers/c_rater/argumentation_blocks_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+    mc2.add_choice('(2)')
+    mc2.add_choice('(3)')
+    mc2.add_choice('(4)')
+    mc2.add_choice('(5) Very certain')
+    page.add_embeddable(mc2, 2, CRater::ARG_SECTION_NAME)
+
+    or2 = Embeddable::OpenResponse.create(prompt: 'Explain what influenced your certainty rating.')
+    page.add_embeddable(or2, 3, CRater::ARG_SECTION_NAME)
+
+    or2_c_rater_settings = CRater::Settings.create(item_id: 'HENRY001')
```
scytacki: is there a reason `create` is used here and `new` is used above?

pjanik: No, it should be `new` too. Perhaps I forgot to change it here. I'll fix it, thanks.

app/views/interactive_pages/edit.html.haml
```diff
@@ -45,22 +45,27 @@
...
           = f.check_box :show_interactive, :onchange => 'this.form.submit();'
           Interactive box
-        %label{:id => 'page_sidebar',:for => "interactive_page_show_sidebar"}
+        %label
           = f.check_box :show_sidebar, :onchange => 'this.form.submit();'
-          Page sidebar text  
+          Page sidebar text
+        - InteractivePage.registered_additional_sections.each do |s|
+          %label
+            = f.check_box "show_#{s[:name]}", :onchange => 'this.form.submit();'
```
scytacki: Is this the only usage of the dynamic attributes added to InteractivePage? 
If so, there might be a different approach here that doesn't require the dynamic attributes. Something that works directly with the additional_sections attribute.

db/schema.rb
```diff
@@ -33,6 +33,39 @@
...
+
+  add_index "c_rater_feedback_items", ["answer_id", "answer_type"], :name => "c_rat_feed_it_answer_idx"
+
+  create_table "c_rater_score_mappings", :force => true do |t|
+    t.text     "mapping"
+    t.datetime "created_at", :null => false
+    t.datetime "updated_at", :null => false
+  end
+
+  create_table "c_rater_settings", :force => true do |t|
```
scytacki: I think it would be better for this model to have a name like c_rater_item (or replace 'item' with whatever the c-rater folks call their items). Or c_rater_item_settings. The use of just c_rater_settings made me think these were settings for the global interaction with c_rater.

db/schema.rb
```diff
@@ -33,6 +33,39 @@
   add_index "authentications", ["uid", "provider"], :name => "index_authentications_on_uid_and_provider", :unique => true
   add_index "authentications", ["user_id", "provider"], :name => "index_authentications_on_user_id_and_provider", :unique => true
 
+  create_table "c_rater_feedback_items", :force => true do |t|
```
scytacki: Because we are going to need to record the answer and feedback given to students on the multiple choice questions. in the argumentation block. It might make sense to make this model not be c_rater specific. It looks like the only c_rater specific field here is the 'item_id', and that could be indirectly found by going through the answer object.

scytacki:
This looks good. I think it is a good compromise between time and maintainability.
I like the additional_sections approach. It would be better if the built in sections used this approach too, but I don't think there is time to make that change. Perhaps the additional_sections support could be more simple because it looks like the dynamic `show_` attribute is only used by the view that renders the checkboxes.

I think adding some documentation about the contact points would be useful. Probably it will go out of date, but it might provide a clue none the less. To this documentation you should also include the reverse contact points: what attributes and methods on the LARA core models does the c_rater code need? 
what partials does the c_rater code expect the LARA core code to have? I saw that it needed there to be a `[embeddabel_type]s/author` partial. And it also expects those embeddable partials work when they are embedded anywhere in the page.

It also seems like this would be good to put in an engine. That will keep all of the code in a single place. However, I don't think there is time to do that. It will probably be better to discuss that option when we come back to this code which will probably be in a couple of months.

db/schema.rb
pjanik: Agree, so I'd use `c_rater_item_settings`. ETS folks use `item` world too.

db/schema.rb
pjanik: Hmm, I was thinking about too. But there is also another option.

We will have to create a new model related to the argumentation block submission. It will obviously reference C-Rater feedback items of the open response questions. When it comes to multiple choice answer and feedback, there are two options:

1) we could create generic feedback item model as per your suggestion
2) we could just store multiple choice answer and feedback explicitly in this new model (so simply add a few columns there, like `main_question_answer`, `main_question_feedback`, etc.)

Separate model makes sense for C-Rater, as it can be useful for debugging. Each feedback item corresponds to one HTTP request and HTTP metadata is also stored there. 

When it comes to multiple choice question feedback, the whole operation would be as simple as copying current answer text and feedback text. It's not too risky and additional feedback item model doesn't bring too many benefits. Except from the fact that it would be similar to C-Rater feedback approach. We also don't need all the columns (status, score, response_info etc.), so we would need to either make them more generic or use another db table.

app/views/interactive_pages/edit.html.haml
pjanik: Yes, only checkboxes are using dynamic `show_` methods. Do you want me to remove them?
I assumed that they will simplify view and controller code (update method), but yes, it can be implemented without them too.

scytacki: You might as well leave them for now. 

Do you think you will use them in the runtime code? 
I suspect there will be a `each` loop in the runtime code too. So then there will probably not be any direct usage of the `show_` methods.  If they aren't ever used directly then I don't think they are necessary. But that depends on how complex the code above would become without them.

db/schema.rb
scytacki: I think option #2 is fine as well.
I think it is also ok to wait on this until we have the runtime code done. At that point it might be more obvious if there is any benefit to unifying the feedback models. 

In my chat with Trudi, it seems the gating and 'feedback on feedback' will be more important, than the researcher report for this current block of work, so we might not need to worry about any of this  for a few months.

pjanik: Okay. I'll see, I feel I'll start with 2 then.

As we need to store somewhere 'feedback on feedback' and mentioned multiple choice feedback, this linking model will be necessary anyway. In fact this model will represent everything we need for the researcher report too. Remaining work would be Portal-LARA communication and .xls file generation.

app/views/interactive_pages/edit.html.haml
pjanik: You're right, runtime code also uses `each` loop.
I agree it would be nice to remove dynamic method generation.
However as it just works, I would do that when we have MVP ready.
Tomorrow I'll try to add PT stories for all the things that you listed in comments (or fix them right away).

db/schema.rb
pjanik: Small update on that. In fact I'm implementing the first option finally.

The problem is about displaying old feedback when user visits page again. I was thinking that for multiple choice answer we could always get feedback dynamically. But... it creates problems when user first selects answer "A", clicks "Submit", then changes answer to "B" and reloads the page.

I assume that we always want to display feedback that was explicitly created by submit action (so feedback on "A") or nothing if answer has been changed in the meantime. I bet we don't want "silently" update feedback to match "B" answer. So we need explicit storage for feedback. Using linking model for that described in 2 would be messier.

pjanik: FYI: Changed in https://github.com/concord-consortium/lara/commit/3f1d919d1a1b442963c9621559e116d157778d84

I was planning to implement multiple table inheritance using this gem:
https://github.com/hzamani/active_record-acts_as
as it looked simple enough. However I didn't notice that it required Rails 4.x. We could use deprecated version of this gem (ver 0.1), but probably it doesn't make sense. I haven't found reasonable alternative (yet).

So at the moment `CRater::FeedbackItem` inherits from `Embeddable::FeedbackItem`, but simply uses its own db table. I could split extra fields into separate table and create explicit association, but I'm not sure if it's worth it.

app/controllers/c_rater/argumentation_blocks_controller.rb
pjanik: Fixed in: https://github.com/concord-consortium/lara/commit/99901dd77351a2963a1110e25b753d6a0076b7c8#diff-35bd40ff1f7e62acd14c3d895a7a9726R64

db/schema.rb
pjanik: Fixed: https://github.com/concord-consortium/lara/commit/502b51923ae4dd2728d3a59bbc82dc139b5e5b41

pjanik:
First, thanks for the comprehensive review! We should do it more often.

Hopefully all the comments are resolved (see line notes) or relevant PT stories are created:
- docs: https://www.pivotaltracker.com/story/show/88052708
- dynamic methods: https://www.pivotaltracker.com/story/show/88053160

Should we go ahead and merge it?

db/schema.rb
scytacki: Yeah I don't think it is worth it.  Perhaps we can updated to Rails 4.x sometime soon. :)

scytacki:
Yes go ahead and merge it.

PR #32, opened 2015-02-11
C-Rater runtime

pjanik:
Summary:

- Basic UI (no HASBot etc.)
- Gating implemented (c_rater.coffee)
- "Submit" button still causes page reload
- `save-on-change.coffee` had to be changed slightly (we should double-check if required questions still work fine during testing)
- Some changes are related to previous C-Rater pull request comments:
  - rename `CRater::Settings` to `CRater::ItemSettings`
  - add generic `Embeddable::FeedbackItem` used by multiple choice questions


app/views/c_rater/argumentation_block/_runtime.html.haml
```diff
@@ -0,0 +1,87 @@
...
+        - fb_visible = fb && !fb.feedback_text.blank?
+        - fb_dirty = fb && fb.outdated?
+        - fb_text = fb && fb.feedback_text
+        - fb_score = fb && fb.score
+        .ab-feedback{:id => "feedback_on_answer_#{m.id}",
+                     :style => fb_visible ? '' : 'display: none',
+                     :class => "ab-score#{fb_score}",
+                     'data-dirty' => fb_dirty ? 'true' : 'false'}
+          .ab-dirty{:style => fb_dirty ? '' : 'display: none'}
+            Answer has been changed, please resubmit to update feedback!
```
scytacki: The user visible strings in this file should be moved into config/locales/en.yml

app/assets/javascripts/c_rater.coffee
```diff
@@ -0,0 +1,141 @@
...
+  updateQuestion: (id, answered) ->
+    q = @question[id]
+    # Undefined means that this question isn't part of the argumentation block.
+    return unless q
+    q.answered = answered
+    q.dirty = q.data != $(q.formElement).serialize()
+
+  submitButtonClicked: (e) ->
+    unless @allQuestionAnswered()
+      return modalDialog(false, 'Please answer all questions in the argumentation block.')
```
scytacki: The user visible strings in this file should be moved into the en.yml file.  I'm not sure how this is done in javascript files, but I'm pretty sure there is example of this in the LARA code somewhere.

scytacki:
@pjanik if it makes sense to you, it if fine with me if me if we merge this into master, and then fix things there. Or you can keep this open if that makes it easier to review the Zeus changes.

pjanik:
@scytacki, I've left pull request open. It's fine, as we are deploying this branch to staging anyway.
All the comments seem to be related to i18n. I've fixed that in https://github.com/concord-consortium/lara/commit/64ad9b9fe679a267fd9100b5cb80f88aaf1a96d7, moved a few more strings there so we can translate the whole argumentation block in case of need.

Obviously I saw i18n file before, but I had an impression that it's not really used in the recent development and it looked pretty smallish (comparing to number all user facing strings we have in LARA probably). But obviously using it won't hurt. I assume that we mostly care about student visible strings and we are less / not concerned about authoring part (all the buttons there, forms, etc.)?

scytacki:
Yeah we only care about student visible strings. We've probably have been neglecting it, but we should update it with any new work we do.

PR #33, opened 2015-02-24
Support copy, export, and import for arg block.

sagartalla:
[#88240746]

- An author/admin can copy, export, and import an activity with an
 arg block.

### DIFF
--- a/app/models/c_rater/settings_provider_functionality.rb
+++ b/app/models/c_rater/settings_provider_functionality.rb
@@ -8,33 +8,33 @@ module CRater::SettingsProviderFunctionality
     base.alias_method_chain :export, :c_rater
     
     class << base
-    	def import_with_c_rater(import_hash)
-    		import_embeddable = self.import_without_c_rater(import_hash.except(:item_settings))
-		 		if import_hash[:item_settings]
-		      item_settings = CRater::ItemSettings.import(import_hash[:item_settings])
-		      item_settings.provider = import_embeddable
-		      item_settings.save!(validate: false)
-		    end
-		    import_embeddable
-    	end
-    	alias_method_chain :import, :c_rater
-  	end
-  	
+      def import_with_c_rater(import_hash)
+        import_embeddable = self.import_without_c_rater(import_hash.except(:item_settings))
+        if import_hash[:item_settings]
+          item_settings = CRater::ItemSettings.import(import_hash[:item_settings])
+          item_settings.provider = import_embeddable
+          item_settings.save!(validate: false)
+        end
+        import_embeddable
+      end
+      alias_method_chain :import, :c_rater
+    end
+    
   end
   
   def duplicate_with_c_rater
-  	copy = duplicate_without_c_rater
-  	if self.c_rater_item_settings
-  		item_settings = self.c_rater_item_settings.duplicate
-    	item_settings.provider = copy
-    	item_settings.save!(validate: false)
-  	end
+    copy = duplicate_without_c_rater
+    if self.c_rater_item_settings
+      item_settings = self.c_rater_item_settings.duplicate
+      item_settings.provider = copy
+      item_settings.save!(validate: false)
+    end
     copy
   end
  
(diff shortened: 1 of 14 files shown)


scytacki:
@pjanik this code is adding references to c_rater_item_settings in interactive_page.  Can you think of a way to isolate this code more? 
Perhaps it should go here: https://github.com/concord-consortium/lara/blob/master/app/models/c_rater/settings_provider_functionality.rb

pjanik:
Yes, it's the best place probably. It could be nice to somehow overwrite `export` method of a class that includes `SettingsProvider` module. It may be non-trivial though, as it's only module, not a subclass. However it's possible:
http://stackoverflow.com/questions/5944278/overriding-method-by-another-defined-in-module
There may be a better approach, that's just my first thought.

sagartalla:
@scytacki @pjanik Should we go ahead and move it to `SettingsProvider`.

pjanik:
@sagartalla, is it a question? If so, then yes, I would move it to `SettingsProvider`. 
If we were using Ruby 2.x (is it problematic?), then we could use `module#prepend` method to extend `export` method.

Otherwise I guess you can:
- separate export code to a new module that is included before `SettingsProvider` (so you can access original export using `super`)
- or use `alias_method_chain`

scytacki:
@pjanik what do you think about adding an 'extensions' list in OpenResponse, then the export and import code can let each extension "extend" the exported json. And on import, each extension is passed the imported json.  This might make it easier to see what is going on, instead of a magical overriding of the export and import methods.  When the `CRater::SettingsProviderFunctionality` is imported it would add itself as an extension to OpenResponse.

pjanik:
Hm, I feel it's a matter of taste. Personally I wouldn't workaround more advanced feature of programming language because there is a bit of magic in it. I feel like `prepend` and `alias_method_chain` (Rails syntax sugar over `alias_method`) have been designed to solve cases like this one. Of course it's possible to overuse them, I agree. 

The biggest benefit for me is that when I read or change OpenResponse code, I don't have to think what "extension" is. And if we removed C-Rater integration, there would be no leftovers.

I think I would definitely prefer extension approach if it was more common pattern in our code. For example if we had a few different extensions of OpenResponse (so obviously chain of `super` calls or aliased methods would become difficult to follow) or extensions of different question types.

I'm fine with both solutions though.

scytacki:
OK, lets go with the alias_method_chain approach.
It will require extra work to switch to Ruby 2.x, so it shouldn't be part of this work.

sagartalla:
@scytacki @pjanik Please let us know if this works for you.

app/models/embeddable/open_response.rb
```diff
@@ -51,5 +51,6 @@ def export
     def self.import (import_hash)
       return self.new(import_hash)
     end
+    include CRater::SettingsProviderFunctionality
```
scytacki: Please add a comment above this line explaining why it needs to be included after the methods are defined.

scytacki:
This looks fine to me, the only change I want is the comment above the `include`.

app/models/c_rater/settings_provider_functionality.rb
```diff
@@ -1,6 +1,42 @@
 module CRater::SettingsProviderFunctionality
   extend ActiveSupport::Concern
-  included do
+  included do |base|
     has_one :c_rater_item_settings, as: :provider, class_name: 'CRater::ItemSettings'
+    
+    base.alias_method_chain :duplicate, :c_rater
+    
+    base.alias_method_chain :export, :c_rater
```
pjanik: Is `base.` needed here? If it's not, I would remove it. Perhaps I doesn't hurt, but when I'm reading this code, I'm simply wondering whether it needs to be there or not. And it would be consistent with `has_one` above.

pjanik:
It looks fine to me too, I left one comment regarding a tiny thing.

sagartalla:
@scytacki @pjanik Done, Thanks.

PR #34, opened 2015-02-25
Single page rendering

pjanik:


app/assets/javascripts/iframe-saver.coffee
```diff
@@ -82,30 +75,39 @@ class IFrameSaver
...
-    @iframePhone.post({ type:'getInteractiveState' })
+    @iframePhone.post({type: 'getInteractiveState'})
 
   confirm_delete: (callback) ->
     if (window.confirm("Are you sure you want to restart your work in this model?"))
       callback()
 
   delete_data: () ->
+    # Disable autosave, as it's possible that autosave will be triggered *after* we send to server "null" state
+    # (delete it). Actually it used to happen quite ofted.
```
knowuh: Nice catch.

app/assets/javascripts/sidebar.js
```diff
@@ -0,0 +1,20 @@
+$(function () {
+  // Dynamically setup position of sidebar handles.
+  function positionMultipleSidebars() {
+    var sidebarTop = $('.sidebar-mod').offset().top;
+    var sidebarHdrHeight = $('.sidebar-hdr').height() + 35;
```
knowuh: Magic number 35 spotted.

knowuh:
@pjanik I reviewed the changes, and they look good.  Do you want someone else to look at it too? just at mention them if you do. Happy to look again later if you want too.

app/assets/javascripts/iframe-saver.coffee
pjanik: We don't have to re-enable autosave, as we reload the page at the end. However I'm using the same function to enable autosave in the constructor. Not sure if that's what you meant.

Re auto-save as an edge case - what about the last page of an activity, where there is no "next" button that forces Iframe saving? I asked about that through Github some time ago.
Single page has the same issues as the last page, so that's why I've added couple of additional autosave scenarios (focus, mouseout and changed 42 seconds to 5).

app/assets/javascripts/sidebar.js
pjanik: Thanks, hopefully improved in https://github.com/concord-consortium/lara/commit/37cb47c448687300cb8cec63d3b67fbabbe320d3.

app/assets/javascripts/image_question_drawing_tool.js.coffee
```diff
@@ -74,6 +73,10 @@ class ImageQuestionDrawingTool
     @$annotation_field          = $("#{@form_sel} [name=\"#{@form_prefix}[annotation]\"]")
     @$answer_text_field         = $("#{@form_sel} [name=\"#{@form_prefix}[answer_text]\"]")
 
+    # Find the first interactive which is on the same page (aka content-module, div.content-mod) as the question.
+    # It's necessary when we are in single page rendering mode and there are multiple interactives rendered.
+    @interactive_element = @$main_form.closest('.content-mod').find('.interactive-mod > *:first-child')[0]
```
scytacki: Instead of finding the interactive like this. Could the ruby code determine the id of interactive on the page and pass that that into here?
We hopped at some point to support multiple interactives on a page even in 'multiple page' layout. In that case the author would need to explicitly select which interactive an image question is connected to. To move in that direction, the ImageQuestion rails model could have a method which returns the the first interactive that is on the same page. And that would be eventually be passed passed into this code in some form.

Then when we really add support for multiple interactives on a single page (in multiple page layout), we would just need to change the ImageQuestion rails model and this code could remain the same.

app/assets/javascripts/logger.js
```diff
@@ -69,21 +69,24 @@ LoggerUtils.prototype._submittedQuestionLogging = function(data) {
 
 LoggerUtils.prototype._interactiveSimulationLogging = function() {
   // work around for bindig click event on cross site iframe
-  var self  = this,
-  myConfObj = {iframeMouseOver: false};
+  // FIXME: this code no longer works, as #interactive has been removed (now .interactive class is available instead
+  //        and we should assume that there can be multiple interactives on a single page).
```
scytacki: It looks like this is fixed now. Can this comment be removed?

app/assets/javascripts/image_question_drawing_tool.js.coffee
pjanik: Makes sense to me, updated: https://github.com/concord-consortium/lara/commit/9f22885ae19fa124f9e50e0c79f4ba2df80a7709
I also added very simple handling of one edge case (alert when interactive is not defined).

app/assets/javascripts/logger.js
pjanik: Yes, it seems it's been fixed, comment removed in https://github.com/concord-consortium/lara/commit/32d9638a6f125c561a20f36c8f7320ca180578e7.

scytacki:
Since the two things I found have been fixed. I'd say lets merge this into master.

PR #36, opened 2015-03-20
Lab logging

pjanik:
@rklancer, I went ahead and created pull request so we have place to leave some comments.


pjanik:
[ I haven't tried to run this branch or test it, I'm just commenting the code looking at the diff, so I can miss something ]

I think there can be a problem with this approach - `IframeSaver` is instantiated only for iframes / interactives that have "save state" feature enabled (== iframe has `savable` class, probably it should be named `saveable`). I guess we want to log events from all the interactives, not only ones that save their state.

And in general I feel it would be nice not to mix logging with `IframeSaver`. 
What about creating an iframe phone for each iframe independently from `IframeSaver` and logging code? Then both `IframeSaver` instance and logging code (and any other code that needs to talk to iframes) could request a phone connected to the particular iframe. For example phones could be available globally in some module and identified by iframe ID or maybe we could even use `jQuery#data` to attach phone to iframe element itself. I guess there are more available options.

knowuh:
Right.  I have my doubts about this because of the points @pjanik brought up.

I like the idea of creating an IframePhone for each iframe, and mediating phone connections in a global scope, possibly using data attribute to identify the phone. 

Could these issues (and a few more) be fixed by spending the time to rethink & rework iFramePhone as @rklancer has suggested? Or is that likely to take a too long?  As part of that, could we imagine iFramePhone trying to auto-configure itself in the parent window, and broker & forward all child iFrame communications?

scytacki:
From what I remember iFramePhone already supports the idea of a parent window having multiple phones. One phone per iframe.  I don't think it should be iFramePhones responsibility to help the container page figure out which phone goes with which iframe.

I'd really like to keep iframephone as simple as possible. And I would like to even remove some of the complexity that is there.

scytacki:
I'll back peddle on that.  If 2 different libraries are using iFramePhone to talk to an iframe. And a container page includes both libraries, then things should still work. This should also work inside the iframe'd pages. I've thought this should work, by having iFramePhone be aware if it is included multiple times by different libraries so it can then make sure only one copy of itself is loaded. And then its public API needs to handle the case where 2 or more 'clients' are asking to get a phone connected to the same iframe.

So then it does seem like iframe-phone needs to be some what of broker.

rklancer:
Re our discussion, the comments about iframephone are in the commit message of https://github.com/concord-consortium/lara/commit/a22d1e72d58b31946a701a35b0975acc678a5bb2

PR #37, opened 2015-03-25
Action Menu: orange Bar or floating menu

scytacki:


scytacki:
@sagartalla, I was really hoping we would not need different html to support the orange bar and the floating menu. Instead those differences should be done with CSS.  Can you give that a shot?

Instead of using the class `no-left-border`, try using `:first-child` to remove the vertical line from the first item.

Another point that wasn't clear based on the story.  If the single page activity is not using the ITSI theme then the floating menu style should be used.

scytacki:
I revised this. I think it is working the way I want it now. Please test it out and see if there is something broken.

scytacki:
I refactored the css and html some more. The main change is that the summary page generates its own activity menu html instead of including a common partial. And there is a `summary-report` class added to the summary page.  This way the only change needed in the itsi theme is the showing of the close button.

I also pulled out the button styling of the menu so it is not included at all in itsi theme. This way the itsi css doesn't need to revert or override the button styling. 

Please test it and let me know if I broke anything. This is deployed to authoring-demo.concord.org

PR #38, opened 2015-04-02
Session debugging -- Dont merge!

knowuh:
Just a place to track changes.

cc:  @scytacki @psndcsrv

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -6,7 +6,12 @@ class ApplicationController < ActionController::Base
 
   # Run authorization on all actions
   # check_authorization
-  protect_from_forgery
+  # np: Rails 4 API would let us do this: http://bit.ly/1HokQ0f 
+  protect_from_forgery with: :exception
+  # ⬆ doesn't seem to have any effect in rails 3.  We still return 200.
+  # rails 3 assumption is that the controller would check the user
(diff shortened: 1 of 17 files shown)


app/controllers/application_controller.rb
```diff
@@ -6,7 +6,12 @@ class ApplicationController < ActionController::Base
 
   # Run authorization on all actions
   # check_authorization
-  protect_from_forgery
+  # np: Rails 4 API would let us do this: http://bit.ly/1HokQ0f 
+  protect_from_forgery with: :exception
+  # ⬆ doesn't seem to have any effect in rails 3.  We still return 200.
+  # rails 3 assumption is that the controller would check the user
+  # We could do eg in controllers/embeddable/open_responses_controller.rb : 
+  # http://bit.ly/1I8FxAw (github)
```
scytacki: It is possible to do this globally by adding a method to the application controller:
https://github.com/concord-consortium/lara/blob/fa72fc3f027243a8c390f9bda49c84391950a552/app/controllers/application_controller.rb#L311-L314
That method could throw an exception instead of logging and reseting the session.

scytacki:
I brought the session variable clearing code into master and staging. That is something we should be doing no mater what.

psndcsrv:
I'm going to close this out since #41 includes all the work here and was merged.

PR #39, opened 2015-04-07
Abstract iframe-phone in LARA so every interactive has an iframe-phone

pjanik:
The main change is that the client code shouldn't use iframe phone module directly, but `IframePhoneManager` instead. It ensures that there is only once instance of the iframe phone connected to given iframe (and RPC endpoint for a given namespace). Also, each client can specify its own `afterConnectedCallback`.

It doesn't solve `IframePhone` issues - it's still possible that some messages will get lost if message handlers are added too late. It just makes the code more sane - we don't store iframe phone instances in `Logger` and `IframeSaver` classes, there is less coupling between them and generally we delegate iframe-phone related tasks to a new tool.

### DIFF
--- a/app/assets/javascripts/iframe-phone-manager.coffee
+++ b/app/assets/javascripts/iframe-phone-manager.coffee
@@ -1,0 +1,62 @@
+# Manager of the iframe phone instances.
+# Do not use iframePhone module directly, use this class instead.
+
+class IframePhoneManager
+  IFRAMES_SEL = 'iframe.interactive'
+
+  @_instance = null
+
+  # PUBLIC
+  # Only class methods are designed to be public.
+  @getPhone: (iframeEl, afterConnectedCallback) ->
+    @_getInstance()._getPhone(iframeEl, afterConnectedCallback)
+
+  @getRpcEndpoint: (iframeEl, namespace) ->
+    @_getInstance()._getRpcEndpoint(iframeEl, namespace)
+
+  @_getInstance: ->
+    @_instance = new IframePhoneManager() unless @_instance
+    @_instance
+
+  # PRIVATE
+  # Instance methods are not intented to be used externally, not even constructor, they are all private.
+  constructor: ->
+    @_iframeData = {}
+    @_processAllIframes()
+
+  _getPhone: (iframeEl, afterConnectedCallback) ->
+    data = @_iframeData[iframeEl]
+    return null unless data
+
+    if afterConnectedCallback
+      if data.phoneAnswered
+        # Ensure that callback is *always* executed in an async way.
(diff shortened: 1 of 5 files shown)


scytacki:
@rklancer can you take a look at this?

pjanik:
Any comments on this?

It would be pretty useful for global interactive save and load feature, as that's yet another piece of code that is going to use iframe phone.

scytacki:
Here is a wrinkle (I think).  Pretty soon LARA will have a 'click to play' options for interactives. The way this works is that initially just an img element is on the page. When the user clicks the image element, it is replaced with an iframe element. So the manager will need a way for this 'click to play' javascript code to 'register' a new iframe. Perhaps `_getPhone` and `_getRpcEndpoint` can handle the case where there is no phone yet for the passed in element. Zeus will be adding this 'click to play' code, so we'll need some clear instructions about how they use the iframe manager in this case.  

I think it would be fine to merge this in so you can start using it, and then we can prepare it for click to play directly in master or a new pull request.  Last call for comments from others.

scytacki:
Looking at this, it seems another possible feature of iframe phone, or the manager, would be to allow other clients to listen for messages sent to the iframe.  So in this case the logging code could listen for the `loadInteractive` message, and then re-enable logging after words. This way the stateLoading code doesn't need worry about the logging.  We could add that to the list of features of iframe phone redux if you think it makes sense. It seems like it is moving us closer to what postal.js does.

pjanik:
Agree, after I implemented this, I was thinking that perhaps it could be better to dynamically create a phone when it's requested for the first time instead of doing that in the constructor. That should be easy to change though.

pjanik:
Okay, I've changed the code a bit, so now iframe phone is created dynamically when it's requested for the first time.

> So in this case the logging code could listen for the loadInteractive message, and then re-enable logging after words. This way the stateLoading code doesn't need worry about the logging.

Then the logging code would have to worry about `loadInteractive` action / message. :wink:  But if we have such feature, I agree, it sounds better / more natural to me.

The manager can't provide such functionality at the moment, as the client code talk directly to the phone. We could wrap iframe phone in some new class or subclass it and add this new functionality there, but I feel it can be better to keep it in mind when we work on iframe-phone improvements.

ddamelin:
How will this affect functionality in CODAP?

scytacki:
@ddamelin this shouldn't affect CODAP. It just a change to the internals of javascript running on the LARA page. However, we should test all of the interactives that talk to LARA through iframe phone to make sure sure they still work.

ddamelin:
Right, thanks. I was thinking that perhaps there was some talk about modifications to the iframe-phone general framework

pjanik:
No vetoes, so I'm merging this pull request.

PR #40, opened 2015-04-14
Global interactive state

pjanik:


app/assets/javascripts/global-iframe-saver.coffee
```diff
@@ -0,0 +1,54 @@
...
+  _saveGlobalState: ->
+    $.ajax
+      type: 'POST'
+      url: @_saveUrl
+      data:
+        raw_data: JSON.stringify(@_globalState)
+      success: (response) =>
+        console.log 'Global interactive save success.'
+      error: (jqxhr, status, error) =>
+        console.error 'Global interactive save failed!'
```
pjanik: Perhaps it would be nice to handle this error better, e.g. display dialog and recommend page reload or waiting a moment (we do something similar when snapshot fails).

scytacki:
@psndcsrv can you fix this to handle the new un-authorized code? Perhaps it already does, but I'm assuming not because it was added before that code was merged.

psndcsrv:
Sure, no problem!

PR #41, opened 2015-04-14
Notify users if they're accessing a Run that doesn't belong to them

psndcsrv:


psndcsrv:
@scytacki @knowuh This supercedes #38 and is ready to go unless you feel like it needs some tweaking.

knowuh:
Looks good to me.

scytacki:
Thanks Aaron. We'll get this up on staging, and hopefully Evangeline can start testing it today or tomorrow.

PR #42, opened 2015-04-20
Add support to LARA for click to play interactives.[#88907694][#92585716...

sagartalla:
...]

Add support to LARA for click to play interactives. They show a static image of the interactive and then when it is pressed the interactive is loaded.


sagartalla:
@scytacki @pjanik Please let us know if this works for you. We have added a new class `InteractiveManager` which will initialize all the interactive-related features.

app/assets/javascripts/interactive-manager.coffee
```diff
@@ -0,0 +1,34 @@
...
+    $data = $interactive.find('.interactive_data_div')
+    $delete_button = $interactive.find('.delete_interactive_data')
+    $click_to_play = $interactive.find('.click_to_play.shown')
+
+    if $click_to_play.length == 0
+      if $interactive.hasClass('savable') > 0
+        new IFrameSaver($iframe, $data, $delete_button)
+
+      if gon.globalInteractiveState?
+        window.globalIframeSaver = new GlobalIframeSaver gon.globalInteractiveState, $iframe
```
pjanik: It's not going to work.

`GlobalIframeSaver` should have only once instance that manages all the interactives. Perhaps it should have been implemented as singleton too.

You should use existing instance and just call `addNewPhone()` (the only public method there, exactly due to his reason). Probably it would be better to change `addNewPhone` into `addNewInteractive` and get phone within this method, so we don't have to do it here.

sagartalla:
@pjanik Thanks Done, We have also added `logInteractiveEvents` method in `LoggerUtils` class which gets called when an interactive is registered from `InteractiveManager`.

scytacki:
@pjanik can you take a look at this again?

pjanik:
IframeManager and GlobalIframeSaver look okay now. I'm assuming it's tested and works (I haven't checked out this branch locally).

scytacki:
We'll test it out today at the demo meeting, from authoring-demo.concord.org

pjanik:
Looks like global state support is broken in this branch:
http://authoring-demo.concord.org/activities/9

The same activity (exported and imported) works fine on staging:
http://authoring.staging.concord.org/activities/706

I think it could be useful to merge master into this branch before debugging / testing it more.

scytacki:
Richard is going to use authoring-demo for his meeting this afternoon. I'll figure out after this meeting if we can switch it back to to the click-to-play branch.

scytacki:
It seems more important for @rklancer to use authoring-demo for this work for the next couple of days. So we'll have to wait to deploy this until Monday probably. If I have time I will try it out locally, but the chances are slim.

scytacki:
I deployed this to authoring-demo, and it seems like global state is still broken. @sagartalla please take a look.
I'm not sure that authoring-demo will continue to have this branch. So to test this locally
You can export
http://authoring.staging.concord.org/activities/706
Then import it locally. When you run this activity, you should click on the pencil icon and draw on the graph. If globalState is working then the line you drew on the first graph should show up on the next graph.

You can also add this test interactive two times to an activity:
http://concord-consortium.github.io/lara-interactive-api/iframe.html
When you change the globalState and then hit the globalSaveState button, the other interactive should update.

sagartalla:
@scytacki We tested it locally using the method you mentioned above. It is working now. The change required is here https://github.com/concord-consortium/lara/commit/24c41b7b37cd6a9a14aa424ad62f73bac62e7829

logging, save state are also working fine.

scytacki:
Yup, I see it working now. OK I'm going to merge this in.

PR #43, opened 2015-04-21
Iframe phone mock

pjanik:
Iframe phone mock implements complete iframe phone interface, but ignores origin check (I don't think it's important for the client code in most cases, but if we find it useful, we can obviously add it). There is no RPC support, as I don't think it's worth it at the moment (I don't think we're going to continue using it after recent discussions).

It lets us test LARA iframe API using Jasmine (or any other JS test framework). I was also considering another approach - we could add a real iframe with a real iframe phone and then use JS API to trigger real messages (as both parent and iframe page would be hosted from the same domain, so they have access to each other's context). However that doesn't seem easy in Jasmine tests (as it would be more like an integration test, not unit) and the initial setup would be probably much more painful. I hope the full mock is more convenient (just `.install()` call, no spies necessary etc.).

I've added two specs:
- specs of mock iframe phone itself, so it's tested and we have an example of use
- specs of global interactive state JS code (real world use case)

BTW, I'm wondering if we should extract this code and add it to iframe phone repository.

### DIFF
--- a/app/assets/javascripts/iframe-phone-manager.coffee
+++ b/app/assets/javascripts/iframe-phone-manager.coffee
@@ -56,8 +56,11 @@ class IframePhoneManager
     @_phoneData[phoneId] =
       phoneAnswered: false
       phoneAnsweredCallbacks: []
-      phone: new iframePhone.ParentEndpoint iframeEl, => @_phoneAnswered(iframeEl)
       rpcEndpoints: {}
+
+    # Make sure that phone data is created before we create phone itself. It lets us support case when
+    # afterConnectedCallback is executed synchronously right away in constructor (useful for tests).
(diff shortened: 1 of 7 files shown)


pjanik:
I've updated this pull request.

The main change is that now it's possible to control when iframe phone connection is initialized.
By default it happens when mock phone is constructed, but you also do it manually in case of need:

```javascript
jasmine.mockIframePhone.autoConnect = false; // usually in beforeEach()
(...)
phone.fakeConnection(); // afterConnectedCallback is executed 
```

It turned out to be useful in some cases. Also, there are new specs using iframe phone mock (`iframe_phone_manager_spec.coffee`).

scytacki:
This looks good to me. It is nice to have this so well tested. Hopefully it will pay off in the future. Thanks!

PR #44, opened 2015-04-24
Author can add Horizontal, Vertical and likert style MC

sagartalla:
[#92616816],[#92616126].

### DIFF
--- a/app/views/embeddable/multiple_choice_answers/_lightweight.html.haml
+++ b/app/views/embeddable/multiple_choice_answers/_lightweight.html.haml
@@ -12,13 +12,12 @@
       - if embeddable.show_as_menu
         = f.select :answers, options_from_collection_for_select(embeddable.choices, 'id', 'choice', (embeddable.answers.empty? ? nil : embeddable.answers.last.id) ), { :include_blank => t("PICK_ONE")}, { :data => {:button_id => check_button_string} }
       - else
-        - embeddable.choices.each do |choice|
-          %label{ :for => "embeddable_multiple_choice_answer_answers_#{choice.id}" }
-            - if embeddable.multi_answer
-              = check_box_tag 'embeddable_multiple_choice_answer[answers][]', choice.id, embeddable.answers.include?(choice), :id => "embeddable_multiple_choice_answer_answers_#{choice.id}", :data => {:button_id => check_button_string}
+        .choice-container{:style => embeddable.is_likert ? "display:table; width:100%;" : ""}
+          - embeddable.choices.each do |choice|
+            - if embeddable.is_likert
+              = render :partial => "embeddable/multiple_choices/show_choice_likert", :locals => {:embeddable => embeddable, :choice => choice, :check_button_string => check_button_string, :form => f}
             - else
-              = f.radio_button :answers, choice.id, :data => {:button_id => check_button_string}, :checked => (embeddable.answers.include?(choice))
-            = choice.choice
+              = render :partial => "embeddable/multiple_choices/show_choice", :locals => {:embeddable => embeddable, :choice => choice, :check_button_string => check_button_string, :form => f}
     - if embeddable.enable_check_answer
       %input{ :class => 'button check_answer_button',
               :type => 'button',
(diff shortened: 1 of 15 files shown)


app/views/embeddable/multiple_choices/_show_choice.html.haml
```diff
@@ -0,0 +1,6 @@
+%div{:style => (embeddable.layout == Embeddable::MultipleChoice::LAYOUT_HORIZONTAL ? "display: inline-block; margin-right: 10px;" : "") }
```
scytacki: @sagartalla `horizontal` should be added as a class, and the styles should be put into a separate css file.
probably best to add this class to the `.choice-container` above.

app/views/embeddable/multiple_choice_answers/_lightweight.html.haml
```diff
@@ -12,13 +12,12 @@
       - if embeddable.show_as_menu
         = f.select :answers, options_from_collection_for_select(embeddable.choices, 'id', 'choice', (embeddable.answers.empty? ? nil : embeddable.answers.last.id) ), { :include_blank => t("PICK_ONE")}, { :data => {:button_id => check_button_string} }
       - else
-        - embeddable.choices.each do |choice|
-          %label{ :for => "embeddable_multiple_choice_answer_answers_#{choice.id}" }
-            - if embeddable.multi_answer
-              = check_box_tag 'embeddable_multiple_choice_answer[answers][]', choice.id, embeddable.answers.include?(choice), :id => "embeddable_multiple_choice_answer_answers_#{choice.id}", :data => {:button_id => check_button_string}
+        .choice-container{:style => embeddable.is_likert ? "display:table; width:100%;" : ""}
```
scytacki: @sagartalla the layout type should be added as a class to the .choice-container then the inline style info here can be moved to a css file. As I mentioned in another comment, this can then also be used for the horizontal and vertical styling.

app/views/embeddable/multiple_choices/_show_choice_likert.html.haml
```diff
@@ -0,0 +1,7 @@
+%div.likert{:style => "display: table-cell; text-align: center; width:150px; vertical-align: bottom;"}
+  = "#{choice.choice}"
+  %br
+  - if embeddable.multi_answer
+    = check_box_tag 'embeddable_multiple_choice_answer[answers][]', choice.id, embeddable.answers.include?(choice), :id => "embeddable_multiple_choice_answer_answers_#{choice.id}", :data => {:button_id => check_button_string}
+  - else
+    = form.radio_button :answers, choice.id, :data => {:button_id => check_button_string}, :checked => (embeddable.answers.include?(choice))
```
scytacki: Because the 4 lines above are duplicated with the `_show_choice.html.haml` partial. It would be better if you combined these two partials into one. The style of the top level div can be done just looking at the inherit layout class from the .choice-container. So the only thing that needs to be conditionalized is the position of the `choice.choice`.

app/views/embeddable/multiple_choices/_form.html.haml
```diff
@@ -39,6 +56,34 @@
...
+          });
+
+          if(show_as_menu.is(':checked')){
+            vert.attr('disabled',true);
+            horiz.attr('disabled',true);
+            likert.attr('disabled',true);
+          }else{
+            vert.removeAttr('disabled');
+            horiz.removeAttr('disabled');
+            likert.removeAttr('disabled');
```
scytacki: If you add an id to `field_set_tag 'Layout'` then you can use a jquery selector to disable or enable all of the radio_buttons inside of this field_set with one call.  So if the id was a variable `embeddabe_layout_options` you could do $radios = $('##{embeddable_layout_options} input')
And then $radios.attr('disabled', true)

You should also clean up the duplicated code here.

app/views/embeddable/multiple_choices/_form.html.haml
```diff
@@ -39,6 +53,24 @@
...
+            }else{
+              $radios.removeAttr('disabled');
+            }
+          }
+
+          show_as_menu.click(function() {
+            toggle_$radios(this);
+          });
+
+          toggle_$radios(show_as_menu);
```
scytacki: Sorry for nit picking, but for consistency:

1. In Javascript we use camel case so the variable should actually be `showAsMenu`.
2. I used the `$` in front of the `radios` variable because it is jquery object.  If you are going to do that, then you should put the `$` in front of the `showAsMenu` variable as well.
3. It is not common to put a `$` in the middle of a function so the function should be `toggleRadios`
4. Since you have a reference to `showAsMenu` already, if you use that directly in `toggleRadios`, Then you can just pass `toggleRadios` directly to the click function: `showAsMenu.click(toggleRadios)` and you can call it directly below: `toggleRadios()`.
5. Finally the name `toggleRadios` doesn't seem right because that implies that the function will just flip the radio buttons  regardless of the value of showAsMenu. A better name would be `updateRadios`.

scytacki:
This looks good to me, I deployed it to authoring-demo.concord.org
But it currently can't be merged, please merge in master and resolve any conflicts.

PR #45, opened 2015-04-27
Arg block report

rklancer:
Together with the following additional changes already in master: f5ce69e...ac1f8ee and 765362a...a3ba8b1 this adds a POST url which, when provided with appropriate identifying information from the portal, initiates a download of a spreadsheet containing a report of the history of the requested learners' argumentation-block submissions, including the feedback returned by C-Rater or built in to the multiple choice

The POST url deliberately has CSRF checking removed, since the POST request comes from a portal, and no state change occurs as a result of the POST request. As noted by the Web Application Security project, CSRF attacks require the ability to effect a state change because untrusted code can't access the data returned by the POST request.

The main open Issue (cc @pjanik) on the LARA side is the complexity of `CRater::ArgumentationBlocksController#report` ([Code Climate report](https://codeclimate.com/github/concord-consortium/lara/compare/arg-block-report#issues)). I'm not sure of the right strategy to refactor this code appropriately, so I'd like to see what others think.

There are a couple of issues in the related portal pull request, but I'll describe them there.


rklancer:
Related portal pull request: https://github.com/concord-consortium/rigse/pull/58

pjanik:
I would consider adding some specs, as I can't see any. Taking into account hardcoded SQL query, some tests would be the easiest way to ensure that this code can survive upcoming DB migrations. I don't think we change table or attribute names too often, but it won't hurt.

Small thing - `select_params` hash seems to include only a few table names and types, while others are still hardcoded in the SQL query. I would either specify all the tables and types there or completely remove it (I lean towards it). It seems to be used in this one place only, so I don't think it brings much benefit. Things are hardcoded anyway and SQL query string is explicit and visible. 

Actually, you could obtain table name dynamically which would be a slight improvement (e.g. `Run.table_name`). But if we have proper specs (that obviously have to fail if tables don't match), I would just keep code as compact and simple as possible.

> The main open Issue (cc @pjanik) on the LARA side is the complexity of CRater::ArgumentationBlocksController#report (Code Climate report).

In general Rails app uses predefined directory structure, but nothing prevents us from adding a new one and using our own code organization. There is already `app/services` directory and I think it's the best place to move this function in (and refactor into class/module with a few smaller functions). There is only one existing example so far:
https://github.com/concord-consortium/lara/blob/master/app/services/create_collaboration.rb

In Portal we also have a few services which are a bit more complex and they follow common pattern related to input, output and `.call` method. In LARA it's a bit simpler, but I've still used `.call`. I don't think it's important, though. I would just organize code in a way which is the most convenient for a given task. It can be a class, module, whatever (I would just keep c_rater namespace there too).

Important benefit of it is that it should be much nicer and easier to test it (so you don't have to deal with excel spreadsheet returned from controller etc.).

scytacki:
@rklancer what is on your plate for tomorrow? Do you have time to work on this?

rklancer:
@scytacki @pjanik I think I'm running out of time to do much with tests

scytacki:
@pjanik can you estimate how long it would take to add a test for this?

It seems the best test would be:
Use Factories to create a few FeedbackSubmissions along with all of the related model objects used in the argblock report. Then generate a report, and compare it to an expected report.
I didn't find it, but I was hoping to find a CSV export from the Spreadsheet book object that this code creates. It would be nicer to compare a CSV diff instead of xls binary diff.

So I see two possibly time consuming parts:
- creating all the mock data
- figuring out how to get an export from Spreadsheet::Book that is nice to diff.  If that is not possible that it might be better to make an alternative output from this code that is CSV or some other format that is text based, perhaps YAML or JSON export.

scytacki:
Good Idea to use the spreadsheet gem to decrypt the spreadsheet result.  Looks good to me.

PR #46, opened 2015-04-28
A teacher can print a blank copy of an activity.[#89519960]

sagartalla:


app/views/shared/_activity_menu.html.haml
```diff
@@ -34,6 +34,13 @@
           %i.fa.fa-times
           %span.text
             Close
+      - if current_user && @activity.layout == LightweightActivity::LAYOUT_SINGLE_PAGE && (current_user.is_admin || @activity.user.email == current_user.email)
+        %li.am-print
+          %a{:href => preview_activity_path(@activity,:print => true), :target => "_blank", :title => "Print a blank copy"}
```
scytacki: It should print the current page instead of opening a new window.

scytacki: After the meeting today we decided that print param should continue to be supported. This way external sites can point directly at a URL which will load the page and immediately bring up the print dialog.

I now notice that you are using `preview_activity_path`, does that work ok for students?

sagartalla: @scytacki `preview_activity_path` does not work for students, we will be changing it to the activity run path.

scytacki: I'm sure this is confusing, but you should remove this path altogether.
The printing link should print the current page, so it shouldn't open a new window, so there is no need for a path.

We want to continue support a print param, so that we can construct a link that will automatically open the single_page layout of the activity and print it. This also does not require a path to be used here.

PR #48, opened 2015-05-22
Fix sorting of embeddable question types

pjanik:
Fixes sorting of Labbook embeddable item. Labbook dom element wasn't defining necessary ID.
I extracted `.authorable{:id => "embeddable_#{e.id}.#{e.class.to_s}"}` div from all partials and moved it one lever higher, so it'll be more difficult to make similar mistake in the future.


scytacki:
Nice approach.

PR #49, opened 2015-06-02
Add default text to open response questions [#95719746]

dougmartin:


scytacki:
Hmm. I was thinking that the default text would actually be place holder text. So it would override the "type answer here" placeholder text. However now I look at itsi.portal and I see that it isn't place holder text. I'll check with Carolyn to see if she needs the text to stick around after the user clicks.

scytacki:
I noted this in the story, but I answered my own question. And default text is the right way to go here.

One thing I wasn't sure about looking at the code: what happens when a a student has answered a question and then reloads the page, will the default_text override the students own answer? It obviously shouldn't, but that wasn't clear looking at the code. It seems it dependent on the implementation of rails `f.text_area` method.  A test of that would be helpful, I looked around though and couldn't figure out where that that should go. It could be a feature test like the single feature test we have in `spec/features`.  Or it could be a view test similar to spec/interactives_pages/show.

Another thing to test is whether the default text of the question gets saved as the students answer even if the student doesn't change it. I think we would prefer that this doesn't happen, but it might be tricky to implement.  I think we use a 'dirty' flag on questions and only save them if they are dirty, if that is all working properly then this might be easy. I think in the past we've done this server side: "if the student answer is blank, and the incoming answer matches the default answer then ignore it"

To test both of these things in one place a feature test would be best.

dougmartin:
@scytacki - I ended up bailing on the feature test yesterday.  I figured out how to switch to the newly opened window but the separate window opening was causing FactoryGirl to generate new mocks so I couldn't test the default text save.  I then trying using stubs to check if the default text was being saved but again the separate window opening was always causing the stub check to not be true.

The testing was moved to a controller test.

scytacki:
Were you able to verify that when the page is loaded, the students previous answer is shown instead of the default test? I suspect that it is, but it depends on how `f.text_area` decides to handle the case when `:value` is set along with a field name (in this case :answer_text).

dougmartin:
No, sorry. I'm going to verify it now.

dougmartin:
The default text was overriding the answer when the activity was run.  I've updated the view to show the default text only if the answer_text is null or an empty, non-trimmed string.

scytacki:
I'm not sure it would work, but it would be better to just check for nil answer_text. If a user decides to erase the whole thing and leave an empty string we should show them that empty string when they return.  I don't know if this will work though because we might already have empty strings in the database for un answered questions. And Active Record might convert empty string to nil. And finally it might be database dependent if nil can be stored in a string field.
Because of all of those unknowns I guess it is best to leave it how you've got it.

scytacki:
I'd merge this now, but github says it has conflicts.

dougmartin:
I'll rebase it and repush.

PR #50, opened 2015-06-09
An author can hide a question in an activity page [#95537580]

dougmartin:
This was done last week but I didn't have a chance to write the tests until today.  The feature test edits a page and toggles the hide/show links.  To get the feature tests working I added the Warden test helpers so that I could login as an admin to edit the page.  I also added a helper that waits for the Poltergeist driver to finish all ajax requests.  This helper is used after each test to ensure that the hide/show toggle completes before the database cleaner runs - otherwise there is a race condition that sometimes throws an error because of contention on the mysql thread.

### DIFF
--- a/app/controllers/interactive_pages_controller.rb
+++ b/app/controllers/interactive_pages_controller.rb
@@ -156,6 +156,17 @@ class InteractivePagesController < ApplicationController
     redirect_to edit_activity_page_path(@activity, @page)
   end
 
+  def toggle_hideshow_embeddable
+    authorize! :update, @page
+    update_activity_changed_by
+    PageItem.find_by_interactive_page_id_and_embeddable_id(params[:id], params[:embeddable_id]).toggle_hideshow_embeddable
+    if request.xhr?
+      respond_with_nothing
+    else
+      redirect_to edit_activity_page_path(@activity, @page)
+    end
+  end
+
   def reorder_embeddables
     authorize! :update, @page
     update_activity_changed_by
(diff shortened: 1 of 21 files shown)


pjanik:
Seems to work fine and generally looks good to me. Publishing also seems to respect `is_hidden` props (if it's easy to add some publishing-related test, it could be useful).

There is one tiny issue - when you try to hide a question and you click "Cancel", it will still disappear (not sure if request is sent or not).

Regarding implementation - we toggle question using `PagesController`, then we use `PageItem` model to finally update `Embeddable` model. :wink: I'm wondering if `PageItem` wouldn't be a good place for `is_hidden` attr or if we could stick only to Embeddable controller and models. But I'm betting that was determined by other factors that I can't see while briefly skimming through the code.

dougmartin:
The show/hide is the same as the delete link - it sends an immediate AJAX request that toggles the field.  That does mean that it can't be cancelled.

I thought about putting it in the PageItem model but adding it directly into each embeddable worked better for the partial views.

pjanik:
> The show/hide is the same as the delete link - it sends an immediate AJAX request that toggles the field. That does mean that it can't be cancelled.

Not sure if we understand each other correctly. Basically we have confirmation dialog saying "Are you sure you want to hidethis element? You will lose data from X learners that have answered this question.". And when I click "Cancel", my question is still being hidden. AFAIR it didn't really send request to the server, but view was modified. It works okay for "remove" link. Perhaps small thing somewhere in JS.

dougmartin:
Good catch.  Looks like the issue is that Rails skips the data-confirm if the request is marked as remote.  I'll move the confirmation code into the click handler.

dougmartin:
Ok, I moved the confirm to the click handler.

pjanik:
@dougmartin, after I merged this pull request I noticed that questions aren't actually hidden in the runtime. :wink: One tricky thing there could be to ensure that questions numbers are still correct.

dougmartin:
Looking now.

PR #51, opened 2015-06-09
Add flag for ITSI style authoring [#95539464]

dougmartin:
Adds a select for a new editor_mode integer field on the lightweight_activities model.  The select is shown both on the new and edit pages.

The style.scss changes are a side fix to share the the edit_lightweight_activity styles with the unstyled new_lightweight_activity.

Per the meeting I also added a mode query parameter to the edit page so that admins can override the edit mode.

### DIFF
--- a/app/assets/stylesheets/style.scss
+++ b/app/assets/stylesheets/style.scss
@@ -506,42 +506,43 @@ ul.quiet_list {
     background: #fef7e0;
   }
 }
-form.edit_lightweight_activity {
+form.edit_lightweight_activity,
+form.new_lightweight_activity {
   fieldset {
     clear: left;
-   	margin-top: 20px;
+     margin-top: 20px;
   }
   legend {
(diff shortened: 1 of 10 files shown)


dougmartin:
Wait on this please.  I'm going to add a temporary third option that actually changes the view to an empty view we can use to start building the React editor.

dougmartin:
Ok - the third option has been added.

app/assets/stylesheets/style.scss
```diff
@@ -506,42 +506,43 @@ ul.quiet_list {
     background: #fef7e0;
   }
 }
-form.edit_lightweight_activity {
+form.edit_lightweight_activity,
+form.new_lightweight_activity {
```
dougmartin: This is the only new content in the file - the rest is space deletion.  The new form was unstyled.

PR #52, opened 2015-06-11
An author can hide pages in an activity.

pjanik:
I've added a new checkbox to the top section of page edit form that let authors hide the page.
Hidden page is not displayed in the runtime (404 if user tries to get there via direct URL) and question numbers should respect visibility of pages. Also, when the first or last page is hidden, navigation buttons visible on the pages between them handle that. So, for example user will still see "Generate a report button" on the last visible page, there will be no "Next >" button if the last page is hidden and so on.

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -28,9 +28,6 @@ LightweightStandalone::Application.configure do
   config.consider_all_requests_local       = true
   config.action_controller.perform_caching = false
 
-  # Raise exceptions instead of rendering exception templates
-  config.action_dispatch.show_exceptions = false
-
   # Disable request forgery protection in test environment
   config.action_controller.allow_forgery_protection    = false
 
--- a/app/controllers/interactive_pages_controller.rb
+++ b/app/controllers/interactive_pages_controller.rb
@@ -4,6 +4,7 @@ class InteractivePagesController < ApplicationController
   before_filter :set_page, :except => [:new, :create]
   before_filter :set_run_key, :only => [:show, :preview]
   before_filter :set_sequence, :only => [:show]
+  before_filter :check_if_hidden, :only => [:show, :preview]
 
   before_filter :enable_js_logger, :only => [:show, :preview]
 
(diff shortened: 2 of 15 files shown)


pjanik:
Okay, I've fixed tests and added handling of one more edge case - when activity run points to the hidden page (e.g. user left activity there and author marked this page as hidden after that), user won't be redirected there anymore. Probably not very likely for students, but quite likely when you're authoring an activity and testing it yourself.

dougmartin:
@pjanik Could you resolve the merge conflict on Monday?  I need to change some code near your changes in this PR.  Thanks.

pjanik:
Done.

config/environments/test.rb
```diff
@@ -28,9 +28,6 @@
   config.consider_all_requests_local       = true
   config.action_controller.perform_caching = false
 
-  # Raise exceptions instead of rendering exception templates
-  config.action_dispatch.show_exceptions = false
-
```
scytacki: @pjanik, Won't this make it harder to see what went wrong when looking at a test console output?

scytacki: I see this is needed to check for the 404 in the test. So in this case there is probably an exception in the log/test.log file.  Can you add a comment to the test(s) that require this change, incase someone tries to add `show_exceptions = true` back in?
This way when the tests start failing, the developer might figure out why more quickly.

scytacki:
This looks good. And now with the bind polyfill, the tests are passing. 
The only change is to document the tests that require `show_exceptions = true`.

config/environments/test.rb
pjanik: I changed it, as my controller was supposed to throw `RecordNotFound` in some cases. I wanted to test this particular case, but then I was getting status 500 instead of expected 404. 

I don't think this option affects a test console output itself. Probably you can see it only if you debug tests and inspect the page content (which is not very handy in console anyway) or if you use Selenium tests and a real web browser (which we don't use in LARA). I feel this option is mostly useful in dev mode, however I can try to look for another solution if you find this option useful in tests too.

scytacki: It is fine to leave it this way, I just added a comment in the test incase someone changes the setting back for some reason.

PR #54, opened 2015-06-12
Move ITSI editor from page controller to activity controller

pjanik:
Simple change, ITSI authoring should work for the whole activity, not page.


dougmartin:
Looks good.

PR #55, opened 2015-06-12
Add simple metadata editor to ITSI authoring page

pjanik:
This PR adds a simple activity editor to the ITSI authoring page. It also adds and uses `ModuleJS`. Features are pretty trivial, but it's more about setting some structure and approach.

It's not production ready, so to speak, but we are working on the page that is marked as WIP anyway, so I think we can merge this kind of pull requests to master. Otherwise, we can keep this branch around and create PR against it instead of master.

@dougmartin, feel free to modify this code if you feel so. React forms were actually something new to me, so I'm not sure if that's the best approach.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended to check this file into your version control system.
 
-ActiveRecord::Schema.define(:version => 20150609202753) do
+ActiveRecord::Schema.define(:version => 20150610120725) do
 
   create_table "admin_events", :force => true do |t|
     t.string   "kind"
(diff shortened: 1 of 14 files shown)


db/schema.rb
```diff
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended to check this file into your version control system.
 
-ActiveRecord::Schema.define(:version => 20150609202753) do
+ActiveRecord::Schema.define(:version => 20150610120725) do
```
pjanik: This PR doesn't include any migrations, I'm betting someone didn't commit updated `schema.rb` before and I accidentally added it.

dougmartin:
The build is failing because the spec wasn't updated when the edit page was moved.  I'm going to merge this and then submit a PR to fix the spec.

dougmartin:
@pjanik I merged this thinking the spec errors were related to the move of the edit page so it was safe to merge.  However, the issue is with the new items added to application.js.  If you remove these the edit page spec errors go away:

    //= require modulejs
    //= require react
    //= require react_ujs
    //= require components

Here is the spec to run to see the error:

    RAILS_ENV=test bundle exec rspec spec/features/hide_show_questions_spec.rb

Could you look at this on Monday?

pjanik:
Sure, I'm going to look at it.

PR #56, opened 2015-06-12
Changed open response save to use nil instead of empty string when it matches default text

dougmartin:


dougmartin:
@scytacki The specs are failing now due to the React include error.  On the first push they failed because the unit test was checking for an answer text of "" instead of nil.

scytacki:
Nice, I didn't know about the `[attribute]_was` feature of active record.

PR #57, opened 2015-06-15
Add Function.bind polyfill so JS code works in PhantomJS

pjanik:
Failing tests were caused by the fact that Phantom 1.x doesn't support `Function.prototype.bind` and React is using it. 

We can either upgrade PhantomJS everywhere to 2.x or simply add polyfill. I've done the latter, as it's simpler. TravisCI has Phantom 1.x preinstalled. You can install PhantomJS 2.x there:
https://mediocre.com/forum/topics/phantomjs-2-and-travis-ci-we-beat-our-heads-against-a-wall-so-you-dont-have-to
but I'm not sure if it's worth it yet. Perhaps sooner or later TravisCI will migrate and homebrew will support 2.x. :wink:

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -11,10 +11,10 @@
 // GO AFTER THE REQUIRES BELOW.
 //
 // N.B. will need to load specific jquery-ui modules for production
+//= require bind-polyfill
 //= require modulejs
 //= require react
 //= require react_ujs
-//= require components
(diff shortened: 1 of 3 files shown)


scytacki:
@pjanik is this code copied from somewhere? If so we need some license information about it. If you have it, you could put it at the top of `bind-polyfill.js`. If you don't have it we should ask for it, or look for a version with license information.

PR #58, opened 2015-06-15
Basic editor skeleton created with section editor placeholders

dougmartin:
This sets up the layout of the ITSI editor, both for existing activities and new activities.  Once the embeddable and interactives are added to the JSON we can flesh out the section editors.


pjanik:
So, following earlier discussion on Slack, I would vote to define the structure of the activity using JSON. It should be more flexible and the code should be a bit simpler too. We wouldn't need developer to change the basic template and I expect that it might be quite useful. Here I can see that sections can be either predefined or specified in JSON, but I'm not sure if it's working for `elements` which seem to be predefined in the top-level component anyway. Also, if section was undefined, it we would need to create it first, what makes things more difficult (more cases to handle). If we stick to the assumption that all the data models are already in DB, it's just about simple edits and toggling visibility. 

Also, a generic editor could be a base for a new, default authoring page in the future (as the existing one is pretty bad).

Another thing, but that might be related to early stage of this work - we would need separate update URLs for all question types probably, not just activityUrl.

In general I would keep React components very close to Rails models when it comes to scope of responsibility, names and so on. So mapping between React and Rails models is clear and each React component talks to corresponding Rails controller. One change is that "page" in Rails == "section" in React.  Another exception would be `Sensor` and `Model`, as there would need to be more complicated interaction between them. So we can't have a simple `Interactive` component only. But all that seems to be true in this PR.

dougmartin:
The code was structured to fit the current export from ITSI where pages/sections that are hidden don't get created in LARA.  If we add the is_hidden flag to export it is a simple change to support that in the React code.

If we do use the JSON to layout the editor then we also assume that all the pages exist for ITSI activities and we'll need to check/add the following:

1. Check if new ITSI activities can be created in LARA and if so we'll need to generate their pages.
2. Add a "sectionType" to the JSON instead of using the page name for future extensiblility.

dougmartin:
I could not squash all the commits so I created a new branch and copied all the files over and opened a new PR here: https://github.com/concord-consortium/lara/pull/63

PR #60, opened 2015-06-15
Add option for "preflighting" an iframe interactive

psndcsrv:
Certain browsers (IE and Safari) won't load cookies for a third-party
domain within in iframe unless the user has been to that domain in a
normal browser window. Pre-flighting pops up a window with a specific
url under that domain which initializes cookies for that domain and then
closes itself. Once that is complete, the iframe can be loaded and
cookies will work as expected.

[#96215356]
[#92480624]

### DIFF
--- a/app/views/mw_interactives/_show.html.haml
+++ b/app/views/mw_interactives/_show.html.haml
@@ -35,11 +35,28 @@
 
 :javascript
   $(document).ready(function() {
-    $('##{dom_id_for(interactive, :click_to_play)}').click(function() {
-      $(this).css('display','none');
-      $(this).removeClass('shown')
+    var loadIframe = function(component) {
+      $(component).css('display','none');
+      $(component).removeClass('shown')
       $('##{dom_id_for(interactive, :interactive_image)}').css('display','none');
       $('#interactive_#{interactive.id}').attr('src','#{interactive.url}');
-      InteractiveManager.register($(this).parent('.interactive-container'))
+      InteractiveManager.register($(component).parent('.interactive-container'));
+    };
+    $('##{dom_id_for(interactive, :click_to_play)}').click(function() {
+      var component = this,
+          preflight_url = $('#interactive_#{interactive.id}').data('preflight_url');
+      if (preflight_url) {
+        // pop up a pre-flight window. When it closes itself, load the iframe.
+        var win = window.open(preflight_url, 'preflight_window', 'height=120,width=120,location=0,menubar=0,resizable=0,scrollbars=0,status=0,titlebar=0,toolbar=0');
+        win.blur(); // pop under the main window
+        var poller = setInterval(function() {
+          if (win.closed !== false) {
+            clearInterval(poller);
+            loadIframe(component);
+          }
+        }, 200);
+      } else {
+        loadIframe(component);
+      }
     })
   })
--- a/app/views/mw_interactives/_show.html.haml
+++ b/app/views/mw_interactives/_show.html.haml
@@ -10,23 +10,12 @@
         See :-
         =link_to "#{interactive.url}","#{interactive.url}"
 
-  - if interactive.click_to_play
-    .click_to_play.shown{:id => dom_id_for(interactive, :click_to_play)}
-      .background
-      .text= "Click here to start the interactive."
+  = click_to_play_barrier(interactive)
+  = interactive_iframe_tag(interactive, @run, !interactive.click_to_play) do
 
-    = interactive_iframe_tag(interactive, @run,false) do
-
-      Your browser does not support inline frames, or is currently not configured to display inline frames.
-      Content can be viewed at the actual source page:
-      = link_to interactive.url, interactive.url
-
-  - else
-    = interactive_iframe_tag(interactive, @run) do
-
-      Your browser does not support inline frames, or is currently not configured to display inline frames.
-      Content can be viewed at the actual source page:
-      = link_to interactive.url, interactive.url
+    Your browser does not support inline frames, or is currently not configured to display inline frames.
+    Content can be viewed at the actual source page:
+    = link_to interactive.url, interactive.url
 
   - if interactive.save_state
     .below_interactive
--- a/app/assets/javascripts/iframe-saver.coffee
+++ b/app/assets/javascripts/iframe-saver.coffee
@@ -59,6 +59,15 @@ class IFrameSaver
             else
               @$delete_button.hide()
 
+        # A preflight url is a url which should be loaded in a popup window. This allows sites that require cookies
+        # to be able to initialize in browsers that require you to visit a page in a full browser window before they
+        # will allow the site to access cookies when embedded in an iframe.
+        if opts.preflight_url?
+          @$iframe.data('preflight_url', opts.preflight_url)
+          ctp = @$iframe.siblings('.click_to_play')
+          ctp.addClass('shown')
+          ctp.show()
+
       @iframePhone.post('getExtendedSupport')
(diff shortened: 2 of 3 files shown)


scytacki:
I don't know how much time you have to spend on this. If you have more time, I think there is a way to do this where LARA authors would not need to put in the preflight URL. Instead the page at the interactive URL would be smarter. LARA would provide an postMessage API for this smarter page to trigger the opening of the pre-flight page.   I documented the approach here:
https://www.pivotaltracker.com/story/show/83971486
However, the approach hasn't been tested. So there might be issues with all of the changing of the iframe location. Also, confusingly, that story is about embedding LARA in the portal, instead of embedding something in LARA.

Finally the story describes doing this in a way that is more visible to the user, but you don't need to do that. I think making it more visible to the user, will improve the chances it will continue working in the future.  But at the same time hidden top level window approach, is the same that is used by Schoology and has survived around 5 years I think.  One rationalization for its survival is noted in a comment of this SO answer: http://stackoverflow.com/a/10098007
" I'm not so sure it'll get axed like other solutions because it requires that a user actually click/interact with the page in order to have the popup not blocked. This solution that Safari has come up with seems to work well: advertisers won't be able to secretly set cross-domain cookies, and embedded apps will require a user to interact before being able to set a cookie."

app/views/mw_interactives/edit.html.haml
```diff
@@ -15,6 +15,12 @@
     .clear
     This width and height will be used to maintain a fixed aspect ratio for the interactive.
   %br
+  = field_set_tag 'Pre-flight URL' do
+    = f.text_area :preflight_url, size: "45x4"
+  %div{:style => "margin-bottom:5px;"}
+    %em Warning: A Pre-flight URL automatically enables Click to Play.
+    Please provide an image_url to use a Pre-flight URL.
```
knowuh: Does this image have to live on the same host as the interactive?  If so, I think we need more instructions for the author.

app/views/mw_interactives/_show.html.haml
```diff
@@ -35,11 +35,27 @@
...
+      InteractiveManager.register($(component).parent('.interactive-container'));
+    };
+    $('##{dom_id_for(interactive, :click_to_play)}').click(function() {
+      if ('#{interactive.preflight_url}' !== '') {
+        // pop up a pre-flight window. When it closes itself, load the iframe.
+        var component = this,
+            win = window.open('#{interactive.preflight_url}', 'preflight_window', 'height=120,width=120,location=0,menubar=0,resizable=0,scrollbars=0,status=0,titlebar=0,toolbar=0');
+        win.blur(); // pop under the main window
+        var poller = setInterval(function() {
+          if (win.closed !== false) {
```
knowuh: There must be some special JS on the page itself which closes the window?  Where is that happening?

knowuh:
I read this commit initially without reading the PT stories, or items that @scytacki mentioned in his comments. I was a bit confused without reading that material.

I had some questions about what the preflight_url should look like, and where it should live.  

It looks good, but it would be great to document  that we are opening the preflight window to work around third party cookie problems.

app/views/mw_interactives/edit.html.haml
psndcsrv: When I was testing it out, it actually looks fine (for CODAP at least) if there's not an image_url specified. You just end up with the words "Click here to start the interactive" over a white space.

If we end up using the flow @scytacki described, then this will get removed from the authoring anyway.

app/views/mw_interactives/_show.html.haml
psndcsrv: Yeah, the page I've set up for testing in the Document Server just uses javascript to close the window immediately:
```
<script>
  window.close();
</script>
```

psndcsrv:
@scytacki That was the more complicated solution that I'd thought about, but went with this version since it was easier and less complicated.

After talking with @ddamelin, I'm going to go ahead and implement your strategy since it'll be less obnoxious to the end user.

scytacki:
You should probably create some kind of mock setup, so you can see if the popup or tab is blocked if the parent page opens it.  Or perhaps you can rework it so the popup is opened by a user click event. In that case it seems fine to automatically close the popup.  Just like the SO post.

psndcsrv:
I think we're basically required to have the user initiate the popup via an interaction, otherwise the browsers will block the popup. I was imagining having the Doc Server tell LARA that it needs a preflight (and pass the url in that communication), and that would trigger a click-to-play style interface prompting the user to click to start things up. That click would trigger the popup and reload (or replace) the iframe.

knowuh:
How is this looking?  @psndcsrv @scytacki?

psndcsrv:
@knowuh I got distracted when I was trying to test my changes. It turns out the original problem I was investigating wasn't (totally? or even partially?) caused by the need to preflight.

I think these changes are ready to go, otherwise. They should probably get tested... and they probably need better documentation as to what needs to happen on the interactive side to make use of them.

scytacki:
I think we'll need this soon for the ModelMyWatershed project. 
@psndcsrv I'm not sure how this code works.
It seems the click_to_play barrier is getting the preflight_url as a data attribute from the interactive element.  But this data attribute is not set until the extendedOptions is returned by the iframe. Since the iframe will not be loaded before the click_to_play barrier is clicked, where does the extendedOptions come from?

psndcsrv:
@scytacki This is piggy-backing the same mechanism that the Document Server uses to disable the reset button. It flows like this:

1. The activity opens and the iframe loads its normal interactive url. The click-to-play barrier is either not shown, or doesn't do the preflight if it was enabled.
1. Once the interactive loads (either automatically as part of the page load, or because of a click), if the interactive decides that its unable to save cookies and needs to, the interactive sends the `extendedOptions` with a preflight url. If the interactive decides that it already can save cookies, then we skip the preflight steps and continue as normal.
1. The activity page detects the desire for a preflight, and (re-)enables the click-to-play barrier.
1. The user clicks the click-to-play barrier, which causes a new window to pop up, loading the preflight url.
1. The preflight url should be a quick and simple page that sets a cookie and then closes itself.
1. Once the popup closes, the activity page reloads the iframe.

Forcing the user to click to  start the iframe was necessary because most browsers don't allow a popup window to be created by javascript unless it's a direct result of a click by the user. It's a little weird if the author specifies to enable click-to-play *and* the interactive needs a preflight, because the user will then have to click twice to get the interactive to start the first time they encounter it.

Does that all make sense? It's a lot of bouncing around, for sure.

scytacki:
If I've got it right, then it means on a browser that supports third-party cookies, the user won't see the click to play barrier at all. Or we add the click to play barrier and then user on a blocking browser will see it twice. That will make it confusing for ITSI teachers. 

So here is a modification to this approach:
- We add a 'interactive_domains' db table to LARA, this table maps an interactive domain to a preflight_url and a cookie_checking_iframe_url.
- When lara sees an interactive from one of these domains, it puts an iframe on the page with a src of the  cookie_checking_iframe_url.  If this iframe tells lara it doesn't have a cookie, then LARA puts up the click to play barrier (if it hasn't already been configured).

The point of the cookie_checking_iframe_url is so the domain can provide a very simple fast page to check the cookies instead of requiring the full interactive to load and then do this checking.

I'm not sure when we can fit this in. It could be shared cost between CODAP and MMW.

If CODAP needs this soon, I'm ok with merging this in and then tweaking it later.  The only thing it really needs is a copy of the flow description in the documentation. Or that could be added here: http://concord-consortium.github.io/lara-interactive-api/docs/

PR #61, opened 2015-06-16
An author can hide an interactive when there is more than one interactive in a section

pjanik:
Note that it's still impossible to render multiple interactives on the single page in runtime. The first visible interactive will be rendered.


dougmartin:
Looks good.

PR #63, opened 2015-06-17
Updated React-based ITSI editor from simple placeholder to working version

dougmartin:


dougmartin:
New PR created after I had merge problems squashing all the commits in this PR: https://github.com/concord-consortium/lara/pull/58

pjanik:
It's a huge PR, so it's difficult to say anything else than that generally it looks very good. :wink:
I would only keep testing in mind.

scytacki:
I agree that this needs some tests, but I'm not sure what the best way to do that is. We could add integration tests with Capybara. But perhaps there is something lighter weight that would make sense. Perhaps one approach would be to create Jasmine tests with XHR mocking to verify that it is sending the right kind of data to the LARA.  I think the last time Piotr looked at that he like using the Capybara matchers better.

PR #64, opened 2015-06-17
Add bg_url to ITSI editor json, fix interactives handling

pjanik:


app/services/itsi_authoring/editor.rb
```diff
@@ -33,22 +33,21 @@ def section_json(page)
...
+        {}
+    end
   end
 
   def open_response_json(e)
     {
+      type: 'open_response',
       name: e.name,
       prompt: e.prompt,
       is_hidden: e.is_hidden,
```
dougmartin: Could you add the default_text here too?  Sorry, I forgot to mention that in the email.

pjanik: Done.

PR #65, opened 2015-06-17
Protal-trigger the creation of this user in LARA

sagartalla:
LARA part for this story https://www.pivotaltracker.com/story/show/95625522


scytacki:
It looks like this is missing a file I don't see the definition of the PortalUser object.

PR #67, opened 2015-06-18
Add feedback element for ITSI editor saves [#97298198]

dougmartin:
This also updates interactive_pages_controller.rb so that it returns properly formatted json on xhr json updates (otherwise the save of the page text shows that it fails with these new alerts).

If the default alert bar seems to intrusive there is also a 'pill' option available that can be used as the default.  It places a rounded "pill" in the top right corner instead of adding an entire bar at the top of the page.

### DIFF
--- a/app/assets/javascripts/components/itsi_authoring/alert.js.coffee
+++ b/app/assets/javascripts/components/itsi_authoring/alert.js.coffee
@@ -1,0 +1,11 @@
+{div} = React.DOM
+
+modulejs.define 'components/itsi_authoring/alert',
+->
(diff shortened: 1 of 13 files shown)


dougmartin:
@scytacki Let me know if you think the alert bar is too intrusive.  If you pull this down you can change the alert to a 'pill' in the upper right by changing the default value of the layout parameter here to 'pill':

https://github.com/concord-consortium/lara/pull/67/files#diff-e5aed04194bdbf5d3f301fe554facfc5R34

PR #70, opened 2015-06-22
Hide hidden embeddable questions in runtime [#95537580]

dougmartin:
This removes the embeddables marked as hidden from the runtime.

### DIFF
--- a/app/models/interactive_page.rb
+++ b/app/models/interactive_page.rb
@@ -102,6 +102,19 @@ class InteractivePage < ActiveRecord::Base
     section_embeddables(nil)
   end
 
+  def visible_embeddables
+    self.page_items.collect{ |qi| qi.embeddable }.select{ |e| !e.is_hidden }
+  end
+
+  def section_visible_embeddables(section)
+    self.page_items.where(section: section).collect{ |qi| qi.embeddable }.select{ |e| !e.is_hidden }
+  end
+
+  def main_visible_embeddables
+    # Visible embeddables that do not have section specified (nil section).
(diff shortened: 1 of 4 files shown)


app/models/interactive_page.rb
```diff
@@ -102,6 +102,19 @@ def main_embeddables
     section_embeddables(nil)
   end
 
+  def visible_embeddables
+    self.page_items.collect{ |qi| qi.embeddable }.select{ |e| !e.is_hidden }
```
knowuh: I was wondering if `qi` is a `page_item` so i would have expected the var to be `pi`  -- but its so obvious, it doesn't matter.  Same in the next method.

knowuh:
Looks good.  :+1:

scytacki:
This looks good, but please add some spec tests. 
There are already some spec tests for hidden pages, so I think just adding some cases with hidden embeddables should be straight forward:

test of the questions method: https://github.com/concord-consortium/lara/blob/staging/spec/models/lightweight_activity_spec.rb#L105
test of the serialization for the portal: https://github.com/concord-consortium/lara/blob/staging/spec/models/lightweight_activity_spec.rb#L271

The serialization for the portal is how LARA updates what is reported in the portal. So if the serialize method is working correctly then the portal reports should too.  The LARA "summary" report is based on the questions method so if that method is working properly then that report should be good.

scytacki:
I'm going to merge this anyway, so we can have a better demo with Carolyn.
Here is chore so we remember to add the tests:
https://www.pivotaltracker.com/story/show/97542862

PR #71, opened 2015-06-22
Editing page looks like the ITSI editing page.  [#97209594]

knowuh:
Small CSS changes to make ITSI editing look like it does in the portal.
https://www.pivotaltracker.com/story/show/97209594

The main change is conditionally adding a the `itsi-edit` css class to the top level `#container` -- and adding several styles matching the `#container.itsi-edit >` selector so that the form matches `template_edit` in the portal. eg: http://itsi-aws.staging.concord.org/activities/3275/template_edit

### DIFF
--- a/app/assets/stylesheets/components/itsi_authoring.scss
+++ b/app/assets/stylesheets/components/itsi_authoring.scss
@@ -1,122 +1,161 @@
-.ia-editor {
-  margin: 20px;
-}
 
-.ia-save-btn {
-  font-size: 12px;
-  display: inline-block;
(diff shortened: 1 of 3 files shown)


dougmartin:
Looks good.  :+1:

knowuh:
We might get a merge fail somewhere, because this and other pending PR's modify `itsi_authoring.scss`.   Please use mine :)

scytacki:
sorry @knowuh you lost in this merging duel. :) Can you update this PR so it merges cleanly?

PR #72, opened 2015-06-23
Replaced answer#question_index with template loop variable

dougmartin:
This fixes the missing question numbers in single page activities.  I tracked this reversion down to this commit inside of Piotr's hide page PR:

https://github.com/concord-consortium/lara/commit/e81b63a2c2a549b1a6138e79cae30f260b055098

Before this commit the question numbers show, after it they are blank.

Each of the lightweight embedded partials use the answer#question_index method to find its 1-based index in the list of questions for the page.  With the above commit the answer.question was no longer being found in the list of activity.questions so it was returning nil which rendered as an empty string in the template.  After looking at the code and reading the "DANGER TODO" comment (see below) above the method I saw that the method call could be replaced by a simple template loop variable in the partials.  I thought this would be the safest way to fix the problem.

    # DANGER TODO:  This memoization isn't going to work
    # the way the author intended, because that instance var is going
    # to be shared by every class that includes this module.
    @question_index = nil
    def question_index(skip_cache=false)

NOTE: This fixes the question number issue but I'm concerned there may be other problems, for instance in reporting, related to the anwer.question not being the the list of activity.questions.  I left the anwer#question_index code in so that we could add a failing test for it and track down a fix in a separate PR.

### DIFF
--- a/app/views/c_rater/argumentation_block/_runtime.html.haml
+++ b/app/views/c_rater/argumentation_block/_runtime.html.haml
@@ -27,7 +27,7 @@
 
     - embeddables.each_with_index do |m,index|
       .question
-        = render :partial => "#{m.class.name.underscore.pluralize}/lightweight", :locals => {:embeddable => m}
+        = render :partial => "#{m.class.name.underscore.pluralize}/lightweight", :locals => {:embeddable => m, :question_index => index + 1}
         - fb = m.get_saved_feedback
         - fb_visible = fb && !fb.feedback_text.blank?
         - fb_dirty = fb && fb.outdated?
@@ -65,7 +65,7 @@
           %tr
             %td
               .ab-robot-figure{:id => 'ab-robot-image'}
-            %td         
+            %td
               %div
                 = t('ARG_BLOCK.FEEDBACK_ON_FEEDBACK')
                 - CRater::FeedbackSubmission.usefulness_score_names.each do |value, label|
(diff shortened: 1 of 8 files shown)


dougmartin:
@scytacki or @knowuh - please review.

knowuh:
Well you change looks clean and simple enough, so I would approve those changes.  A simple view template variable seems like a reasonable approach, and a similar pattern could be used in the report view.  I haven't done a 'search in project' to see where else we are relying on `question_index`, but that could shed some light.

If e81b63a was behaving like you described:

> With the above commit the answer.question was no longer being found in the list of activity.questions so it was returning nil which rendered as an empty string in the template

Then we might be masking a bigger problem. The answers need to exist when the page is rendered, so that they can be updated on change via JSON.  If I understand your comment, the (unanswered) answer id's might not be set correctly.

I will look at this some more a bit later.

dougmartin:
@knowuh answer#question_index is not used anywhere else after this change.  I left it in as I mentioned in the trailing note so that we can potentially create a failing test with it in order to track down any other potential other reversions caused by the page hide commit.

dougmartin:
I fixed the larger issue with hidden pages which resolves this issue with no changes needed.  Closing without merging.

ddamelin:
This is probably unrelated, but we currently have a mismatch between the LARA numbering and the portal report numbering when an interactive saves state but does not have a report URL. LARA reserves a question number for the saved state, but the portal report just numbers things sequentially, so if the saved state is not manifested in the report (typically as a link as it does for CODAP and the lab book), the portal numbering sequence gets behind the LARA numbering sequence.

PR #73, opened 2015-06-23
Fixed single and multi-page show to exclude hidden pages

dougmartin:
This commit has two parts:

1. Adds a migration to remove the nullability of is_hidden in pages.  The visible_pages has_many association used a is_hidden => false condition which was false for pages that had is_hidden set to null.  This caused multiple issues with code using this association.

2. Updates the single page and multi-page runtime template to use the visible_pages instead of pages association.


dougmartin:
@knowuh This fix corrects the question numbering so I closed that PR without a merge.

dougmartin:
Looking at failed tests now.

knowuh:
Yeah, I see the issue now.  @dougmartin code looks good.  I suspect some of those tests might have assigned `@pages` directly, and maybe didn't setup or mock `@activity` -- just a guess though.

dougmartin:
I've fixed all but one test locally.  Waiting to see if Travis encounters the remaining test error.

PR #74, opened 2015-06-25
Associate a Labbook with an Interactive. 

knowuh:
For the first part of this story:  https://www.pivotaltracker.com/story/show/97734786

Next step is:
> In standard mode authoring show the user which interactive a LabBook is associated with.

@dougmartin if you get a chance.

### DIFF
--- a/spec/models/embeddable/labbook_answer_spec.rb
+++ b/spec/models/embeddable/labbook_answer_spec.rb
@@ -5,7 +5,7 @@ describe Embeddable::LabbookAnswer do
 
   let(:labbook) { Embeddable::Labbook.create }
   let(:run) { Run.new }
-  let(:labbok_answer) { Embeddable::LabbookAnswer.create(question: labbook, run: run) }
+  let(:labbook_answer) { Embeddable::LabbookAnswer.create(question: labbook, run: run) }
 
   describe '#portal_hash' do
     let(:expected_hash) do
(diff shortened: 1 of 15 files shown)


spec/models/embeddable/labbook_answer_spec.rb
```diff
@@ -41,4 +41,19 @@
       expect(a_request(:post, labbook_replace_url)).to have_been_made.once
     end
   end
+
+  describe '#show_in_runtime?' do
+    describe 'with a disabled labbok' do
+      let(:labbook) { stub_model(Embeddable::Labbook, show_in_runtime?: false) }
+      it "should return false" do
+        expect(labbok_answer.show_in_runtime?).to eql(false)
```
dougmartin: Since you are in this file could you fix the typo "labbok_answer" => "labbook_answer"?

spec/models/embeddable/labbook_spec.rb
```diff
@@ -39,4 +39,34 @@
...
+        expect(labbook.is_connected?).to eql(false)
+      end
+
+      it 'it should not show up in the runtime' do
+        expect(labbook.show_in_runtime?).to eql(false)
+      end
+    end
+    describe 'when it has an interactive' do
+      let(:interactive) { MwInteractive.new  }
+      let(:args) { {interactive: interactive}}
```
dougmartin: I learned something here.  So inside this describe the outer :labbook would be reevaluated using the changed :args symbol?

dougmartin:
Looks good :+1:

spec/models/embeddable/labbook_spec.rb
knowuh: Yes. This is one of my favorite features of how lazy let evaluation works in specs.

spec/models/embeddable/labbook_answer_spec.rb
knowuh: Thanks. Fixed in 8634ce7

PR #77, opened 2015-06-26
Add interactive to labbook

knowuh:
First ½ of [Associate a Labbook embeddable with an interactive](https://www.pivotaltracker.com/story/show/97734786)

Second piece will be:

> In standard mode authoring show the user which interactive a LabBook is associated with.

BTW: This is the second attempt to merge this branch. Second time is a charm.   :)


knowuh:
@dougmartin this should be the same PR as before.

PR #80, opened 2015-06-29
Add redirects to image proxy [#98026438]

dougmartin:


scytacki:
Instead of rolling our own redirect handler. Can you try using the HTTPParty gem:
http://johnnunemaker.com/httparty/

The gem is already included in LARA. Perhaps it will have issues handling the binary image bytes, but I think we should try to use it and find out.

dougmartin:
@scytacki I updated it to use HTTPParty

dougmartin:
Wait on this - it isn't sending the content type down correctly yet.

dougmartin:
Ok, the content-type is being correctly sent.

scytacki:
:+1:

PR #81, opened 2015-06-29
In standard authoring show which interactive a LabBook belongs to

knowuh:
This should complete [#97734786]

Update Labbook forms & tests

https://www.pivotaltracker.com/story/show/97734786

### DIFF
--- a/app/assets/stylesheets/style.scss
+++ b/app/assets/stylesheets/style.scss
@@ -271,6 +271,13 @@ button.disabled, .btn-primary.disabled, span.edit_trigger.disabled {
     width: 98%;
   }
 }
+.embeddable_options {
+  .warning {
+    color: #ba431e;
+    font-weight: bold;
+    margin-top: 1em;
+  }
+}
 .questions.full-width {
   .embeddable {
     float: left;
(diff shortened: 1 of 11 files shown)


knowuh:
@dougmartin when you get a chance.

dougmartin:
This looks good to me but I think someone with deeper Rails experience should look at it.  @scytacki ?

scytacki:
I think the labbook author should be able to select hidden interactives. 
Here is the scenario:
- an admin loads a ITSI activity in standard authoring mode
- they open a page that has two labbooks and to interactives
- one interactive is hidden
- and their are two labbooks on the page one for each interactive.
- the admin goes to edit one of the labbooks but they choose the one pointing to the hidden interactive
- the admin forgets and hits save instead of cancel
- now this labbook will be pointing at the wrong interactive.

Because of this scenario I think it is safer to allow the author to select even the interactives that are hidden on the page. You could put some text "(hidden)" in the interactive label, to make that more clear to author.

Other than this, the code looks good to me.

knowuh:
@scytacki I added your suggestions, enabling an author to select a hidden interactive for the labbook.  Also provided a warning when an author has specified a snapshot labbook and the interactive isn't visible.

scytacki:
:+1:

PR #82, opened 2015-06-30
Added model id to dropdown list of models [#95983092]

dougmartin:
I also added left side zero padding so the list stays indented at the same level and sorted the list by name like the current editor.


scytacki:
Authors are used to typing in the model number in order to find that particular model. This is because they start at a listing of the models, find one they like, remember the id, then go to the activity and hit the model pull down list.  With zero padding, They will have trouble typing in the id number. They will forget to add zeros for model ids below 100.

Perhaps this can be easily improved using something like chosen: http://harvesthq.github.io/chosen/  Then the id can be put at the end and when an author types an id it will still help them find the model.

dougmartin:
@scytacki I've removed the leading zeros.  If the goal is to keep the new editor looking like the existing editor I think it is the better alternative than adding the chosen jquery plugin.

scytacki:
works for me.

PR #83, opened 2015-06-30
Remove its-dev editor links [#97499816]

dougmartin:
NOTE: Once this is live the existing itsi-dev marked activities with edit as standard mode.  You'll need to set the edit mode to ITSI.  Since this was internal I didn't think a migration was needed.


scytacki:
looks good :+1:

PR #84, opened 2015-06-30
Display labbooks under interactive

knowuh:


ddamelin:
Does this work mean that a lab book *must* be associated with an interactive? The way we are currently using it is to upload user generated images in CODAP would not work with this.

knowuh:
Hmn. I had forgotten that use case. As implemented, a labbook wouldn't show up in single page view if it didn't have an associated interactive element.  @scytacki please advise.

ddamelin:
We could artificially associate it with an interactive, for example, the CODAP launch button, but the purpose is to upload images the user has on their desktop rather than taking a snapshot. Can it still do that? Ultimately, I'd rather add the ability to the draw tool to have the user upload a background image rather than using the lab book, but we are working with the assessment item we have not the one we wished we had. :-)

knowuh:
I can modify this work and the editing work, so that a labbooks can be *explicitly* set as un-connected to an interactive.  In that case it would always show up in the usual assessment item area.

knowuh:
I added some comments to the PT stories to reflect the conversation here:
https://www.pivotaltracker.com/story/show/97734828
https://www.pivotaltracker.com/story/show/97734786

scytacki:
I'll add comments to the story.

knowuh:
OK, the new implementation goes like this:

* If its a single page layout, and the labbook is associated with an interactive, we show it under the interactive.
* If the labbook is snapshot interactive associated with either no-interactive, or a hidden interactive, we don't display the labbook.
* If the Labbook is an 'upload' labbook, then we display it along with the rest of the assessment items.

scytacki:
:+1:

PR #85, opened 2015-07-02
Import activity from portal

sagartalla:
LARA part for this story https://www.pivotaltracker.com/story/show/96186450

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -7,6 +7,11 @@ class ApplicationController < ActionController::Base
   # Run authorization on all actions
   # check_authorization
   protect_from_forgery
+  skip_before_filter :verify_authenticity_token, if: :json_request?
+
+  def json_request?
+    request.format.json?
+  end
 
   # What to do if authorization fails
   rescue_from CanCan::AccessDenied do |exception|
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -7,11 +7,6 @@ class ApplicationController < ActionController::Base
   # Run authorization on all actions
   # check_authorization
   protect_from_forgery
-  skip_before_filter :verify_authenticity_token, if: :json_request?
-
-  def json_request?
-    request.format.json?
-  end
 
   # What to do if authorization fails
   rescue_from CanCan::AccessDenied do |exception|
--- a/app/controllers/import_controller.rb
+++ b/app/controllers/import_controller.rb
@@ -1,5 +1,13 @@
 class ImportController < ApplicationController
 
+  include PeerAccess
+
+  skip_before_filter :verify_authenticity_token, if: :verified_json_request?
+
+  def verified_json_request?
+    verify_request_is_peer
+  end
+
   def import_status
     @message = params[:message] || ''
     respond_to do |format|
@@ -50,6 +58,9 @@ class ImportController < ApplicationController
     if response_code == 201
       response_publish = import_activity.republish_for_portal(portal,req_url)
       response_code = response_publish.code
+    else
+      response.headers["data"] = {:response_code => response_code}.to_json
+      render :nothing => true and return
     end
 
     response_body = JSON.parse "#{response_publish.body}", :symbolize_names => true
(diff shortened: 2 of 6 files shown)


app/controllers/application_controller.rb
```diff
@@ -7,6 +7,11 @@ class ApplicationController < ActionController::Base
   # Run authorization on all actions
   # check_authorization
   protect_from_forgery
+  skip_before_filter :verify_authenticity_token, if: :json_request?
+
+  def json_request?
+    request.format.json?
+  end
```
scytacki: This opens a security hole. Please remove this and work on another way.

app/controllers/import_controller.rb
```diff
@@ -26,5 +26,35 @@ def import
       end
     end
   end
-end
-  
+
+  def import_portal_activity
```
scytacki: How is this authenticated?  Can any anonymous user import a portal activity?

When LARA communicates with the portal it uses an authentication token. This is how the portal can trust LARA. So I think there needs to be something similar here.  The same token can be used.  LARA knows the token and the portal knows the token too.  The portal would use the Client#app_secret for LARA.  This will require adding a new field to Client: `Client#domain`, so we can figure out which Client is LARA.  So the portal looks at the `:authoring_site_url` configuration variable.  Gets the domain (for example `authoring.concord.org`), and then looks for a matching domain in the list of clients.

Then on the LARA side, LARA would look at all of its Concord::AuthPortal models to see if there is one with a matching `AuthorPortal#secret`

Here is how it works in the other direction:
LARA sends a republish request to the portal:
https://github.com/concord-consortium/lara/blob/staging/app/models/publishable.rb#L87
It adds the Bearer token header for the portal, if you trace it back you can see it is getting this token from `Concord::AuthPortal#secret`

And on the portal side this is checked using the PeerAccess module. It starts here:
https://github.com/concord-consortium/rigse/blob/master/app/controllers/external_activities_controller.rb#L14
PeerAccess looks for a `Client` that has a matching `Client#app_secret`, here:
https://github.com/concord-consortium/rigse/blob/master/app/controllers/peer_access.rb#L17

sagartalla: Once we add the `Client#domain`, we might need to change the rake sso:add_client in the portal here https://github.com/concord-consortium/rigse/blob/master/lib/tasks/sso.rake to adapt to the change.
Should we go ahead and do the necessary?

scytacki: I don't see how the user_id would work since I assume the request is coming from the Portal directly. So there would be no cookie sent from the users browser. But perhaps this doesn't work the way I think?

If  the user_id approach is secure then that is fine. But if it is not secure, then yes change the sso:add_client rake task so the domain is asked for. The domain can be optional.

app/controllers/import_controller.rb
```diff
@@ -26,5 +34,39 @@ def import
...
+      user = User.create(
+        email: request_json[:activity][:user_email],
+        password: User.get_random_password,
+        is_author: true
+      )
+    end
+
+    portal = Concord::AuthPortal.portal_for_url(request_json[:portal_url]) 
+    authentication = Authentication.find_by_provider_and_uid(portal.strategy_name,request_json[:domain_uid])
+    user = User.find(authentication.user_id)
```
scytacki: Which user is this? Is the admin that is triggering this import? Or is the author of the activity? If it is the author then this won't work because there won't be an Authentication yet.  If it is the admin, then this will mean the external activity in the portal will be owned by the admin instead of the author.  Perhaps you addressed these issues on the portal side of things.  I'm going to merge this in, but please add a comment here clarifying which user this is since I there are two possibilities.

sagartalla: @scytacki Yes, the user is the admin and he is triggering the import. The publishing code already links the author of the activity in LARA in the portal here https://github.com/concord-consortium/rigse/blob/import_itsi_schools_district_users/lib/activity_runtime_api.rb#L55 , in addition to this we have given the author role to this user(author of the activity and not the importer) in the portal here https://github.com/concord-consortium/rigse/blob/import_itsi_schools_district_users/lib/import_external_activity.rb#L29.

PR #87, opened 2015-07-06
Split up the publishable module into two pieces [#98300946]

dougmartin:
Split publishable model into two pieces in preparation for the changes to publishable for auto publishing.


knowuh:
Looks good.  :+1:

PR #89, opened 2015-07-06
Add after_update observable to publishable models [#98300886]

dougmartin:
This turned out to be very simple.  The model associations for the page and page_items child and grandchildren objects trigger the after_update callback in the lightweight activity for any updates to its contained models.  The "trigger a stubbed something" mentioned in the PT story just outputs a log line.  This will be fleshed out when the auto publishing queuing story is completed.

### DIFF
--- a/app/models/publishable.rb
+++ b/app/models/publishable.rb
@@ -66,4 +66,16 @@ module Publishable
       self.republish_for_portal(portal,self_url)
     end
   end
+
+  ## auto publishing methods
+
+  def self.included(clazz)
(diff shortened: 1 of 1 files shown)


scytacki:
Please change this to `logger.debug` this way it won't fill up the production log (if it gets deployed).  It would be OK to also set the default log level in `config/environments/development.rb` to be debug. That way other devs will see this message. The way to set the log level is shown in a comment of production.rb.

dougmartin:
@scytacki I changed the logger line.

PR #90, opened 2015-07-06
Integrate Spring to LARA Guard. rm guard-jasmin

knowuh:
Simplified working guard setup.  Uses Spring for faster running.

Note: Jasmine must be run from CLI.  But they weren't running in guard before either.


knowuh:
This should make running guard much faster for Devs.  Quick peak anyone? @psndcsrv @scytacki

PR #91, opened 2015-07-07
Added portal publication status checker [#98301304]

dougmartin:
This adds/changes the following:

1. Adds publication_hash (sha1) to the portal_publications, lightweight_activities and lightweight_activity_sequences tables to track the version of the activity/sequence published.
2. Moves the portal_publications has_many association from publication_status.rb to publishable.rb
3. Adds a json endpoint to poll for autopublishing information
4. Adds a React component to the ITSI editor that displays the initial autopublishing information and then polls every 5 seconds for updates to that information.

### DIFF
--- a/app/models/publishable.rb
+++ b/app/models/publishable.rb
@@ -1,4 +1,24 @@
+require 'digest/sha1'
+
 module Publishable
+
+  def latest_publication_portals
+    counts = portal_publications.group(:portal_url).count
+    rows = portal_publications.select("portal_url, success, created_at, publication_hash").where(["id IN (SELECT MAX(id) FROM portal_publications GROUP BY portal_url)"])
+    portals = []
+    rows.each do |row|
+      portals << {
+        :url => row.portal_url,
+        :domain => row.portal_url.gsub(/https?:\/\/([^\/]*).*/){ |x| $1 },
+        :success => row.success,
+        :count => counts[row.portal_url],
+        :publication_hash => row.publication_hash,
+        :date => row.created_at.strftime('%F %R')
+      }
+    end
+    portals
+  end
+
   def find_portal_publication(concord_auth_portal)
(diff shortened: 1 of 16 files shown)


dougmartin:
Note: there were no existing tests for publishing other than checking portal urls.  Since publishing will change with autopublishing I put off adding tests until the autopublishing job story.

dougmartin:
The specs were running fine before I rebased to master.  Looking into the failure now.

dougmartin:
@scytacki @knowuh Could you guy look at this today or tomorrow morning?  Some of the work I'm doing now for the autopublishing delayed job touches these same files.

scytacki:
I'm just starting to look at this. One general note is that we'll need to add this same portal publication status checker to the standard authoring view. It initially looks like it is modular, so that shouldn't be a problem. I just wanted to mention it before I forget.

app/models/publishable.rb
```diff
@@ -72,10 +98,12 @@ def publish_to_portals(self_url)
   def self.included(clazz)
     clazz.class_eval do
       after_update :auto_publish_to_portal
+      has_many :portal_publications, :as => :publishable, :order => :updated_at
     end
   end
 
   def auto_publish_to_portal
+    hash =
```
scytacki: this looks like an error, ruby probably doesn't complain because it sets the value of hash to the return value of logger

scytacki:
This looks to me, I just removed what looks like a typo.
@sagartalla this will require you to make a some changes to:
https://github.com/concord-consortium/lara/pull/85

app/models/publishable.rb
dougmartin: Sorry, that 'hash =' was introduced when I was fixing the merge error I saw after the tests ran on Travis and I was switching branches back and forth and using undo/redo in my editor to apply changes between branches.

PR #92, opened 2015-07-07
Ensure that copied activities retain editor mode and layout.

knowuh:
Small change (only updated a test)

[#98468784]  Test only because the feature was already working when bug
was filed.

https://www.pivotaltracker.com/story/show/98468784


knowuh:
@dougmartin can you take a look. (or @ mention someone to review)

PR #93, opened 2015-07-08
Handle labbook reporting.  [#98300988]

knowuh:
It took me a while to figure out that most of this work was already completed.

The only task to do was to add logic about hiding a labbook for reporting purposes.  This matches the logic for when to display in the runtime.

https://www.pivotaltracker.com/story/show/98300988


knowuh:
Tiny PR,  @dougmartin @scytacki ?

scytacki:
:+1:

PR #94, opened 2015-07-08
Add auto publishing job queue [#98301172]

dougmartin:
This adds a new table that contains queued auto publishing jobs.  The table has a unique index on the portal_url + publishable_id + publishable_type so that only one auto publishing job per (portal, publishable) tuple is allowed.  When at least one new record is successfully added a delayed job is invoked to process the queue.

NOTES:
1. To build the host url for the serialized data in the model the url is stored in the thread when the application starts.
2. This code assumes that the jobs are not processed in parallel by delayed job.
3. The host url is passed to the delayed job.  This works locally since jobs are not delayed in development mode.  Looking at the docs it looks like the params are serialized into the delayed_jobs table but I was not able to test that locally.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -17,6 +17,7 @@ class ApplicationController < ActionController::Base
   before_filter :portal_login
   before_filter :reject_old_browsers, :except => [:bad_browser]
   before_filter :set_locale
+  before_filter :store_auto_publish_url # to enable auto publishing to build an url from the request object
   after_filter :log_session_after
 
   # Try to set local from the request headers
@@ -235,7 +236,7 @@ class ApplicationController < ActionController::Base
       redirect_to user_omniauth_authorize_path(strategy_name, :origin => uri.to_s)
     end
   end
-  
+
   # check if any of them have a matching domain and uid
   def has_matching_domain_uid(domain,domain_uid)
     return current_user && current_user.authentications.where(provider:Concord::AuthPortal.strategy_name_for_url(params[:domain]),uid:domain_uid).length > 0
@@ -247,8 +248,8 @@ class ApplicationController < ActionController::Base
     clear_session_response_key
     request.env['omniauth.origin'] || stored_location_for(resource) || signed_in_root_path(resource)
   end
-  
-  
+
+
   def after_sign_out_path_for(resource)
     if params[:re_login] && params[:user_provider]
       provider = params[:user_provider]
@@ -260,9 +261,9 @@ class ApplicationController < ActionController::Base
         :re_login => true,
         :redirect_uri => redirect_url,
         :provider => provider
-      }  
+      }
     end
-    params[:re_login] && params[:user_provider] ? "#{Concord::AuthPortal.url_for_portal(provider_id)}users/sign_out?#{params_hash.to_query}" : root_path 
+    params[:re_login] && params[:user_provider] ? "#{Concord::AuthPortal.url_for_portal(provider_id)}users/sign_out?#{params_hash.to_query}" : root_path
   end
 
   def respond_with_edit_form
@@ -322,5 +323,9 @@ class ApplicationController < ActionController::Base
   def handle_unverified_request
     logger.info("unverified_request session:" + session.inspect)
     reset_session
+  end
+
+  def store_auto_publish_url
+    Thread.current[:auto_publish_url] = "#{request.protocol}#{request.host_with_port}"
   end
 end
--- a/app/models/process_pending_portal_publication.rb
+++ b/app/models/process_pending_portal_publication.rb
@@ -1,0 +1,40 @@
+class ProcessPendingPortalPublication < Struct.new(:pending_portal_publication_id, :auto_publish_url, :backoff)
+  def perform
+    # immediately find and remove the pending publication so that new publications can be queued
+    pending_portal_publication = PendingPortalPublication.find(pending_portal_publication_id)
+    return if pending_portal_publication.nil?
+    pending_portal_publication.delete
+
+    # get shorter local references
+    portal_publication = pending_portal_publication.portal_publication
+    return if portal_publication.nil?
+    publishable = portal_publication.publishable
+    return if publishable.nil?
+    portal = Concord::AuthPortal.portal_for_publishing_url(portal_publication.portal_url)
+    last_portal_publication = publishable.last_publication(portal)
+    return if last_portal_publication.nil?
+
+    # if the portal publication is not the latest one another job has published behind us so we are done
+    return if portal_publication.id != last_portal_publication.id
+
+    # skip autopublish if the hash hasn't changed since the last publish
+    json = publishable.serialize_for_portal(auto_publish_url).to_json
+    return if last_portal_publication.publication_hash == Digest::SHA1.hexdigest(json)
+
+    # publish to the portal
+    response = publishable.republish_for_portal(portal, auto_publish_url, json)
+    unless response.code == 201
+
+      # if it fails requeue with an exponentially increased backoff time with a max of 3 hours or 10800 seconds
+      # will actually stop at (2^11)*5=10240 seconds or 2.84 hours
+      new_backoff = backoff * 2
+      if new_backoff * publishable.auto_publish_delay < 3*60*60
+        publishable.queue_auto_publish_to_portal(auto_publish_url, new_backoff)
+      end
+    end
+  end
+
+  def debug(text)
+    Delayed::Worker.logger.add(Logger::DEBUG, text) if Delayed::Worker.logger
+  end
+end
--- a/app/models/process_pending_portal_publication.rb
+++ b/app/models/process_pending_portal_publication.rb
@@ -11,7 +11,7 @@ class ProcessPendingPortalPublication < Struct.new(:pending_portal_publication_i
     publishable = portal_publication.publishable
     return if publishable.nil?
     portal = Concord::AuthPortal.portal_for_publishing_url(portal_publication.portal_url)
-    last_portal_publication = publishable.last_publication(portal)
+    last_portal_publication = publishable.last_successful_publication(portal)
     return if last_portal_publication.nil?
 
     # if the portal publication is not the latest one another job has published behind us so we are done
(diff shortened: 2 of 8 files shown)


scytacki:
We've run into problems in the past using handle_async on model objects. That approach ends up serializing the whole model object into the delayed job table. At least in the portal we make special "Job" objects that are serialized by delayed job.

scytacki:
I see in LARA we are *not* taking this clean approach: https://github.com/concord-consortium/lara/blob/staging/app/models/run.rb#L260

dougmartin:
Yes, I based this on the run.rb code

app/controllers/application_controller.rb
```diff
@@ -323,4 +324,8 @@ def handle_unverified_request
     logger.info("unverified_request session:" + session.inspect)
     reset_session
   end
+
+  def store_request_in_thread
+    Thread.current[:request] = request
```
knowuh: One comment here.  Maybe instead of saving the whole request object, we could just save the data we need, under an well named key.  I feel like Thread.current[:request] could very well be something important we don't want to mess with. At least I would assume that when reading it out of context.  Like maybe  `Thread.current[:saved_server_url]="#{request.protocol}#{request.host_with_port}"` ?

dougmartin: Good idea.  I'll update it.

dougmartin:
Would it be better to make the queue processor a class method so that it doesn't have to serialize the object values?

scytacki:
I think using a custom object will encapsulate the state and logic better. I'm still looking through this to understand how it works. Here is an example from the portal: https://github.com/concord-consortium/rigse/blob/master/app/models/dataservice/bundle_content_observer.rb#L9 and
https://github.com/concord-consortium/rigse/blob/master/app/models/dataservice/process_bundle_job.rb

scytacki:
We can't count on the no parallel job processing. LARA will likely be running on multiple servers and with our current setup that means multiple job workers.

app/models/publishable.rb
```diff
@@ -97,12 +101,66 @@ def publish_to_portals(self_url)
...
+          # immediately remove the job so that new edits can be queued
+          PortalPublicationJob.find(job.id).delete
+
+          # get the json and store it in a local cache in case there are multiple portals
+          cache_key = "#{job.publishable.class.name}:#{job.publishable.id}"
+          json = json_cache[cache_key] || job.publishable.serialize_for_portal(self_url).to_json
+          json_cache[cache_key] = json
+
+          # skip autopublish if the hash hasn't changed since the last publish
+          return if publication_hash == Digest::SHA1.hexdigest(json)
```
scytacki: Where is this publication_hash coming from? It seems it needs to come from the last PortalPublication for this combination of portal and publishable.

scytacki:
This might be out of date now, that just updated. But here are my comments: 

`PortalPublicationJob` makes me think this is actually the job object that is being processed by delayed job. Perhaps PendingPortalPublication would be more clear.

Using a separate Job struct that has a perform method that fires off the publication, should make the saved state more clear and also make it possible encapsulate the code working with this state.

It also seems more clear to have separate Jobs for each portal being published too.  This way with multiple workers we'll be able to publish to several portals at the same time. However I see you are optimizing the json generation. For now I'd remove that optimization, most activities are only published to a single portal.  If we need to, we can find a way to optimize that and keep separate jobs for each portal.

Also the current approach might be broken, seem my comment about the publication_hash.

With separate delayed jobs for each portal you could possibly store the last publication_hash (from portal_publications) in the delayed job object.  But probably it is better to look this up while processing, because that will be one more check against sending duplicate publications to the same portal (in case multiple jobs get queued for the same portal somehow).

dougmartin:
You are right.  This code was outside of the loop initially when I wasn't taking into consideration that the queue processor might pull in multiple activities - when i moved it into the loop I neglected to update the reference to use the object loaded in the loop.

dougmartin:
Ok - I've updated the code to do the following:

1. Use 1 job per portal/activity 
2. Changed the comparison to use the last portal_publication publication_hash for the portal instead of using the publication_hash of the activity/sequence.

app/models/publishable.rb
```diff
@@ -97,12 +101,53 @@ def publish_to_portals(self_url)
...
+            # if the job saved then process it
+            process_auto_publish_to_portal_job(job.id, Thread.current[:auto_publish_url])
+
+          rescue ActiveRecord::StatementInvalid => e
+            raise e unless /Duplicate entry/.match(e.to_s)
+          end
+        end
+      end
+
+      def process_auto_publish_to_portal_job(job_id, self_url)
```
knowuh: This method is looking good to me now.  :+1:

dougmartin:
The code now uses a serialized struct (app/models/process_portal_publication_job.rb) to perform the publication job.  It is serialized as the following in the delayed_jobs table:

    --- !ruby/struct:ProcessAutoPublishJob
    job_id: 7
    self_url: http://railsdev:3000

I wasn't seeing it serialize before because I failed to restart the Rails server after I made the config change.

scytacki:
This is looking really good.

I just thought of another failure mode. If the republish fails because the portal is unavailable, then we will be in an unpublished state. Delayed Job has built in support for retrying jobs. I think this happens if the job throws an exception, but there is probably another way to mark the job and failed.  I imagine that republish might throw an exception if the Portal is unavailable. However in the current code, the PortalPublicationJob will have been deleted, so if Delayed Job tries to run ProcessPortalPublicationJob again, it will bailout when it can't find a PortalPublicationJob. It is tempting to not deleted PortalPublicationJob until a successful publish, then this would fix one problem and cause another: We'd miss activity changes that came in between the serialize_for_portal call and when the republish completes successfully. 

A possible solution is for ProcessPortalPublicationJob to recreate the PortalPublicationJob object if there is a failure publishing to the portal.  This will prevent duplicate ProcessPortalPublicationJob while we wait for Delayed Job to try running it again. There might be a duplicate that sneaks in while the serialize and republish is happening. For example the author changes something during this time. That scenario can be handled by bailing the Job and not retrying if ProcessPortalPublicationJob fails to create a new PortalPublicationJob.  I haven't thought through all of the race cases though, so perhaps there is a case where this solution will fail.

dougmartin:
The current portal publish returns false is the portal doesn't return the 201 http status code.  We can use that for the retry.

I think the best way to handle publishing failure is to add the id of the last portal publication to the ProcessPortalPublicationJob parameters.  We can still delete the job that way and then on a retry if the job is deleted we can lookup the current last portal publication and compare it to the passed portal publication id - if they are the same it means no other job has run so we can retry publishing.  If the passed id points to an older portal publication it means we can skip publishing.  The portal publication also has the publishable polymorphic property so we don't need the one in the portal_publications_jobs table.

Thinking this through I think the portal_publications_jobs table could be renamed/restructured as republishable_portal_publications which would simply point to the id of portal_publication to republish - the rest of the current data is redundant.  We can still use the table as the throttling mechanism for adding jobs but we would use the portal_publications table as both the source of the publishable item and if it is the latest published version.

scytacki:
Sounds good, and removing duplicate information is always good.

app/controllers/application_controller.rb
```diff
@@ -323,4 +324,8 @@ def handle_unverified_request
     logger.info("unverified_request session:" + session.inspect)
     reset_session
   end
+
+  def store_auto_publish_url
+    Thread.current[:auto_publish_url] = "#{request.protocol}#{request.host_with_port}"
```
knowuh: @scytacki should we use a protocol relative URL here?  This is called the `auto_publish_url`, but its actually the authoring server, eventually It's used in `lightweight_activity.rb`: 

     def serialize_for_portal(host)
        local_url = "#{host}#{Rails.application.routes.url_helpers.activity_path(self)}"
        …
     end

scytacki: I think we want to include the protocol, if we use protocol relative then an http portal will always be stuck running http activities.  Also I think this URL gets used as a redirect so I'm not sure how redirects handle protocol relative. Many of our activities are https compatible so it would make sense to be able to publish those as https even though the portal is running http.  I haven't worked through the cases though so there might be gotchas there.  If there aren't gotchas then it would be best if we had a flag (probably hidden) on the activity that indicated if it was https compatible and we use that to determine what protocol to use when publishing the activity. But that should be another PR not this one.

dougmartin:
This has the rewritten to so that the throttle table (renamed pending_portal_publications) just references the id of the portal_publications table. An exponential backoff has also been added with a max of 3 hours in case of publishing failure.  On failure the job is attempted to be re-enqueued into the pending_portal_publications table and then the delayed job time is set based on the backoff.  This was used instead of using delayed_job's retry so that we don't have race conditions between a new publish and a retrying publish.

I did not change the auto_publish_url as it needs to match the url generated by publications_controller::add_portal:

      def add_portal
          # TODO: we must not forget to actually publish this later, see publis_activity_path(activity)
          @portal = find_portal
          @message = ''
          req_url = "#{request.protocol}#{request.host_with_port}"
          @publishable.portal_publish(current_user,@portal,req_url)
          redirect_to :action => 'show_status'
        end

app/models/process_pending_portal_publication.rb
```diff
@@ -0,0 +1,40 @@
...
+    return if portal_publication.nil?
+    publishable = portal_publication.publishable
+    return if publishable.nil?
+    portal = Concord::AuthPortal.portal_for_publishing_url(portal_publication.portal_url)
+    last_portal_publication = publishable.last_publication(portal)
+    return if last_portal_publication.nil?
+
+    # if the portal publication is not the latest one another job has published behind us so we are done
+    return if portal_publication.id != last_portal_publication.id
+
```
scytacki: I think this should find the last successful publication.  There might be a failed publication that occurs in between.

scytacki:
With the exceptions of the 'successful publication' issue. This is good to go.
Tests would be great to document all the logic and edge cases. That might require a little refactoring to make testing easier. But that is another PR.

dougmartin:
@scytacki I pushed an update a few minutes after you commented.  It's all ready.

scytacki:
@dougmartin I'm not seeing code that makes sure `publishable.last_publication(portal)` returns the last successful publication. It might return a failed publication. In which case ProcessPendingPortalPublication will exit even though it should continue. Maybe I'm still seeing old code somehow.

scytacki:
I went ahead and added this. I didn't test it locally though, but it seems simple enough (famous last words). This really does need tests, but I know they are going to be painful.

scytacki:
And now that I'm testing it I see that it was right before. I've also found another issue that I'm working through. I'll create a new PR that includes both fixes.

PR #95, opened 2015-07-08
Serialize labbooks (½): Handles a user directly copying an activity.

knowuh:


knowuh:
Note:  This is based off labbook reporting branch. It will be easier to review once the other PR is merged.

app/services/page_serialization_helper.rb
```diff
@@ -0,0 +1,14 @@
+class PageSerializationHelper
+  def chache_interactive_copy(original_interactive, copy_of_interactive)
```
scytacki: chache - cache

knowuh: Thanks for catching that. -- I will roll that into the next set of changes.

dougmartin: Noah loves Chache?

scytacki:
This looks good to me. As we talked about, it might make sense in the future to add a more complex system for handling references when copying/import/export handling of references.

PR #96, opened 2015-07-09
Serialize labbooks (import / export / copy)

knowuh:
This finishes earlier work.  Unfortunately I renamed two classes, and because of how I did it, git missed the renames, and they appear as a deleted & added here.

Finishes [#98605752] -labbook & interactive import


knowuh:
@scytacki since you have been following this story, maybe you want to look at the latest code.  Or @ mention some one else to delegate.

PR #97, opened 2015-07-09
Serialize labbooks2

knowuh:
Finishes work on copy / import / export for labbooks.


app/services/lara_serialization_helper.rb
```diff
@@ -0,0 +1,53 @@
...
+    @entries[key_sting]
+  end
+
+  def lookup_new_item(original_item)
+    lookup_item(key(original_item))
+  end
+
+  def key(interactive)
+    "#{interactive.id}-#{interactive.class.name}"
+  end
```
scytacki: Since this now seems generic `interactive` should be renamed. Perhaps `item` to follow the argument names below.

scytacki:
:+1: this looks great to me. I'd just do the rename above.  I can also seen argument for keeping it interactive since that is all that is actually passed to it right now. Feel free to merge this.

PR #100, opened 2015-07-10
Preserve hidden attributes on embeddables and interactives on copy & export

knowuh:
Fixes https://www.pivotaltracker.com/story/show/98560552 

Note:  We need to factor-out common attributes for
embeddables and interactives.

Also we need to unify serialization methods: duplicate, export,  to_hash, serialize_for_portal (!)

Making this small change for serializing visibility cut across too many files and too many specs.

I was tempted to address, but ran out of time.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -240,8 +240,8 @@ ActiveRecord::Schema.define(:version => 20150706174054) do
     t.boolean  "is_prediction",            :default => false
     t.boolean  "give_prediction_feedback", :default => false
     t.text     "prediction_feedback"
-    t.boolean  "is_hidden",                :default => false
     t.string   "default_text"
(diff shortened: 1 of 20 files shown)


knowuh:
@psndcsrv or @scytacki  have a look?  Its kind of tedious repetitive code, a shotgun changeset, where all embeddables and all interactives had their to_hash and export methods changed.

db/schema.rb
```diff
@@ -240,8 +240,8 @@
     t.boolean  "is_prediction",            :default => false
```
knowuh: This schema change was checked in 'by accident' -- but was never edited by hand, and probably reflects that someone else didn't check it in earlier?

scytacki:
:+1:

PR #103, opened 2015-07-13
only set attributes that are set in the import

scytacki:
This seems generally like a better approach then what was there before.

Additionally it fixes an issue when importing pages that don’t have an is_hidden
field. In that case the is_hidden attribute was set to nil, which isn’t
valid. If the attribute is just skipped, then the default value for the 
attribute is used.


scytacki:
@dougmartin can you take a look when you get a chance?

dougmartin:
Looks good.  :+1:

PR #105, opened 2015-07-15
Fixed publish header in sequence edit broken by new react publishing details component

dougmartin:


scytacki:
Can you update the naming here? It seems the ITSIAuthoring::PublicationDetails is no longer specific to ITSIAuthoring, and it isn't specific to working with activities.  Perhaps `publishable` instead of `activity` would work. That is just based on a quick skim so perhaps I'm missing some context.

dougmartin:
Sure - updating now.

dougmartin:
@scytacki I've moved the service out of itsi_authoring into the root of the services folder.

scytacki:
:+1:

PR #106, opened 2015-07-15
duplicate can publish back to a portal

scytacki:
When passed a URL like:
http://localhost:3000/activities/8/duplicate?domain=http://portal.local/&domain_uid=2&add_to_portal=http://portal.local/

LARA will first log in the user in the portal, then duplicate the activity, and then publish it back to the portal.  This pull request also includes a few changes to some existing spec tests of the duplicate code.

[#97737250]


scytacki:
@dougmartin can you look at this?

dougmartin:
Looks good :+1:

PR #108, opened 2015-07-15
Add trigger the autopublishing of sequences when activities change [#99005644]

dougmartin:
The :touch option does not work for has_many associations so I had to do it manually in the queing code.


scytacki:
:+1: Good enough for now.

PR #110, opened 2015-07-16
publish with the correct authentication_token

scytacki:
Without this, the wrong authentication toke was being used when a user
was logging into two different portal and then trying to add the material
to the first portal.
[#99167076]

### DIFF
--- a/app/models/publishable.rb
+++ b/app/models/publishable.rb
@@ -35,7 +35,7 @@ module Publishable
     end
 
     self.update_attribute('publication_status','public')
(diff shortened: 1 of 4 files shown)


scytacki:
@dougmartin can you look at this?

scytacki:
oops failing tests.  please hold.

dougmartin:
Looks good :+1:

scytacki:
After merging another PR the test for this one broke. That should be fixed now, just waiting for Travis confirmation.

PR #112, opened 2015-07-16
Remove default prompt for imported drawing response questions[#98561666]

sagartalla:
https://www.pivotaltracker.com/story/show/98561666

### DIFF
--- a/app/models/embeddable/image_question.rb
+++ b/app/models/embeddable/image_question.rb
@@ -55,7 +55,10 @@ class Embeddable::ImageQuestion < ActiveRecord::Base
   end
   
   def self.import(import_hash)
-    return self.new(import_hash)
+    new_ques = self.new(import_hash)
+    new_ques.prompt = ""
+    new_ques.save!
+    return new_ques
   end
   
   def is_shutterbug?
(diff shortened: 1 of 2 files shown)


scytacki:
Can you add a spec test to explain this more?
Is the prompt being set as a default value in LARA. Or is the prompt being set by ITSI?
If the prompt is being set by ITSI then it should be fixed there.
If the prompt is being set by LARA, then instead of clearing the prompt in the controller, the import method should clear the prompt only if the imported json doesn't specify a prompt.

This current code would clear the prompt all of the time which would prevent a legitimate prompt from being imported.

scytacki:
Oops I realize I didn't look at this close enough.  So this is already in the model. But most of the comments above are still relevant:
If the prompt is being set by ITSI then it should be fixed there.
If the prompt is being set as a default by LARA, then instead of always clearing the prompt, the import method should clear the prompt only if the imported json doesn't specify a prompt.

sagartalla:
@scytacki ITSI is not setting the prompt in the export. Now the code clears only the default prompt set by LARA i.e the import method clears the prompt only if the imported json doesn't specify a prompt.

scytacki:
:+1:

PR #113, opened 2015-07-16
Update ITSI model editor to use live model export [#98224298]

dougmartin:
This also adds a ENV variable for the prediction list.  Both ENV variables fall back to a S3 url.


scytacki:
:+1:

PR #115, opened 2015-07-20
Remove banner from itsi editor when iframed and adjust height [#97848200] [#99173350]

dougmartin:
This runs Javascript right after the body tag is defined in application.html.erb that addes an 'iframed' class to the body if the page is not the top document.  The body.iframed class then removes unneeded elements from the page like the header and footer.  It then waits for the parent document to announce it needs iframe_site_manager services via PostMessage and then handles those requests.  Currently the only requests are the the iframe height  [#99173350] but other stories in the current PT may require other types of lara-iframed communication.


knowuh:
:+1:

PR #117, opened 2015-07-24
Copying an activity redirects to the iframed editor [#99311988]

dougmartin:
Note: there is a parallel rigse PR to support this feature (https://github.com/concord-consortium/rigse/pull/114)

### DIFF
--- a/app/controllers/lightweight_activities_controller.rb
+++ b/app/controllers/lightweight_activities_controller.rb
@@ -169,12 +169,21 @@ class LightweightActivitiesController < ApplicationController
 
     if @new_activity.save(:validations => false) # In case the old activity was invalid
       # check if we should publish this new activity somewhere
+      json_response = nil
       if params['add_to_portal']
         req_url = "#{request.protocol}#{request.host_with_port}"
         # this might take a little time so it might be better do this in the background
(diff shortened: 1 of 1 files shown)


app/controllers/lightweight_activities_controller.rb
```diff
@@ -169,12 +169,25 @@ def duplicate
...
+        begin
+          json_response = JSON.parse(response.body)
+        rescue
+        end
+      end
+      if params['redirect_on_success']
+        if params['redirect_on_success'].include?(':activity_id') && json_response && json_response.has_key?("activity_id")
+          redirect_to params['redirect_on_success'].sub!(':activity_id', json_response['activity_id'].to_s)
+        else
+          redirect_to params['redirect_on_success']
```
knowuh: Seems like we could redirect to the portal, even if the publication wasn't successful.  How do we find out if we failed?

dougmartin:
Good point.  I simplified the code since we know that the redirect url will always have the :activity_id template variable.  So on a publish failure then the LARA edit page is shown which shows the publishing failure - there is no way to iframe the failed publish so it is the best we can do.

dougmartin:
@knowuh Do you have any other issues with this PR?

knowuh:
Looks good.

PR #118, opened 2015-07-27
Added json response to unauthorized ajax calls

dougmartin:
Currently unauthorized ajax calls return a 500 error because the application controller does not handle json formatted errors for unauthorized requests (it tries to load the shared/unauthorized partial in json format).  This change is needed in another story (https://www.pivotaltracker.com/story/show/98616438) so that the client side ajax can detect the unauthorized request and present a better error message.  I thought it best to split PR from that story so that it stands alone.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -10,7 +10,11 @@ class ApplicationController < ActionController::Base
 
   # What to do if authorization fails
   rescue_from CanCan::AccessDenied do |exception|
-    render :partial => "shared/unauthorized", :locals => {:action => exception.action,:resource=> exception.subject},:status => 403
+    response_data = unauthorized_response_data(exception.action, exception.subject)
+    respond_to do |format|
+      format.html { render :partial => "shared/unauthorized", :locals => {:message => response_data[:message]}, :status => 403 }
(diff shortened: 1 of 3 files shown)


app/controllers/application_controller.rb
```diff
@@ -10,7 +10,11 @@ class ApplicationController < ActionController::Base
 
   # What to do if authorization fails
   rescue_from CanCan::AccessDenied do |exception|
-    render :partial => "shared/unauthorized", :locals => {:action => exception.action,:resource=> exception.subject},:status => 403
+    response_data = unauthorized_response_data(exception.action, exception.subject)
+    respond_to do |format|
+      format.html { render :partial => "shared/unauthorized", :locals => response_data, :status => 403 }
```
scytacki: it would be more clear if the the `unauthorized_response_data`, was assigned to a local like: `unauthorized_info`, so this would change the use of `message` below to `unauthorized_info.message`.

Without this, a developer might add new keys to the unauthorized_response_data and not realize this is injecting new (possibly conflicting) local variables in the haml partial.

dougmartin:
@scytacki I've updated the code to explicitly pass in the message since it is the only thing used in the template.

scytacki:
:+1:

PR #120, opened 2015-07-29
Added imported_activity_url to activities [#100064630]

dougmartin:


scytacki:
this caused from tests to fail.
Other than that it looks good to go.

PR #126, opened 2015-09-09
Fix single page interactive sizing

dougmartin:


dougmartin:
NOTE: The width/height of the predictions is also saved, however there is no width/height in the current prediction json list from S3 so that code is a no-op for now (the React component only posts changed values, since the width/height never change from undefined when switching predictions they are not included in the posted data).

app/helpers/interactive_run_helper.rb
```diff
@@ -56,15 +56,17 @@ def interactive_iframe_tag(interactive,run=nil,iframe_src=true)
       url = interactive_run.learner_url
     end
     url = interactive.url if url.blank?
-    
+
     data = {
       :aspect_ratio => interactive.aspect_ratio,
       :id => interactive.id,
-      :iframe_mouseover => "false"
+      :iframe_mouseover => "false",
```
scytacki: this looks like a typo

dougmartin: During development I had some other data in there so I added the comma.  I'll remove it.

dougmartin:
@scytacki I've removed the stray comma.

scytacki:
:+1:

PR #127, opened 2015-09-09
Changed ITSI rich text editor to TinyMCE [#102699432]

dougmartin:


scytacki:
Did you look to see if you could get a rails engine gem that provided the assets of tinymce? Then all of the javascript and images wouldn't need to be included.

dougmartin:
@scytacki I've switched to the gem.

scytacki:
This doesn't seem to include the changes to the other rich text boxes in LARA, but I think that is an OK place to start out. :+1:

PR #129, opened 2015-09-11
Removed paste as text and code icons

dougmartin:


dougmartin:
@scytacki I've removed the toolbar icons.

PR #130, opened 2015-09-13
Interactive sizing

scytacki:
This fixes a bug when multiple interactives are on the same page and have different aspect ratios.
It also fixes another bug when the click to play image for an interactive has a different aspect ratio than the interactive. Previously the image could extend down the page.
Finally it adds a shortcut for some display problems. Only interactives with a height greater than 1 are render. Otherwise they are not rendered at all.  

For ITSI we have some interactives that are fake interactives, which just hold a spot in the template system. Sizing them with a height of 1 or 0, didn't work well because the click to play text and area were not limited to this height.  With this change none of that is rendered so the interactive is complete invisible at runtime.

### DIFF
--- a/app/assets/stylesheets/partials/_click_to_play.scss
+++ b/app/assets/stylesheets/partials/_click_to_play.scss
@@ -10,6 +10,8 @@
     .img-container {
       img {
         width : 100%;
+        height: 100%;
+        object-fit: contain;
       }
     }
     .click_to_play {
--- a/app/assets/javascripts/interactive-sizing.js
+++ b/app/assets/javascripts/interactive-sizing.js
@@ -19,6 +19,8 @@ ResizableInteractive.prototype.fixSize = function () {
 
 // Setup
 $(document).ready(function () {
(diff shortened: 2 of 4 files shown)


app/assets/stylesheets/partials/_click_to_play.scss
```diff
@@ -10,6 +10,8 @@
     .img-container {
       img {
         width : 100%;
+        height: 100%;
+        object-fit: contain;
```
dougmartin: I never heard of this property until now :) It doesn't look like it is supported in IE though.

dougmartin:
Looks good.

app/assets/stylesheets/partials/_click_to_play.scss
scytacki: I knew it existed but didn't know its name or how to use it until yesterday. Yeah in IE things are going to be stretched but normally it should be minor since people ought to be matching the aspect ratio of the image to the interactive.

PR #131, opened 2015-09-14
add toggling metadata editor to match rest of style

scytacki:
Here is the PT story:
https://www.pivotaltracker.com/story/show/103014778

This also changes overall padding of most elements on the page.  It didn't seem like we needed all of that padding.


dougmartin:
Looks good :+1:

scytacki:
Thanks for reviewing these.

PR #132, opened 2015-09-14
Tinymce improvements

scytacki:
- auto resize the editor
- increase font size in editor
- fix some UL spacing broken by recent ul change


scytacki:
I'm going ahead and merging this in, so Evangeline can test it on staging.  @dougmartin if you have time, please still take a look to see if there is something I could be doing better here.

This use of a `:javascript` section seemed like the only way to have access to `asset_path` while setting the options for tinymce in ITSI. `asset_path` is needed to get the correct hashed path of the compiled asset.  I also tried the tinymce helper but it just called `tinymce.init` which didn't affect how tinymce was being used in the richtext editor.

dougmartin:
Looks good to me.

PR #133, opened 2015-09-15
use json response with json request

scytacki:
This fixes an error with updated ITSI editing of activity metadata.
[#103014778]

### DIFF
--- a/app/assets/javascripts/components/itsi_authoring/section_element_editor_mixin.js.coffee
+++ b/app/assets/javascripts/components/itsi_authoring/section_element_editor_mixin.js.coffee
@@ -61,7 +61,7 @@ modulejs.define 'components/itsi_authoring/section_element_editor_mixin',
       success: =>
         @props.alert 'info', 'Saved'
         @setState changedValues: {}
-      error: =>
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/components/itsi_authoring/section_element_editor_mixin.js.coffee
```diff
@@ -61,7 +61,7 @@ modulejs.define 'components/itsi_authoring/section_element_editor_mixin',
       success: =>
         @props.alert 'info', 'Saved'
         @setState changedValues: {}
-      error: =>
+      error: (xhr, textStatus, errorThrown) =>
         @props.alert 'error', 'Save Failed!'
```
dougmartin: Do you want to return the actual error message you return with @activity.errors in the controller here?

scytacki: I didn't want to go that far. I added the extra arguments so I could figure out was going on using the javascript debugger. I'd need more time to spend on this to check if the error message is useful to display and to display it in a more persistent way.

PR #134, opened 2015-10-13
Fix  LARA numbering when interactives save state, but don't report.

knowuh:
Fix bug where question numbers in the report and the LARA runtime get out of sync following the page that contains the state saving interactive.

This commit is small, it just changes how 'questions' are calculated by activities.  The rest of the commit is just testing that method.

## Before:##
image

## After: ##
image

https://www.pivotaltracker.com/story/show/83009202

### DIFF
--- a/app/models/lightweight_activity.rb
+++ b/app/models/lightweight_activity.rb
@@ -254,7 +254,7 @@ class LightweightActivity < ActiveRecord::Base
           }
           elements.push(mc_data)
         when MwInteractive
-          if embeddable.save_state && embeddable.has_report_url
+          if embeddable.is_reportable
             iframe_data = embeddable.to_hash
             iframe_data["type"] = 'iframe_interactive'
             iframe_data["id"] = embeddable.id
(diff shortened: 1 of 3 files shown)


knowuh:
@scytacki  small commit to review if you have a second.

scytacki:
Looks good to me, but I'm curious how the code that sends the questions to the portal works. Does it use this same `questions` method on LightweightActivity? Or does it have its own logic for selecting which questions to send?  Based on the behavior I suspect this logic is duplicated in 2 places.

knowuh:
It definitely is duplicated.  I decided not to touch it, but I can revisit that.  Here is what is in serialize_for_portal

    def serialize_for_portal(host)
       …
       when MwInteractive
          if embeddable.save_state && embeddable.has_report_url
       …
  

Probably worth at least using the same method  `is_reportable` above.  The rest of that method is also pretty ugly, but I didn't wade into that minefield.

scytacki:
yeah using the is_reportable seems good enough for now.

PR #136, opened 2015-10-26
Vagrant docker

knowuh:
@dougmartin  if you want to give this a try...


knowuh:
@scytacki   You can try this on your mac too.  I am going to continue to try and tweak it.  @dougmartin has been testing it on his win setup, and this solution should work for both platforms.

scytacki:
I'm giving this a try. The downloading of `ubuntu/trusty64` is very slow. The current estimate for me is 5.5 hours. Usually my home internet is fast, so I suspect this is problem with the server.

knowuh:
@dougmartin @scytacki  – I have discovered a few bugs locally.  Sorry to have dragged you in a bit early.  Hang tight.

scytacki:
I had to run `vagrant plugin install vagrant-docker-compose` before `vagrant up` worked. So that should be added to the readme (if it is the right thing to do).  After doing that `vagrant up`, seemed to work.

There was a very long pause with this line in the output `Installing Docker (latest) onto machine...` I didn't time it, but it might have been an hour or so. I guess at least the user could be warned in the readme. Or perhaps if that step can have a progress indicator or something that would be better. 

I'll wait until Noah gives the go ahead before continuing. I'm hoping that the large downloads have been done and cached so the process will be faster next time even if I start from scratch.

knowuh:
There are still improvements I would like to make, but I have found your early feedback helpful. Thanks for taking the time.

New documentation can be found here: https://github.com/concord-consortium/lara/blob/vagrant-docker/documentation/vagrant-docker.md

knowuh:
After you pull the latest changes: 

1. Copy / edit / create  the file  `config/app_environment_variables.rb`
2. `vagrant reload --provision`
3. `vagrant ssh`
4. `cd /lara && docker-compose up`

scytacki:
OK I got this working.  
What would be the recommended way to run rails commands like rake or the rails console?

What I've come up with so far is:
- run `vagrant ssh` and `cd /lara`, and then use this terminal open
- then to run a command: `docker-compose run app [cmd]`

I also tried running `docker-compose run app bash`, so I could just have a terminal open all of the time in the app environment. This didn't work, because of some environment setup issues I'm guessing. When I run a rails command it complains of not being able to connect to the database.

It would be nice if there was an easy and fast way to run rake commands and startup the rails console.

Also I was confused for a while by the way `docker-compose up` seems to print out old log messages when it starts up.

knowuh:
This should be ready for a merge into master now.  @scytacki  have a look (only changed documentation mostly)

scytacki:
:+1:

PR #138, opened 2015-11-16
handle the missing template exceptions

scytacki:
We are getting a number of missing template exceptions. In my opinion, it is not important to know if someone requests a format of a route that we don't support.

The most common example is when Edmodo tries to request the `image/*` version of a activity. 

@knowuh since you probably care the most about watching exceptions, do you agree?

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -17,6 +17,15 @@ class ApplicationController < ActionController::Base
     end
   end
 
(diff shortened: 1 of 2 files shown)


scytacki:
@knowuh I finally fixed the failing test caused by the initial change. Any comments on this one?
I'd like to try the same approach in the Portal to cut down all of the exceptions we've been seeing for the same reason.

knowuh:
A 404 page seems like a better default.   :+1:

PR #139, opened 2015-11-16
Don’t save objects to DB before authorizing

scytacki:
This was allowing some ownerless objects into the database.

@knowuh can you take a look?


knowuh:
:1 looks good.

PR #140, opened 2015-12-02
Support new format of logs sent from Lab

pjanik:
Lab PR:
https://github.com/concord-consortium/lab/pull/105

This PR ensures that both RPC-style and new format is supported. When Lab is released, we can remove the old code.


sfentress:
:+1:

PR #141, opened 2015-12-22
downgrade the version of aws-sdk

scytacki:
Hopefully this will fix an issue with the capistrono-autoscaler gem


scytacki:
It is just a dependency change so there isn't much code to review. It is mainly just testing that it doesn't break the runtime, so I'm going to merge and deploy to staging.

PR #142, opened 2016-01-13
Dashboard report

pjanik:
Two new API endpoints:
1. `/activities/:id/dashboard_toc`
2. `/runs/dashboard`

I've used a draft provided by @knowuh (action names, routes, code organization). `dashboard_toc` hasn't been changed much, I've removed embeddables listing and provided arg block questions instead. `runs/dashboard` has been totally reimplemented.


spec/services/dashboard_runlist_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'spec_helper'
+
+describe DashboardRunlist do
+  include_context "activity with arg block submissions"
```
knowuh: This is cool.  I didn't know about included context before now.

knowuh:
The test refactoring looks great.  I learned a few tricks.

The first two changes listed in the "files changed" view here, are two controller methods which are fairly identical.  One is `app/controllers/lightweight_activities_controller.rb#dashboard_toc` and the other `app/controllers/api/dashboard.rb#remote_toc`  --  Seems like we can get rid of the older `app/controllers/api/dashboard.rb#remote_toc`

knowuh:
Looks good to me.  :+1:

pjanik:
> Seems like we can get rid of the older app/controllers/api/dashboard.rb#remote_toc

Right, I didn't notice it was there. I've just removed it.

PR #143, opened 2016-01-19
Update dashboard API to use endpoint_urls instead of run keys

pjanik:
1. Use endpoint URLs instead of run keys in `runs/dashboard` API
2. Filter out endpoint URLs that use numeric IDs, accept only those which use long UUID keys.
3. Return `last_page_id` for each run (will be useful in  Dashboard TOC view).

### DIFF
--- a/app/controllers/runs_controller.rb
+++ b/app/controllers/runs_controller.rb
@@ -87,12 +87,6 @@ class RunsController < ApplicationController
   end
 
   # Used by Dashboard app.
-  # This API is open, there is no authorization, but it requires secure
-  # form of endpoint_urls that use long UUID keys instead of numeric IDs.
-  # E.g. this endpoint URL will be accepted:
-  # https://learn.concord.org/dataservice/external_activity_data/key:19679d16-e79e-4f36-8253-5deb1321e6d9
-  # but this will be filtered out:
-  # https://learn.concord.org/dataservice/external_activity_data/123
   def dashboard
     page_id = params[:page_id]
     endpoint_urls = params[:endpoint_urls] || []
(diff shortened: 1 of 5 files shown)


scytacki:
I don't think we should do any filtering in LARA for the new format of endpoint_url. It is adding a dependency in LARA to the URL format the portal is using.  It does mean that if someone found a endpoint_url that has an id in it, then they could walk the ids. But it would only work for the older endpoints. I hope that we will have time to make a script to clean up the older end points.

pjanik:
I didn't like `key:`, but thought it would be better than making other assumptions about form of the remote endpoint URL (more explicit than looking for digits between `/` and so on). 

However, this API returns only arg block data. So in fact, 99% of IDs won't return anything. I just thought we wanted to make sure that it's impossible to get **any** data. But personally I'm fine with it and happy to remove filtering.

pjanik:
Filtering removed, I've also added "group_id" fields.

app/controllers/runs_controller.rb
```diff
@@ -87,10 +87,16 @@ def details
   end
 
   # Used by Dashboard app.
+  # This API is open, there is no authorization, but it requires secure
+  # form of endpoint_urls that use long UUID keys instead of numeric IDs.
+  # E.g. this endpoint URL will be accepted:
+  # https://learn.concord.org/dataservice/external_activity_data/key:19679d16-e79e-4f36-8253-5deb1321e6d9
+  # but this will be filtered out:
+  # https://learn.concord.org/dataservice/external_activity_data/123
```
scytacki: This comment should be updated now that filtering is removed.

scytacki:
:+1:

PR #144, opened 2016-01-22
Add group support to arg block submissions

pjanik:
This PR adds group support to argumentation block submissions.
If user is running activity with collaborators, each submission is automatically copied to other students.
Each submission copy keeps reference to the base submission, so when "usefulness score" (aka "feedback on feedback") is updated, we can also update submission copies.


scytacki:
:+1:

PR #145, opened 2016-01-22
Add ability to log event_value separately from params

sfentress:
Previously we just stringified the data in the event_value, which was
redundant as it was also being saved in `parameters`. It also led to
errors when the string was > 255 chars.

Now we can log like so:

`phone.post('log', {action: 'actionName', value: 'actionValue', data: {someValue: 1, otherValue: 2})`

Note: We could support backward compatibility by putting `data` in
`eventValue` if `value` doesn't exist (and trimming it to 255 chars).
But if only Lab is logging like this, it would be quick to update Lab.

### DIFF
--- a/app/assets/javascripts/logger.js
+++ b/app/assets/javascripts/logger.js
@@ -173,7 +173,7 @@ LoggerUtils.prototype._logInteractiveEvents = function(iframe) {
   phone.addListener('log', function (content) {
     this._logger.log({
       event: content.action,
-      event_value: JSON.stringify(content.data),
+      event_value: content.value,
       parameters: content.data,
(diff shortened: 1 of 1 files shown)


scytacki:
@pjanik do you see any problems with this?

The LARA and Lab logging data from Sept 1 to Dec 16, has been sent off to some folks at ETC, but I don't think they will mind if the JSON form of the data is removed from the event_value column.

pjanik:
Absolutely not, it looks good to me. I've added this weird line only to be backward compatible with old RPC logging handler (a few lines above in the same file). It's sending unparsed data as "event_value", so this line ensured it was still working like this.

PR #150, opened 2016-02-04
AuthPortal was not checking port numbers.

knowuh:
This change will only benefit developers who want to run multiple portals on localhost (each on its own port).  Looking up the portal by URL will now take into account the port.


scytacki:
:+1:

PR #151, opened 2016-02-05
Change labbook prompt wording in so that students know its important.

knowuh:
Suggested wording in PT: "A snapshot must be included for the activity to show as complete."

This change also tries to put all that text in I18n files.

Running `rake itsi:update_labbook_prompts`  should make the necessary changes to close [#112928599].  

[#112928599]
https://www.pivotaltracker.com/story/show/112928599

### DIFF
--- a/app/models/embeddable/labbook.rb
+++ b/app/models/embeddable/labbook.rb
@@ -11,8 +11,8 @@ module Embeddable
       ['Snapshot', SNAPSHOT_ACTION]
     ]
 
-    WONT_DISPLAY_MSG     = I18n.t('LABBOOK_WONT_DISPLAY')
-    NO_INTERACTIVE_LABLE =   I18n.t('LABBOOK_NO_INTERACTIVE')
+    WONT_DISPLAY_MSG     = I18n.t('LABBOOK.WONT_DISPLAY')
+    NO_INTERACTIVE_LABLE =   I18n.t('LABBOOK.NO_INTERACTIVE')
     NO_INTERACTIVE_VALUE = "no-interactive"
     NO_INTERACTIVE_SELECT = [NO_INTERACTIVE_LABLE, NO_INTERACTIVE_VALUE];
     attr_accessible :action_type, :name, :prompt,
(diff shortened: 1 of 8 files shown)


knowuh:
I am going to look a these test failures over the weekend.

scytacki:
Apparently a FactoryGirl created model doesn't have access to the `t` method.

scytacki:
Perhaps it is a legitimate but and you should use `I18n.t` instead.

scytacki:
@knowuh I just tried to run this on staging, but it failed a couple of ways:
First I got:
```
$ bundle exec rake itsi:update_labbook_prompts
rake aborted!
NameError: uninitialized constant Embeddable
/web/portal/releases/20160223155232/lib/tasks/itsi_tasks.rake:53:in `block (2 levels) in <top (required)>'
```

So then I ran the rails console and ran:
```
Embeddable::Labbook.find_in_batches do |labbooks|
      labbooks.each(&:update_itsi_prompts)
end
```
That resulted in:
```
  Embeddable::Labbook Load (19.0ms)  SELECT `embeddable_labbooks`.* FROM `embeddable_labbooks` ORDER BY `embeddable_labbooks`.`id` ASC LIMIT 1000
NoMethodError: undefined method `downcase' for nil:NilClass
	from /web/portal/releases/20160223155232/app/models/embeddable/labbook.rb:191:in `close_enough'
	from /web/portal/releases/20160223155232/app/models/embeddable/labbook.rb:199:in `update_itsi_prompts'
	from (irb):3:in `each'
	from (irb):3:in `block in irb_binding'
	from /web/portal/shared/bundle/ruby/1.9.1/gems/activerecord-3.2.19/lib/active_record/relation/batches.rb:72:in `find_in_batches'
	from /web/portal/shared/bundle/ruby/1.9.1/gems/activerecord-3.2.19/lib/active_record/querying.rb:8:in `find_in_batches'
	from (irb):2
	from /web/portal/shared/bundle/ruby/1.9.1/gems/railties-3.2.19/lib/rails/commands/console.rb:47:in `start'
	from /web/portal/shared/bundle/ruby/1.9.1/gems/railties-3.2.19/lib/rails/commands/console.rb:8:in `start'
	from /web/portal/shared/bundle/ruby/1.9.1/gems/railties-3.2.19/lib/rails/commands.rb:41:in `<top (required)>'
	from script/rails:6:in `require'
	from script/rails:6:in `<main>'
```

I'm guessing the first error requires an import in the rake file. The second error probably means the update_itsi_prompts doesn't handle a nil prompt.

knowuh:
OK I will look at this

PR #152, opened 2016-02-17
Send processing times to portal. Version Protocol

knowuh:
NOTE:  There are still some pending tests that I am going to look into before asking for a review.

* PortalSender::Protocol handles  different versions of learner data
* Send additional processing time info to potral.

[#112926927]

https://www.pivotaltracker.com/story/show/112926927

### DIFF
--- a/spec/services/portal_sender_spec.rb
+++ b/spec/services/portal_sender_spec.rb
@@ -1,0 +1,79 @@
+require 'spec_helper'
+
+describe PortalSender::Protocol do
+  let(:endpoint)            { "http://localhost/dataservice/external_activity_data/6" }
+  let(:auth_stubs)          { {} }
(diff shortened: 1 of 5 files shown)


knowuh:
I think the PR is ready for review.  A little more work remains on the portal side.

spec/services/portal_sender_spec.rb
```diff
@@ -0,0 +1,172 @@
...
+  end
+
+
+  describe "add-hoc routing maddness" do
+
+  end
+
+  describe "when the portal supports the latest version of the protocol" do
+    let(:post_results) { happy_result }
+    it "should not be using the most rescent protocol version after sending" do
```
scytacki: I think the `not` here should be removed.

scytacki:
:+1:

PR #153, opened 2016-02-18
Update README.md

pjanik:
Simpler setup tips, following new Portal readme style. Added missing `rake secret` step and link to Portal readme section about SSO integration.


knowuh:
Looks good.  Thanks for linking to portal docs.  :+1:

PR #154, opened 2016-02-19
Add unauthorized dialog to single_page activities

scytacki:
We forgot to add this unauthorized dialog to the single_page (ITSI style) activities. So students were getting in a state where their work was not saving, but they didn't really notice.


scytacki:
Don't merge this yet. I discovered another path that needed to brought over from the multiple page world. When a user looses credentials and then reloads the page. This is now working with the second commit. But the styling is wrong, so it will be confusing.

scytacki:
This is ready to be reviewed now.

scytacki:
@pjanik or @knowuh can you take a look at this? I'd like to get this tested and deployed ASAP to prevent more ITSI data loss.

knowuh:
Code looks good @scytacki .  Do you want me to check it out locally and smoke test it?

scytacki:
I smoked tested it locally. I don't think you need to. I'm going to get it up on staging, and then see if Evangeline can test it out today.

PR #155, opened 2016-02-22
require branch on deploy, show branch name to users

scytacki:
This is the same deploy approach that the portal is using now.
This PR also displays the deployed branch to the user.


scytacki:
@knowuh can you take a look at this one too. This isn't urgent though, so no rush.

knowuh:
:+1:

PR #156, opened 2016-02-23
Display warning when user tries to leave page and data is not saved

pjanik:
The confirmation dialog will be displayed when:
- Image question dialog is open (or still saving). That's easy to test.
- When regular question is changed, but it's not saved to server. To test that, you need to type something and quickly use browser reload button. If you try to use regular page link in the LARA navigation bar, you won't notice anything, as we handle that case and force save (unless you make sure that it fails, e.g. by killing your local rails server).

Note that browsers have predefined text that is displayed. Some display our custom message too (Chrome, Safari), but other ignore it (Firefox). But even when it's ignored, the confirmation dialog still makes sense (as it mentions possible data loss).


scytacki:
:+1:

PR #157, opened 2016-02-23
Fix issues with ITSI labook prompt renaming.

knowuh:


knowuh:
@scytacki   The issue with the rake task was not passing in the environment.    Added some tests, and handled cases where prompt is nil.

scytacki:
:+1:

PR #158, opened 2016-02-25
Fix completed banner (handle Labbook answers better)

pjanik:
LARA sends Labbook album link to Portal when user opens Labbook dialog. That causes Portal report update and only then Portal treats Labbook question as completed.

However, internally LARA was always assuming that Labbook is a completed question. That was causing difference between progress reported in Portal and LARA. 

This PR makes that consistent with Portal reporting - Labbook is completed (non-blank) in LARA after user closes Labbook dialog for the first time. Implementation may seem a bit strange, but I hope that comment explains that.

### DIFF
--- a/app/models/embeddable/answer.rb
+++ b/app/models/embeddable/answer.rb
@@ -69,9 +69,13 @@ module Embeddable::Answer
     end
   end
 
-  # Overwrite this method in class that includes this module.
+  # Overwrite these methods in class that includes this module:
   def copy_answer!(another_answer)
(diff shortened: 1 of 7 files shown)


scytacki:
This use of `blank?` seems confusing. Could you change it to be easy to change it to `answered?`? 
I realized that you'd need to change all of the other *_answer object. I couldn't figure out what is calling this `blank?`

scytacki:
:+1:

PR #160, opened 2016-02-29
don't send annotation to portal

scytacki:
The portal is not saving this annotation data, so we are wasting resources sending it.

As far as I can tell here is the code in the portal that is ignoring the annotation:
https://github.com/concord-consortium/rigse/blob/master/app/models/dataservice/process_external_activity_data_job.rb#L69


knowuh:
Several `response_for_portal` test errors related to this change in, for example `spec/embeddable/image_question_answer_spec.rb`

PR #162, opened 2016-03-01
Fix image answer #answered? logic

pjanik:
It's another issue which is affecting completed banner in LARA. I think it's quite likely that it was causing problems in Dan's activities (as he wasn't using Labbooks).
`image_url` is a background URL and it can be preset by the author. In such case, LARA would assume that given question is always answered and it might cause that banner shows up while it shouldn't.

I've also removed `answer_text` from the `#answered?` conditions, as there is no way to provide answer text without saving an annotated image. It always happens when you click "Save" button, so we can make code a bit simpler.


scytacki:
:+1:

PR #163, opened 2016-03-08
send more data to the portal with answers

scytacki:
We saw some hard to explain to data coming from LARA. This extra data should help explain issues like what we saw.  This PR adds:

`oldest_answer_id` and `run_id`


scytacki:
Don't review or merge this yet. I want to change it to be more generic.

PR #166, opened 2016-05-09
fix issue with https portals

scytacki:
Recently LARA was changed to use both the host and the port when looking up which authentication token should be used. This change broke the run with collaborators support, because that was only passing in the host and assuming the protocol was http with a port of 80. 

This PR fixes that issue, and also requires that the collaborators_data_url include the protocol.  Originally the portal did not send the correct protocol with this url, but that was fixed in early 2015. So every production portal should be sending it by now:
https://github.com/concord-consortium/rigse/commit/fae61205fe6696b5519cb4afde3fbbe7b22635c5


scytacki:
@knowuh can you review this?

knowuh:
Looks good 👍

PR #168, opened 2016-05-19
Adds Rollbar support.

knowuh:
@scytacki I Added rollbar to the major layouts.  Have a look.

We will have to add ROLLBARS_KEY to our deployment environments.


scytacki:
Looks good 👍

PR #169, opened 2016-05-25
Improve print styles

pjanik:
Bunch of CSS changes:
- Remove the light grey text in each textarea that has a prompt to write text here.
- Remove the boxes around open response questions and the text areas for the open response questions.
- For interactives, images, and videos, float the image on the right and have the questions and text boxes in the info/assessment block flow around them.
- Hide interactive iframes (as they usually don't have enough time to load). There's a placeholder box instead + URL below. I've also removed "Click to play" overlay from the image if it's available.
- Hide "menu" and it's icon in the upper left side of the navigation bar.
- Hide "Welcome ...".
- Replace pulldown menu in multiple choice question is with normal list
- Hide buttons (check answer, draw image, etc.)

Also, I've extracted the print-specific styles to `print-styles.scss`.

There're quite a few `!important` rules which are typically considered to be a bad idea, but I think it makes sense in this case. Otherwise, I would need to provide lots of nested selectors, making sure that it's more specific than the original one. I guess it's more error prone long-term, for example if the dom structure changes.

### DIFF
--- a/app/assets/stylesheets/partials/_click_to_play.scss
+++ b/app/assets/stylesheets/partials/_click_to_play.scss
@@ -1,6 +1,6 @@
 .interactive-container {
     position: relative;
-    .img-container,.click_to_play,.click_to_play .background,.click_to_play .text,.interactive-print-text {
+    .img-container,.click_to_play,.click_to_play .background,.click_to_play .text{
       position: absolute;
       top: 0;
       right: 0;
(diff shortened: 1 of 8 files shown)


pjanik:
Demo:
http://print-styles.lara.demo.concord.org/activities/1?print=true

ddamelin:
@pjanik The link above requires permissions to open it. The error message looks like it is from a LARA build. Can you mark that activity as "anyone with the link can run"?

pjanik:
Thanks, you're right, it was private. Fixed.

ddamelin:
This looks great to me. Thanks!

knowuh:
Looks good  :+1

PR #172, opened 2016-05-27
Update themes

pjanik:
I've made themes mixing a bit more flexible + simple CSS changes:
https://www.pivotaltracker.com/n/projects/736901/stories/118921669


scytacki:
👍

PR #173, opened 2016-06-02
InteractiveRunStates can be linked

knowuh:
(PR not ready)

@dougmartin and I will be working here on these features:

An iframed interactive on a LARA activity page that saves [state can be linked](https://www.pivotaltracker.com/story/show/118940579)  to an iframe interactive on an earlier activity page.

If an author [copies an activity](https://www.pivotaltracker.com/story/show/118941131) with interactives that refer to earlier interactives, it needs to modify the unique identifiers so they remain unique.

If an iframe interactive refers to an earlier interactive, and that interactive has no saved state, it should [launch from the original](https://www.pivotaltracker.com/story/show/118941289) document for the earlier interactive.


scytacki:
I started to look at this, but then I wanted to rename 'parent' to 'linked' in the code. I don't want to interrupt y'all so I'll try to make that change myself.

scytacki:
I think this is good to go. The changes I made are:
- renamed `parent` to `linked_interactive`
- renamed `has_linked_state` to `has_linked_interactive`, and changed its logic.

The point of the `has_linked...` property was to let an interactive know if there is a linked interactive even if the user hasn't created any state for it yet. Now

PR #174, opened 2016-06-09
Max score in feedback

knowuh:
This should be enough (on the LARA side) to get us to view same score numbers & colors in dashboard report and student run.

As per these stories:

https://www.pivotaltracker.com/story/show/120595151
https://www.pivotaltracker.com/story/show/120595223

@pjanik if you can spare a moment for review?


pjanik:
Generally looks good to me. 
Since we obtain max_score dynamically from the feedback_item /  score_mapping, why do we need to save it in feedback_item model? Couldn't we always delegate this method to the item_settings / mapping objects? It seems to be a bit redundant and probably makes handling of the old submissions a bit more difficult (if we don't store it, I guess we won't have to do anything). But I might be missing something (performance?).

knowuh:
@pjanik -- You are right.  I don't think we need to save the max_score.

knowuh:
I added https://github.com/concord-consortium/lara/pull/174/commits/577cfb3a8fc494613389a28ed0c291231c3a896d to update the coffeescript c-rater code, and now I think this PR is ready.

pjanik:
Have you decided to remove max_score field from DB or not?
It seems like feedback_functionality delegates max_score to question now, but there are still migrations and it's being saved.

knowuh:
Thanks for the reminder @pjanik . I just committed 931e43b which removes the save / schema changes.   Now I *really* think this pr is ready ;)

pjanik:
Thanks, looks good. :)

scytacki:
I don't think it is worth changing now. But we were thinking of saving the max-score because the author might change the score-mapping after the student has seen the feedback. We save the feedback text that they saw for this reason. If we don't save the max-score then we wouldn't know which graphic they saw along with their feedback text.  

I think it is unlikely that an author would make a change like this so it is not worth changing now.

PR #175, opened 2016-06-14
Fix completed banner when activity doesn't have any questions

pjanik:
@eireland has found following bug:
> activities with just interactives that don't save state or have report URLs and no questions always show Complete banner in student sequence TOC and preview sequence TOC.

This tiny change should fix it. Banner is shown only if user has opened an activity before.


knowuh:
Looks good.

PR #176, opened 2016-06-14
Add authorization to score_mapping_controller

pjanik:
[#120788853]

### DIFF
--- a/app/controllers/c_rater/score_mappings_controller.rb
+++ b/app/controllers/c_rater/score_mappings_controller.rb
@@ -1,5 +1,6 @@
 class CRater::ScoreMappingsController < ApplicationController
-  authorize_resource
+  load_and_authorize_resource
+  skip_load_resource :only => [:index, :create]
 
   def index
     @filter = CollectionFilter.new(current_user, CRater::ScoreMapping, params[:filter] || {})
@@ -7,7 +8,6 @@ class CRater::ScoreMappingsController < ApplicationController
   end
 
   def new
-    @score_mapping = CRater::ScoreMapping.new()
     respond_to do |format|
       format.js { render :json => { :html => render_to_string('new'), :css_class => "feedback-set" }, :content_type => 'text/json' }
       format.html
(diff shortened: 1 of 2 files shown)


scytacki:
@pjanik you probably can just add a `load_and_authorize_resource` to the top of the controller and remove the individual `authorize!` calls.  You can then also get rid of the `@score_mapping = CRater::ScoreMapping.find(params[:id])` statements.

pjanik:
Thanks. I've tried `load_and_authorize_resource`, but it's not really working with the way how parameters are handled when new score mapping is created. So, I've used just `authorize_resource`.

scytacki:
Unfortunately authorize_resource is probably not doing what you expect. If the `@score_mapping` variable is not set before authorize_resource runs, then it will just try to authorize `ScoreMapping`.  This works because admins have the ability `can :manage, :all`. Which means it doesn't mater what the resource or action is admins can do it.

If the only problem is in the create method, then you can do:
`load_and_authorize_resource :except => :create`

Then you can manually use the authorize method in the create method. Something like:
`authorize :create, ScoreMapping`

This approach has the advantage of allow use to use more fine grained abilities specific to instances later on without modifying the controller.

scytacki:
I think this would also do the same thing:
```
load_and_authorize_resource
skip_load_resource :only => :create
```

pjanik:
> Unfortunately authorize_resource is probably not doing what you expect. If the @score_mapping variable is not set before authorize_resource runs, then it will just try to authorize ScoreMapping. 

Actually, it was intentional. I was assuming that only admins can modify ScoreMappings and we don't really need to check anything specific to given instance, e.g. an owner. I didn't update abilities file anyway.

> This approach has the advantage of allowing us to use more fine grained abilities later without modifying the controller.

Agree, that's the main difference. I can't see an use case for more specific permissions in this case, but well, it won't hurt to follow this pattern. I'll do it.

pjanik:
Done, thanks.

scytacki:
👍

PR #177, opened 2016-06-15
Set max-score styles for arg blocks in c_rater.cofee

knowuh:
Also:  small bug fix / tweek to /dev/test_argblock

[#120595151]
https://www.pivotaltracker.com/story/show/120595151


knowuh:
Super small change, so I will merge myself.

PR #178, opened 2016-06-15
Fix tiny layout bug in 'Begin activity button'.  (truncates button)

knowuh:
Confluence of single-page layout & print layout styles.

[#121613893]

https://www.pivotaltracker.com/story/show/121613893


knowuh:
@scytacki this addresses the "Begin activity".

I tested the "Submit" button in single page activities and the "Begin activity" button in the HAS theme. I am not sure that the `.submit` button in the single page activities has to be 80px.... but I didn't want to introduce more change.

scytacki:
looks good 👍 
I might have to cherry pick this onto a new branch because other commits were added to master after Evangeline started testing.

PR #181, opened 2016-06-20
Optimize Dashboard API (reduce N+1 queries, support update timestamps)

pjanik:
I was playing with Dashboard API and I had following observations:
- There were some N+1 queries, easy to fix so I did it. It helped a bit, but still calls to `/runs/dashboard` was expensive.
- The most expensive part was calculating submissions - we need submissions, answers and related questions data. I created a test setup (5 learners, a few submissions) and single request was taking around 500-800ms in total.
- Basic run info (endpoint, last_page_id, etc.) can be returned pretty fast, around 50-100ms.

So, I decided that the best approach would be to remember data timestamp and make sure that we don't calculate old submissions info again and again during each request. Fortunately, it's quite simple here in LARA API and a bit more complicated in the client code (I guess that's our preference anyway). Related HASDashboard PR:
https://github.com/concord-consortium/HASDashboard/pull/17

Also, I' changed data format a bit, so I added version number. The client app is backward compatible, it would work with old and new API (e.g. fake data is equivalent to old API). 

Results:
dashboard
Initial request that returns all the submissions took around 500-600ms. Then we had a few requests < 100ms that returned no submissions. Then I logged in as a student, created one new submission and it was returned by request which took 350ms.

### DIFF
--- a/app/services/dashboard_runlist.rb
+++ b/app/services/dashboard_runlist.rb
@@ -1,21 +1,32 @@
 class DashboardRunlist
-  def initialize(endpoint_urls, page_id)
-    @runs = Run.includes(:activity).where(remote_endpoint: endpoint_urls).map do |run|
+  DATA_VERSION = 2
+
+  def initialize(endpoint_urls, page_id, submissions_created_after = 0)
+    # DB query can be optimized by filtering out submissions older
+    # than provided timestamp. If it's not provided (or nil), we will simply return all the submissions.
+    submissions_created_after = Time.at(submissions_created_after ? submissions_created_after.to_i : 0)
+    runs_data = Run.includes(:activity).where(remote_endpoint: endpoint_urls).map do |run|
       {
         endpoint_url: run.remote_endpoint,
         group_id: run.collaboration_run_id,
         last_page_id: run.page_id,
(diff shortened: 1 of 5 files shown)


app/services/dashboard_runlist.rb
```diff
@@ -24,8 +35,19 @@ def to_json
...
 
-  def submissions(run, page_id)
-    CRater::FeedbackSubmission.includes(:c_rater_feedback_items => :answer).where(run_id: run.id, interactive_page_id: page_id).order(:id).map do |submission|
+  def submissions(run, page_id, submissions_created_after)
+    CRater::FeedbackSubmission
+      .includes(
+        c_rater_feedback_items: {answer: [:run, :question]},
+        embeddable_feedback_items: {answer: [:run, :question]}
+      )
+      .where('created_at > ?', submissions_created_after)
```
knowuh: The index for this query currently uses runs & page.  I wonder if we should also add new index on all three columns `interactive_page_id`, `run_id`, and `created_at`   What do you think?

pjanik:
@knowuh, good point, thanks. Index is updated in the commit above.

Also, another tiny optimization is that the client app won't call LARA API when the browser tab is inactive.

PR #182, opened 2016-06-20
squash the filtering of interactives and embeddables

scytacki:
The intention of this PR is to make the code easier to maintain. It also fixes a few bugs in the process.

Here are the basic changes:
removes show_in_report from answers
changes is_reportable to reportable? on interactives
adds reportable? to all embeddables
moves old Activity#questions logic into Pages#reportable_items
removes extra answer filtering from Run

Fixed bugs:
- completion ribbons on Sequence TOC would never show if an activity had a LabBook that was not hidden, but didn't have a valid interactive.  This LabBook would be counted as a question but the student would never see it in order to complete it.
- question numbers would be wrong after a LabBook that was not hidden, but didn't have a valid interactive.
- completion ribbons on Sequence TOC would not show if a interactive was reportable and the student had opened the interactive. All interactives were not counted as answers in the completion equation, but reportable interactives were counted as questions.

### DIFF
--- a/documentation/embeddable-interactive-filtering.md
+++ b/documentation/embeddable-interactive-filtering.md
@@ -8,61 +8,46 @@ Interactives on the page bypass this "answer" approach. They use the interactive
 
 The numbering of items in the activity boils down to:
 
-    activity.questions.index(item) + 1
+    activity.reportable_items.index(item) + 1
 
-This happens when individual 'answer' partials call the Embeddable::Answer#question_index to find the index of the 'question' on the page. The Xhtml partial  doesn't look for an index. Reportable interactives technically have an index, but they do not display it.
+This happens when individual 'answer' partials call the `Embeddable::Answer#question_index` to find the index of the 'question' on the page. The Xhtml partial  doesn't look for an index. Reportable interactives technically have an index because they are part of reportable_items, but they do not display it.
 
 ## Activity Progress
 
 LARA shows activity progress in completion ribbons on the sequence table of contents. The progess is also shown in some admin info reports.
 
-The progress is based on `Run#num_answers` / `Run#num_questions`.
+The progress is based on `Run#num_answers` / `Run#num_reportable_items`.
 
-The num_questions is the count of `Activity#questions`.
+`Activity#reportable_items` is the combination of `Page#reportable_items` on all visible pages.
 
-`Activity#questions` is the combination of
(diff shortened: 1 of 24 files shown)


documentation/embeddable-interactive-filtering.md
```diff
@@ -8,61 +8,46 @@ Interactives on the page bypass this "answer" approach. They use the interactive
 
 The numbering of items in the activity boils down to:
 
-    activity.questions.index(item) + 1
+    activity.reportable_items.index(item) + 1
 
-This happens when individual 'answer' partials call the Embeddable::Answer#question_index to find the index of the 'question' on the page. The Xhtml partial  doesn't look for an index. Reportable interactives technically have an index, but they do not display it.
+This happens when individual 'answer' partials call the `Embeddable::Answer#question_index` to find the index of the 'question' on the page. The Xhtml partial  doesn't look for an index. Reportable interactives technically have an index because they are part of reportable_items, but they do not display it.
```
knowuh: I think this is right.  If we have an interactive that is reportable, with two multiple choice questions `between` it, the question numbers for the multiple choice questions would be say 5 and 7. (skipping one for the interactive that is reportable).  It might be good to have @ddamelin verify this is acceptable behavior.

spec/models/embeddable/labbook_answer_spec.rb
```diff
@@ -42,27 +42,21 @@
...
 
     describe 'with a disabled labbook' do
       let(:stubs) { { show_in_runtime?: false, show_in_report?: false } }
       it "should return false for #show_in_runtime?" do
         expect(labbook_answer.show_in_runtime?).to eql(false)
       end
-      it "should return false for #show_in_report?" do
-        expect(labbook_answer.show_in_report?).to eql(false)
-      end
     end
```
knowuh: We probably want to test when a lab book is hidden it doesn't have any effect on the students progress.

knowuh:
Looks good.  👍

knowuh:
@scytacki I am going to let you merge this.

documentation/embeddable-interactive-filtering.md
scytacki: Yes I'm pretty sure that Dan accepted this. 
Also interactives are not interleaved with questions. So the skips in numbers will only happen between pages.

scytacki:
Yeah I'll add the test that you pointed out, and then fix the merge conflicts.

PR #183, opened 2016-06-22
image and video interactives fix, plus export fix for labbooks without interactives

scytacki:
This fixes an issue with image and video interactives not handling the new `reportable?` method.
It also moves some of the reportable_items tests around and expands them.
Finally it fixes an issue when trying to export activities with Labbooks that don't have a interactive set.


knowuh:
Looks good!  👍

PR #184, opened 2016-06-27
Change URLS  for Has Dashboard C-Rater reports to /api/*

knowuh:
Put all the Dashboard report endpoints in /api/v1/ 

This is the LARA side of https://github.com/concord-consortium/HASDashboard/pull/23/

https://www.pivotaltracker.com/story/show/121491615

### DIFF
--- a/app/controllers/api/v1/dashboard_controller.rb
+++ b/app/controllers/api/v1/dashboard_controller.rb
@@ -1,0 +1,39 @@
+class Api::V1::DashboardController < ApplicationController
+
+  # Controller for HAS Dashboard (aka HASBot C-Rater) report.
+
+
+  # `#runs` returns the json representing report data for students.
+  # Params:
+  #   page_id: The page to report on.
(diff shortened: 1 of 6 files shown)


knowuh:
I will have a look at the failing `question_trackers` tests in a few.  Not sure how its related.

knowuh:
@scytacki  would you care to give this a quick glance?   If not I can hit up some other people. Its small, mostly routing changes.

scytacki:
Did you take a look at the result of `request.params.to_query`? 

I assume the intention is to just include the query params from the original request. I wonder though if rails or rake injects more things into the requests.params that we don't really want to be in the redirect.

knowuh:
@scytacki  simplified in a037320

scytacki:
👍 I'd merge this, but we are still trying to get out the 1.10.0 release. So I'd like to hold off on merging and updating staging until that release is done.

PR #185, opened 2016-06-28
Update project logo tag not to default to MW [#72204218]

dougmartin:


dougmartin:
@scytacki I'm not sure if we wanted to go all the way and remove the default MW logo images in the application wrapper but I went ahead and did it.  I can pull that (and the wide template change) back out if needed.

scytacki:
I think pulling them out of the wrapper is the right thing todo. What happens with `<%= nil %>`? Is it going to emit `"nil"` or does it emit `""`?
I think that application is only used during authoring, in which case `@project` might often be `nil`.  Probably it would be best to have some kind of 'LARA' icon for use during authoring, but I think a blank space is better than the MW logo.

One thing to watch out for though: if there is no MW logo during authoring, does it change the size of the header? I think the portal iframe integration offsets the page by a fixed amount so if the header bar changes size we'll need to update the portal iframe integration.

dougmartin:
@scytacki When the tag helper emits nil an empty string is inserted into the template.

I went ahead and set a min-height on the header logo area so that an empty logo header matches the height of the existing MW logo header (86px).

I also went ahead and changed the default project from MW to "Default Project" which has no logo and points to the Concord homepage.  If an activity has no assigned project if defaults to the default project which caused the MW logo to show.  Now no logo shows.

PR #186, opened 2016-06-29
fix bug in interactive data sent to the portal

scytacki:
[##123607881]

### DIFF
--- a/app/models/image_interactive.rb
+++ b/app/models/image_interactive.rb
@@ -11,6 +11,10 @@ class ImageInteractive < ActiveRecord::Base
     "image interactive"
   end
 
+  def self.portal_type
+    "unsupported"
+  end
+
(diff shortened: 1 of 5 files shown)


scytacki:
@knowuh can you take a look at this one?

knowuh:
OMG so glad you changed the name of the method to "portal_type"  👍 

Reviewing this code made me also wish all interactives were defined in under `interactives` as the embeddable types are under `embeddables`..  But I think you got em all.

scytacki:
I'm hoping we can remove the polymorphic-ness of InteractiveRunState it only really applies to MwInteractives which should really be named IframeInteractives.

I also want to merge embeddables and interactives. It is really the layout of the page that makes them different. They are all just embeddable components.

PR #188, opened 2016-07-05
Print activities

knowuh:
* Add `print_url` attributes to portal published activities and sequences.
* Add new `print_blank` routes for activities and sequences.
* Update sequence and activity listings to include print links.

(+ associated tests / style changes)

[PT Story](https://www.pivotaltracker.com/story/show/102391640)

### DIFF
--- a/app/controllers/lightweight_activities_controller.rb
+++ b/app/controllers/lightweight_activities_controller.rb
@@ -85,6 +85,12 @@ class LightweightActivitiesController < ApplicationController
     @labbook_is_under_interactive = true
   end
 
+  def print_blank
+    authorize! :read, @activity
+    @run = Run.new() #Temporary run for the answer finder. Don't persist.
+    setup_show
+  end
+
   def summary
     authorize! :read, @activity
     current_theme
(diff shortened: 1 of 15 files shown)


app/controllers/lightweight_activities_controller.rb
```diff
@@ -85,6 +85,12 @@ def single_page
     @labbook_is_under_interactive = true
   end
 
+  def print_blank
+    authorize! :read, @activity
+    @run = Run.new() #Temporary run for the answer finder. Don't persist.
```
scytacki: I'm not sure, but I believe this run will get persisted by the AnswerFinder. The AnswerFinder calls `default_answer` for the Answer model associated with each question in the activity. `default_answer` then calls create on the Answer model. It refers to the run when it does this, so my guess is that the run will be persisted.   It starts here:

scytacki:
It looks like the `layouts/print_blank.html.erb` file is very similar to the runtime layout. I'm not sure there is a better way but here are some questions:

Would it be better to add conditionals to that layout so it isn't duplicated so much? 
How does the existing print view handle these changes?
I saw one change was to hide the navigation. Perhaps the existing print view does this with css?

scytacki:
oops, looks like I created a conflict when I merged the completion page.

knowuh:
@scytacki :  I rebased and removed the misleading run comment in c539c71

scytacki:
👍

PR #189, opened 2016-07-07
fix issue with the sequence TOC

scytacki:
Rails does not support 2 layout definitions at the top of one controller
It just picks up the last definition if there is more than one.
[#125534641]

@pjanik can you review this super small change.


pjanik:
Looks good. 👍

PR #190, opened 2016-07-07
Need to show the sequence title even with no logo

scytacki:
The sequence title is the only way to get to the sequence TOC.

The code here is not very obvious unfortunately. The method is sequence_logo_tag but really it is the sequence_header.  And the sequence_logo_tag method is called even if an activity is being rendered because the activity might be running inside of a sequence.

I'll see if I can make this a bit cleaner...

[#72204218]

### DIFF
--- a/app/helpers/logo_helper.rb
+++ b/app/helpers/logo_helper.rb
@@ -18,7 +18,7 @@ module LogoHelper
 
   def sequence_logo_tag
     sequence = @sequence
-    return nil unless sequence && sequence.logo.present?
+    return nil unless sequence
     logo = sequence.logo
 
     if @sequence_run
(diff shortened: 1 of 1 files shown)


scytacki:
I looked at this a bit more, and cleaning up would take more time then I have right now. The word 'logo' is used all over the place for this header section. So I'd need to do a lot of careful search and replace in order to make this more clear. @pjanik if you see minor ways to make it better let me know.

pjanik:
`sequence_logo_tag` seems to be used only locally, so renaming it to `sequence_header` could help a bit and shouldn't be risky. But well, probably it would be nice to rename almost all the methods in this helper. So, I guess let's leave it as it is if you don't have time for proper cleanup.

PR #191, opened 2016-07-13
Display report link next to activity title (in TOC)

pjanik:
A small PR, adds a report link next to the activity title. It looks like this:

screen shot 2016-07-13 at 17 41 39


scytacki:
👍

PR #192, opened 2016-07-15
Add optional, expanding hints to questions

pjanik:
Also, provide default hints for arg block.

[#93136700]
[#126323697]


scytacki:
This PR doesn't have any new tests, but the main change is in the header for the questions. And we don't have a good practice of testing views. So I think it is OK to skip tests on this one.

It is doesn't handle LabBooks.  I'm not sure yet if LabBooks should have hints. `hint` is delegated from Embeddable::Answer and LabBookAnswer include Embeddable::Answer.  So it seems for consistency LabBooks should have hints too. I'll take a look at the LabBook issue.

Also the hints are not added to the `export` and `to_hash`. I'm not sure if they need to part of `to_hash', but they should be part of `export`. I'll work on that too.

scytacki:
OK I think this is good to go now.

pjanik:
Thanks!

PR #193, opened 2016-07-19
Set Google Analytics content group in various places

pjanik:
Set Google Analytics content group on:

- sequence index page
- activity index page
- activity page 
- activity summary / report 

Name is based on the project name.

PT story: https://www.pivotaltracker.com/story/show/124827621

### DIFF
--- a/app/views/interactive_pages/show.html.haml
+++ b/app/views/interactive_pages/show.html.haml
@@ -36,3 +36,8 @@
         .next-page
           %a{ :href => runnable_activity_page_path(@page.lightweight_activity, @page.next_visible_page), :class => 'next forward_nav' }
             %input{ :class => 'button next forward_nav', :type => 'submit', :value => t("NEXT")}
+
+-# Set Google Analytics content group based on the project. Note that project might be undefined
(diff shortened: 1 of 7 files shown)


pjanik:
Hm, I've never actually used GA too much, but now I think I should probably set content group before we call `ga('send', 'pageview');` in shared/_analytics.html.haml partial. Well, at least it would make sense. I'll double check that and fix it if necessary.

eireland:
Hi Piotr,
According to GA, the code would look like:

ga('create', 'UA-XXXXXXXX-Y', 'example.com');
ga('set', 'contentGroup5', 'Group Name');
ga('send', 'pageview');

https://support.google.com/analytics/answer/2853546?hl=en#code
Evangeline

pjanik:
@eireland, thanks! Yeap, I realised that once I pushed that code and read rest of the `ga` calls.
Updated now. Do we need to set `contentGroup5` or 1 or any of them? I've used 1.

Implementation is a bit strange, but it follows the current pattern with templates and at least the code doesn't need to be duplicated too much. Finally we end up with following script tags:

screen shot 2016-07-20 at 10 48 54

scytacki:
👍

PR #195, opened 2016-07-26
disable the print_url because of rendering issues

scytacki:
By removing the print_url from the LARA publication, it will prevent the print button from being shown in the portal.  When the print rendering issues are fixed the print_url can be published again.

[#102391640]


scytacki:
@knowuh does this seem like it should work to you?

knowuh:
This commit disables the publication of the print_url.  Should work.  Activities that already have print_urls defined (on staging) will still have them in the portal though.

scytacki:
Yeah I go in and manually delete the urls from the portal so there is no confusion.

PR #197, opened 2016-08-19
Include sequence path in 'launch next activity' (bug fix)

knowuh:
…e sequence information.

This approach will launch the next activity, saving the sequence.

It also opens the activity index page (even if the student has completed some work in the activity)


scytacki:
👍

PR #198, opened 2016-08-24
Better handling of C-Rater errors

pjanik:
I think C-Rater errors were handled that bad before - C-Rater error message was displayed to user and saved in DB (including the complete request data). Also, we weren't blocking a page, so there should be no data loss, user could go to the next one. But I tried to improve things a bit more.

1. Now, if some error happens, either network or C-Rater one, LARA will try to issue the same request a few times (3). I've decided it's better to do it client side rather than server side (even it's LARA server talking directly to C-Rater). Putting some sleep calls in server-side code could be quite risky - each request would take much longer, and if C-Rates goes down and there's a whole class trying to submit their answer, the result could be obvious. Doing that client side, I could add some long delays between attempts.

2. If all the attempts fail, there's an alert displayed:
screen shot 2016-08-24 at 10 21 35

3. Also, I display C-Rater errors, additional note that there was an error (text in red) and update buttons, so the user can resubmit answers (it wasn't possible before, as typically we let users do it only after some answers are changed) or go to the next page. Also, I don't display "feedback on feedback" box in case of errors.
screen shot 2016-08-24 at 10 21 45

### DIFF
--- a/app/assets/javascripts/c_rater.coffee
+++ b/app/assets/javascripts/c_rater.coffee
@@ -6,6 +6,7 @@ class ArgumentationBlockController
   FEEDBACK_ID_SEL = '#feedback_on_answer_'
   FEEDBACK_TEXT_SEL = '.ab-robot-feedback-text'
   DIRTY_MSG_SEL = '.ab-dirty'
+  ERROR_MSG_SEL = '.ab-error'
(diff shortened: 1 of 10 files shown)


scytacki:
👍

PR #199, opened 2016-08-24
Update completion page for the single activity and the last one in a sequence

pjanik:
Changes are described pretty precisely (including screenshots) in this comment:
https://www.pivotaltracker.com/story/show/125534221/comments/147630335

### DIFF
--- a/app/models/sequence_run.rb
+++ b/app/models/sequence_run.rb
@@ -58,4 +58,8 @@ class SequenceRun < ActiveRecord::Base
       end
     end
   end
+
+  def completed?
+    runs.count == sequence.activities.count && runs.all? { |r| r.completed? }
+  end
 end
--- a/app/views/interactive_pages/_show_completion.html.haml
+++ b/app/views/interactive_pages/_show_completion.html.haml
@@ -5,12 +5,18 @@
   page_name          = @page.name || @activity.name + ":#{@page.position}"
   exit_link          = @sequence ? sequence_path(@sequence, show_index: true) : activity_path(@activity, show_index: true)
   completed_activity = @run && @run.completed?
-  completed_sequence = completed_activity && @sequence && (!next_activity)
+  last_activity      = @sequence && !next_activity
+  completed_sequence = completed_activity && last_activity
+  material_completed = completed_activity || completed_sequence
+  material_thumbnail = completed_sequence ? @sequence.thumbnail_url : @activity.thumbnail_url
+  material_name      = completed_sequence ? @sequence.title : @activity.name
 
   complete_header_t   = t("COMPLETION_PAGE.HEADER", {activity: @activity.name})
   congratulations_t   = t("COMPLETION_PAGE.CONGRATULATIONS")
   finished_activity_t = t('COMPLETION_PAGE.FINISHED_ACTIVITY', activity: @activity.name)
   not_finished_activity_t = t('COMPLETION_PAGE.UNFINISHED_ACTIVITY', activity: @activity.name)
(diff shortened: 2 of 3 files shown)


knowuh:
👍

pjanik:
@knowuh, there's one new commit that handles a case when the last activity is completed but one of the previous ones is not. I'm assuming it doesn't change much in this review, but let me know otherwise. :) I'll merge this PR when Sarah confirms that she likes the wording.

scytacki:
@pjanik and @knowuh as I look at this code and comparing it to the PT story. I had an idea and I wonder what you think about it.   The PT story documents 7 different states the learner could be in when they view this completion page.  What if the code followed this design as well.  So in the main template there would be logic to decide which of the 7 states the user is in, and then there would be 7 different partials, one partial for each state.  This would result in redundant haml, but it would eliminate some of the redundant logic. And it would be easier for a developer to see what each of the states look like. Also this would make it clear that there are these 7 states, so future developers would hopefully know they need to think about all 7 when we want to make changes to the completion page.

knowuh:
@scytacki I like your suggestion as to mapping the 7 states to the seven partials. I believe it would be easier to maintain going forward. In my opinion in this case it totally justifies some duplication.

pjanik:
I also agree. 👍 Logic is shorter now, but it doesn't list clearly all the cases.
I think 7 partials could work or just the very explicit and `switch/if` statements listing all the cases in one partial, so we can avoid duplication. I'll improve it, thanks!

pjanik:
1. Switch / case variant:
https://github.com/concord-consortium/lara/compare/completion-page-case

2. Partials:
https://github.com/concord-consortium/lara/compare/completion-page-partials

I started with 1, but finally switched to partials and I guess I prefer it now.

scytacki:
I spent a while looking at this. The 7 partials didn't look as clear as I was imagining.  So then I came up with more concise way to represent the differences. However after spending too time on it, I'm not sure the concise approach is better than the other options we've tried. Here is my concise representation:
```
### sequence_act_completed_last_seq_completed
congratulations: true
status: 'finished-activity-and-sequence'
thumbnail: sequence
complete-ribbon: true
prompt: 'complete-sequence-exit'

### sequence_act_completed_last_seq_not_completed
congratulations: true
status: 'finished-activity'
thumbnail: activity
complete-ribbon: true
prompt: 'last-activity-exit'

### sequence_act_completed_not_last
congratulations: true
status: 'finished-activity'
thumbnail: next_activity
prompt: 'next-activity-or-exit'

### sequence_act_not_completed_last
status: 'unfinished-activity'
thumbnail: activity
prompt: 'go-back-or-exit-non-completed-activity'

### sequence_act_not_completed_not_last
status: 'unfinished-activity'
thumbnail: next_activity
prompt: 'next-activity-or-exit'

### single_act_completed
congratulations: true
status: 'finished-activity'
thumbnail: activity
complete-ribbon: true
prompt: 'complete-activity-exit'

### single_act_not_completed
status: 'unfinished-activity'
thumbnail: activity
prompt: 'incomplete-activity-go-back-or-exit'
```
This concise representation looks nice.  However, the view code would then contain all the conditionals to render this representation, so it seems the code would not really be any better to maintain.

Any thoughts?

pjanik:
@scytacki, I think I like it. It's better than 1. for sure. It was actually easy to refactor 1 into your approach, so I've tried it:

https://github.com/concord-consortium/lara/compare/completion-page-concise

At least all the use cases are described nicely. And the view code is a bit shorter too.

scytacki:
I've made the concise approach into a new PR, so I'm going to close this one:
https://github.com/concord-consortium/lara/pull/205
Thanks for all of your work @pjanik

PR #201, opened 2016-08-25
Change interactive_run_state api to use random keys instead of ids [#128580257]

dougmartin:


scytacki:
LARA uses CanCanCan for authorization. Did you try using it and ran into a road block? If not can you give it a shot instead of using the `is_authorized` method. The part I'm not sure of is how you can control the error message that should be returned if the authorization fails. I believe you can add a `rescue_from` to this specific controller and handle the `CanCan::AccessDenied` exception.

dougmartin:
I changed the controller to use Cancancan.  You can only define rescue_from in the application controller - if you do it in sub-controllers for custom error message it seems to be ignored.  So I simply wrapped the authorize! calls in the block and rescued the CanCan::AccessDenied exception and called my custom error response from there.

scytacki:
👍

PR #202, opened 2016-08-29
Moved interactive run states endpoint to /api/v1/

dougmartin:


scytacki:
The actual controller file should be moved into the api/v1/ folder as well. That will make it easier for devs to find it.

PR #203, opened 2016-08-30
Added interactiveStateUrl to iframe-saver initInteractive message

dougmartin:
This is needed so that the v2 launch page can add the url to the codap launch url.

Also fixed the get/put url helpers for interactive run state api endpoints.  When the endpoints were moved until the /api/v1 namespace the url helpers needed to have the namespace prepended to them.

### DIFF
--- a/app/assets/javascripts/iframe-saver.coffee
+++ b/app/assets/javascripts/iframe-saver.coffee
@@ -134,12 +134,14 @@ class IFrameSaver
     # this is the newer method of initializing an interactive
     # it returns the current state and linked state
     init_interactive = (err, response) =>
+      loc = window.location
       @iframePhone.post 'initInteractive',
         version: 1,
         error: err
         interactiveState: if response?.raw_data then JSON.parse(response.raw_data) else null
         hasLinkedInteractive: response?.has_linked_interactive or false
         linkedState: if response?.linked_state then JSON.parse(response.linked_state) else null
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/iframe-saver.coffee
```diff
@@ -32,7 +32,12 @@ class IFrameSaver
     phone_answered = =>
       # Workaround IframePhone problem - phone_answered cabllack can be triggered multiple times:
       # https://www.pivotaltracker.com/story/show/89602814
-      return if @already_setup
+      if @already_setup
+        # if the interactive has already been loaded (for instance when the launch iframe reloads when resetting the document) send the initInteractive response under a new message
+        # so that the reloaded frame can look for it as well as initInteractive.  The initInteractive message is not reused in case it serves as an initial starting message. 
+        if @initInteractiveResponse
+          @iframePhone.post 'previousInitInteractive', @initInteractiveResponse
+        return
```
scytacki: This is confusing. Let me see if I have this right:
If the iframe reloads itself, then the iframe will send the iframe phone 'answered' message again. 
Since the iframe is reloaded it no longer know what the initInteractiveReponse is.
So in this case we are sending the initInteractiveResponse again.

Perhaps the iframe doesn't really need to reload itself.  Also it seems there should be a way to identify for sure that is a new 'window' that is now communicating with us. So then the already_setup variable could be based on the identifier of the child window. If the browsers don't have a way to identify windows like this, then perhaps we should add some kind of window identifier to iframe phone that is sent along with each message.

scytacki:
👍

PR #205, opened 2016-09-02
Completion page fixes

scytacki:
This is mostly work that @pjanik did to refactor the completion page.
I added on fixes for:
- handling anonymous users running sequences.
- handling the sequence TOC page correct from the exit button


pjanik:
Improvements look good, thanks. 👍

PR #206, opened 2016-09-08
Removed check for CSRF token on interactive_run_states_controller#update

dougmartin:


scytacki:
👍  Too bad about the rspec limitation.

PR #207, opened 2016-09-08
Speedup bulk prompt change for Labbook items (ITSI)

knowuh:
This commit speeds up the bulk update on Labbook prompts and simplifies the code.

Original Story: https://www.pivotaltracker.com/story/show/112928599
The text included for Lab album instructions in ITSI activities says that "A snapshot must be included for the activity to show as complete."

[#112928599]


scytacki:
👍

PR #209, opened 2016-09-13
Add print url to sequences

scytacki:
This adds in a missing print_url when sequences are published to the portal.

@knowuh or @dougmartin can you take a look?

### DIFF
--- a/spec/models/sequence_spec.rb
+++ b/spec/models/sequence_spec.rb
@@ -102,8 +102,8 @@ describe Sequence do
 
     let(:simple_portal_hash) do
       url = "http://test.host#{Rails.application.routes.url_helpers.sequence_path(sequence)}"
-      author_url = "http://test.host/sequences/#{sequence.id}/edit"
-      print_url = "http://test.host/sequences/#{sequence.id}/print_blank"
+      author_url = "#{url}/edit"
+      print_url = "#{url}/print_blank"
       {"type"=>"Sequence", "name"=> sequence.title, "description"=> sequence.description,
         "abstract" => sequence.abstract,
         "url"=> url,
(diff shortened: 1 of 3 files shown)


spec/models/sequence_spec.rb
```diff
@@ -102,10 +102,14 @@
 
     let(:simple_portal_hash) do
       url = "http://test.host#{Rails.application.routes.url_helpers.sequence_path(sequence)}"
+      author_url = "http://test.host/sequences/#{sequence.id}/edit"
+      print_url = "http://test.host/sequences/#{sequence.id}/print_blank"
```
dougmartin: This might be cleaner as `"#{url}"/edit"` and `"#{url}/print_blank"`

spec/models/sequence_spec.rb
```diff
@@ -115,11 +119,15 @@
 
     let(:complex_portal_hash) do
       url = "http://test.host#{Rails.application.routes.url_helpers.sequence_path(sequence_with_activities)}"
+      author_url = "http://test.host/sequences/#{sequence_with_activities.id}/edit"
+      print_url = "http://test.host/sequences/#{sequence_with_activities.id}/print_blank"
```
dougmartin: Same note as line 106.

dougmartin:
Looks good :+1:

spec/models/sequence_spec.rb
scytacki: Yeah I thought about that, but I followed the pattern here: https://github.com/concord-consortium/lara/blame/master/spec/models/lightweight_activity_spec.rb#L330

I figured @knowuh had a reason for doing it that way.  But now that I look at it, I don't see an advantage. I'll probably change all three places unless I can think of a benefit to this literal approach.

PR #210, opened 2016-09-15
Fix run lookup in arg block controller and make sure that answer IDs are unique

pjanik:
Fixes: https://www.pivotaltracker.com/story/show/130399081

I've been able to reproduce this bug locally and this fix helps. The problem was that run lookup was incorrect and sometimes we're asking for feedback for the wrong run (it could happen if user has run given activity a few times, using different sequences).

Now run key is provided explicitly in URL.

Also, previous code was using only numbers as answer ID. Theoretically there could be conflicts between them. I've changed that to use class name + ID.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended to check this file into your version control system.
 
-ActiveRecord::Schema.define(:version => 20160722195644) do
+ActiveRecord::Schema.define(:version => 20160824191003) do
 
   create_table "admin_events", :force => true do |t|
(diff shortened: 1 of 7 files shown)


db/schema.rb
```diff
@@ -332,8 +332,10 @@
     t.datetime "updated_at",                          :null => false
     t.text     "learner_url"
     t.boolean  "is_dirty",         :default => false
+    t.string   "key"
```
scytacki: Weird. I guess @dougmartin didn't include the changes to schema.rb when he did the interactive_run_states work.

PR #211, opened 2016-09-15
Fix arg block update (wrong IDs), add #answer_id method to Answer module

pjanik:
Feedback wasn't showing up, as the feedback DOM elements ID were still using numeric-only answer IDs, like "feedack_on_answer_123" instead of "feedback_on_open_response_answer_123".

Also, answer ID calculation was done in quite a few places (#{embeddable.class.to_s.demodulize.underscore}_#{embeddable.id}"), so I've added `#answer_id` method to `Answer` module / mixin and tried to use it where possible.


scytacki:
👍

PR #212, opened 2016-09-19
widen print boxes, open activities in a new window

scytacki:
Long activity titles were being cut off. This widening fix should reduce the chances of that. 
Also when a user clicks on the activity buttons on the sequence print page they should open in new tabs or windows.  

[#130152563]


knowuh:
@scytacki merged.

PR #213, opened 2016-09-21
Add linked state to copy

dougmartin:


scytacki:
@dougmartin can you make time to add these tests during this portal/lara iteration you are on?

PR #214, opened 2016-09-21
Added migration task

dougmartin:


scytacki:
This looks good. I want to check if the report_url is correct though given the recent bug report by Evangeline.

scytacki:
Using the server param for the root of the report URL seems right.  So I don't think this code has a problem with it.

PR #216, opened 2016-09-30
remove re-login, this has caused problems

scytacki:
Currently, if a student clicks the "not me" link:
1. they are logged out of the LARA
2. redirected to a logout route of the portal with the re-login and provider param
3. redirected back to LARA with the re-login and provider param
4. redirected back to the portal and shown the simple portal login screen
5. redirected back to LARA and shown the page that they clicked the "not me" link on.

This last step causes problems because this page is likely somewhere this user is not supposed to go.

This PR changes this to:
1. they are logged out of the LARA
2. redirected to a logout route of the portal with no parameters

They should end up on the portal home page with the login on the top right (either a button or the actual login fields).

It was not strictly necessary to remove the support for 're-login' to make this change, but since no other code was using it it seems best to remove it. One bit of removed code can be partially supported by some slightly different parameters. I'll comment on that inline.

[#131411477]

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -1,15 +1,11 @@
 class HomeController < ApplicationController
   def home
-    if(params[:re_login])
-      redirect_to user_omniauth_authorize_path(params[:provider])
-    else
-      @filter  = CollectionFilter.new(current_user, LightweightActivity, params[:filter] || {})
-      # TODO: Add 'oficial' to the criteron?
-      @activities = @filter.collection.includes(:user,:changed_by,:portal_publications).first(10)
-      @filter.klass = Sequence
-      # TODO: Add 'oficial' to the criteron?
-      @sequences  = @filter.collection.includes(:user,:lightweight_activities).first(10)
-    end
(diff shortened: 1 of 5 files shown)


app/controllers/home_controller.rb
```diff
@@ -1,15 +1,11 @@
 class HomeController < ApplicationController
   def home
-    if(params[:re_login])
-      redirect_to user_omniauth_authorize_path(params[:provider])
```
scytacki: This feature can still be partially supported by using the "domain" parameter on any route in LARA.
One difference though, is the redirecting.  The code removed here leaves the user on the referrer page. So if this is code is run after a series of redirects then the user will end up back where they started. The "domain" parameter returns the user to same URL without the domain parameter, not the referrer.

As far as I can tell this re_login was only used by other code removed in this PR, so I'm not worried about removing it.

knowuh: OK, so this won't change logging into Lara (as an author say) via the 'login from <hostname>' links?

knowuh:
This looks fine to me. I checked out the code locally and verified that the basic authentication workflows we are used to still seem to work.

PR #220, opened 2016-10-05
Add v2 run with collaborators

dougmartin:


app/helpers/interactive_run_helper.rb
```diff
@@ -34,6 +34,13 @@ def interactive_data_div(interactive,run)
     if (run)
       data['puturl'] = put_url(interactive,run)
       data['geturl'] = get_url(interactive,run)
+      if run.collaboration_run
+        collaborator_urls = []
+        run.collaboration_run.collaborators_runs(run.activity, run.user).each do |collaborator_run|
+          collaborator_urls.push(get_url(interactive,collaborator_run))
```
scytacki: Why not use full URLs here instead of paths?
The javascript code above has to add the protocol, host, and path. But Rails should be able to add that for you. If `get_url` used `..._url` instead of `..._path` I think this would work.

dougmartin: I was keeping with the get_url/put_url design which is used inside iFramePhone so it can use relative urls.  These urls are also put in the encoded laraParameters url parameter so duplicating the host and port is going to take up a lot of space.

dougmartin: I see what you mean know about using `get_url` since we just use it in one place and use Javascript to add the domain.  I'll update the commit.

PR #221, opened 2016-10-06
Hide question hints by default. Only when (?) clicked.

knowuh:
Question hints were styled with this selector:  `.help-content.screen-only`

`.screen-only` forced `display` to `block`.  By default all divs are display `block`
so it is not necessary to specify it when not `@media print`

[#131758055]

https://www.pivotaltracker.com/story/show/131758055


scytacki:
👍

PR #222, opened 2016-10-06
Completion page sometimes clips next-activity icon.

knowuh:
Small CSS fix.

[#131758617]
https://www.pivotaltracker.com/story/show/131758617


scytacki:
👍

PR #223, opened 2016-10-12
increase upload size for nginx

scytacki:
This fixes a bug found in the draw tool. If drawings are bigger than 1MB then they get rejected by nginx.  I tested a drawing that was bigger than 1MB and it works with this change.
There might be other limits in our stack through.

[#132108989]


knowuh:
Looks good.

PR #224, opened 2016-10-13
without collaborator URLs interactives fail

scytacki:
If there are no collaborators then this code was breaking. And that then broke any CODAP interactives because they didn't get the initInteractive call.

[#132185691]


dougmartin:
Looks good 👍

PR #225, opened 2016-10-26
destroy answers when an embeddable is destroyed

scytacki:
image_questions already did this.
this commit also destroys multiple_choice choices when the multiple_choice
is destroyed.
[#133189183]


knowuh:
Looks good.

PR #226, opened 2016-11-01
this fixes a few issues with the migration

scytacki:
1. the migration needs to clear the learner_url of the IRS otherwise the
student will continue to see the old style interact
2. some of the interactives have been switched over already but not all
of the corresponding run states. The migration should migrate those
older run states anyway.


dougmartin:
Looks good 👍

PR #227, opened 2016-11-09
Full window interactives

pjanik:
Interactive can be configured to use "Full Window Mode". "Click to Play" option has to be tuned on first. "Full window mode" can be set using either normal or ITSI authoring and it also works in both runtimes (normal and ITSI).

Once it's enabled, when user clicks on the interactive image, it automatically opens full window mode. In ITSI mode Labbook buttons are not displayed under the interactive, but they're added to the top bar instead:
screen shot 2016-11-09 at 15 52 10

User can exit full window mode and get back to the interactive without losing any progress. 

I tested it in:
- Chrome
- Safari
- Firefox
- iOS Safari (which was problematic due to some bugs related to virtual keyboard and fixed positions of HTML elements, workarounds are visible in the code)

### DIFF
--- a/app/assets/javascripts/components/itsi_authoring/model_editor.js.coffee
+++ b/app/assets/javascripts/components/itsi_authoring/model_editor.js.coffee
@@ -29,6 +29,7 @@ modulejs.define 'components/itsi_authoring/model_editor',
       'mw_interactive[image_url]': 'image_url'
       'mw_interactive[native_width]': 'native_width'
       'mw_interactive[native_height]': 'native_height'
+      'mw_interactive[full_window]': 'full_window'
 
     getInitialState: ->
       modelsByName: {}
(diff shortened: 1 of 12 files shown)


scytacki:
I'm deploying this to a lara demo server, so we can try it out. Hopefully will be ready before the demo meeting.

An additional part of this work that I forgot in the story, is the integration with the ITSI authoring.  When an ITSI author selects certain interactives from the list, then LARA should automatically enable this full window mode.  This list of interactives is pulled from the portal, so I think this means that portal will need to be updated to support this new option.  There are a couple of other issues with this interactive list so it probably make sense to fix those at the same time. I'll try to pull that all together before the demo meeting.

pjanik:
Checkbox is now removed from ITSI authoring and greyed out in standard authoring when needed. I double checked that copying and export / import preserves full window mode. So, I guess it's ready to be reviewed.

scytacki:
Looks good to me.

PR #228, opened 2016-11-10
Don't add "Copy of" prefix to name of the duplicated interactive

pjanik:
Interactives in ITSI authoring model editor are matched using names. Don't add "Copy of" prefix, so the same of the interactive stays the same. When it's deployed, we can fix old interactives using this script:

```ruby
MwInteractive.all.each do |i|
  new_name = i.name.gsub "Copy of ", ""
  i.update_attributes!(name: new_name)
end
```

### DIFF
--- a/app/assets/javascripts/components/itsi_authoring/model_editor.js.coffee
+++ b/app/assets/javascripts/components/itsi_authoring/model_editor.js.coffee
@@ -29,9 +29,10 @@ modulejs.define 'components/itsi_authoring/model_editor',
       'mw_interactive[image_url]': 'image_url'
       'mw_interactive[native_width]': 'native_width'
       'mw_interactive[native_height]': 'native_height'
+      'mw_interactive[model_library_url]': 'model_library_url'
 
     getInitialState: ->
-      modelsByName: {}
+      modelsByLibraryId: {}
       modelOptions: []
 
     initialEditState: ->
@@ -39,20 +40,22 @@ modulejs.define 'components/itsi_authoring/model_editor',
 
     onSelectChange: (key, value) ->
       # update url and image_url when the select changes
-      model = @state.modelsByName[value]
+      model = @state.modelsByLibraryId[value]
       @valueChanged 'mw_interactive[url]', model.url
+      @valueChanged 'mw_interactive[name]', model.name
       @valueChanged 'mw_interactive[image_url]', model.image_url
       @valueChanged 'mw_interactive[native_width]', model.width
       @valueChanged 'mw_interactive[native_height]', model.height
 
     fetchModelList: ->
+      url = @props.jsonListUrls?.models or 'https://s3.amazonaws.com/sensorconnector-s3.concord.org/model_list.json'
       cachedAjax
-        url: @props.jsonListUrls?.models or 'https://s3.amazonaws.com/sensorconnector-s3.concord.org/model_list.json'
+        url: url
         success: (data) =>
           if @isMounted()
             models = data?.models
             modelOptions = []
-            modelsByName = {}
+            modelsByLibraryId = {}
 
             if models
               models.sort (a, b) ->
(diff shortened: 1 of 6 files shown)


scytacki:
I think it is good to remove the "Copy of". But I think also the matching should be done by URL instead of name.

pjanik:
@scytacki, have you seen my comment in PT story? 
https://www.pivotaltracker.com/story/show/134010161/comments/155059253

I feel it we should allow authors to create two Interactives with the same URL and slightly different config and names if they want to. But that would break select menu again. I'd rather block possibility of having two interactives with exactly the same name, as it would be confusing in select menu and Portal search. And can't see any reason for that, it's easy to come up with unique name in case of need.

But matching by URL is easy, so let me know if you still prefer that.

scytacki:
I added a comment in the PT story yesterday.

pjanik:
@scytacki, I was planning to get back it after wrapping up authoring mode. However, I've just updated this PR. There's a new field in `MwInteractive` model - `model_library_url`. And it's used to match interactives in ITSI authoring select menu.

Also, here's the script to fix old interactives:
```javascript
require "open-uri"
url = "https://itsi.portal.concord.org/interactives/export_model_library"
lib = JSON.load(open(url))
modelByUrl = {}
lib["models"].each do |m|
  if modelByUrl.has_key?(m["url"])
    puts "Warning: interactive #{m["name"]} has the same URL as other interactive"
  end
  modelByUrl[m["url"]] = m
end
count = 0
# editor_mode = 1 => ITSI authoring
MwInteractive.
  joins(interactive_page: :lightweight_activity).
  where('lightweight_activities.editor_mode = 1 AND mw_interactives.model_library_url IS NULL').
  each do |i|
  if modelByUrl.has_key?(i.url)
    ml_url = url + '#' + modelByUrl[i.url]["id"].to_s
    MwInteractive.find(i.id).update_attributes!(model_library_url: ml_url)
    count += 1
  else
    puts "Warning: can't find interactive in model library: #{i.url}"
  end
end
puts "Updated #{count} models"
```
Copy-paste in rails console should work, I've tried it.
Note that it assumes that we want to add model_library_url **only** to the Interactives that are part of the ITSI activities (editor_mode = 1). When I ran it locally, I had bunch of dataset-sync-wrapper interactives that couldn't be found in the model library (what makes sense to me).

Previous script that removes "Copy of" prefix isn't necessary anymore, but it won't hurt either.

scytacki:
```
MwInteractive.find_each do |i|
  next if i.name.nil?
  new_name = i.name.gsub "Copy of ", ""
  i.update_attributes!(name: new_name)
end
```
This is very slow.  This might be faster:
```
MwInteractive.find_each do |i|
  next if i.name.nil?
  new_name = i.name.gsub "Copy of ", ""
  i.update_column("name", new_name)
end
```
That approach doesn't do any callbacks. It seems the callbacks do slow it down because there are lots of labbook calls.

scytacki:
Here is an updated script that I ran on staging:
```
require "open-uri"
url = "https://itsi.portal.concord.org/interactives/export_model_library"
lib = JSON.load(open(url))
modelByUrl = {}
lib["models"].each do |m|
  if modelByUrl.has_key?(m["url"])
    puts "Warning: interactive #{m["name"]} has the same URL as other interactive"
  end
  modelByUrl[m["url"]] = m
end

count = 0
interactives = MwInteractive.
  joins(interactive_page: :lightweight_activity).
  where('lightweight_activities.editor_mode = 1 AND mw_interactives.model_library_url IS NULL')

interactives.find_each do |i|
  if modelByUrl.has_key?(i.url)
    ml_url = url + '#' + modelByUrl[i.url]["id"].to_s
    MwInteractive.find(i.id).update_column(:model_library_url, ml_url)
    count += 1
  else
    puts "Warning: can't find interactive in model library: #{i.url}"
  end
end
puts "Updated #{count} models"
```

PR #229, opened 2016-11-15
put portal_hash in each embeddable

scytacki:
This has been bothering me for a while. This still isn't great, but at least it feels better than what was there before.  Now lightweight_activity doesn't explicitly list all of the embeddable types.

There are still 3 'hashes' for each embeddable though:
portal_hash, to_hash, and export.

- portal_hash needs to match the format expected by the portal
- export should be the most complete, however by itself it doesn't include the type or id of the embeddable during an actual export the type is added and for some things an id is added
- to_hash is used for duplicating

[#134366975]

### DIFF
--- a/app/models/embeddable/image_question.rb
+++ b/app/models/embeddable/image_question.rb
@@ -26,8 +26,7 @@ class Embeddable::ImageQuestion < ActiveRecord::Base
     page && page.visible_interactives.first
   end
 
-  # NOTE: publishing to portal doesn't use this hash. See app/models/lightweight_activity.rb
-  # for the hash used in portal publishing.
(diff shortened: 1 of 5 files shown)


pjanik:
Looks good. 👍 On a similar note, I'd say it would be nice to merge `to_hash` and `export` or at least make `export` use `to_hash` and modify its output in case of need. But it should be another PR. I haven't checked why they're separate, but both methods usually look pretty similar AFAIR.

scytacki:
Thanks! My preference would be to update the 'duplicate' system to use the export and import code.  This way we we'd be catching any issues with export code sooner since the duplicate code is used more often.  But if that is too hard then yes, merging to_hash and export is a good start.

PR #230, opened 2016-11-17
Add questions to tracker API

scytacki:
This adds a question list to the tracker report API.

This path to this API is:
/api/v1/question_trackers/:id/report

It is using the `portal_hash` method which isn't the best because that means this portal_hash is no longer just for the portal.  However this tracker report was already using the portal_hash method on the answers, so at least it is consistent.   One reason it uses the portal_hash is because that includes the id of the question which is needed so it can be correlated with the answer.  The other export mechanisms don't have ids.


pjanik:
I'm not familiar with question tracker yet, so I'm definitely missing context. But changes are pretty small and look okay to me. 👍

PR #231, opened 2016-11-21
WIP: move auto_publish_url from a Thread variable to inside the model

dougmartin:
This has not been tested yet - it was done during the 45 window we had in the Developer Retrospective.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -30,7 +30,6 @@ class ApplicationController < ActionController::Base
   before_filter :portal_login
   before_filter :reject_old_browsers, :except => [:bad_browser]
   before_filter :set_locale
-  before_filter :store_auto_publish_url # to enable auto publishing to build an url from the request object
   after_filter :log_session_after
 
   # Try to set local from the request headers
@@ -338,10 +337,6 @@ class ApplicationController < ActionController::Base
   def handle_unverified_request
     logger.info("unverified_request session:" + session.inspect)
     reset_session
-  end
-
-  def store_auto_publish_url
-    Thread.current[:auto_publish_url] = "#{request.protocol}#{request.host_with_port}"
   end
 
   def unauthorized_response_data(action, resource)
--- a/app/models/pending_portal_publication.rb
+++ b/app/models/pending_portal_publication.rb
@@ -1,5 +1,5 @@
 class PendingPortalPublication < ActiveRecord::Base
-  attr_accessible :portal_publication_id, :auto_publish_url
+  attr_accessible :portal_publication_id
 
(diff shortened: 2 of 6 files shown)


dougmartin:
I remember why the thread variable was used now.  The portal publication is queued automatically inside the model and it uses the request object to build the publishing url and the request object is not available in models in Rails.

scytacki:
I think it would work if the 'auto_publish_url' was added to the PortalPublication model.  I believe the PortalPublication is created synchronously.  And also think it is OK to reuse the same URL on each auto publish. So this URL will be saved the very first time the material is published. That might result in some strange cases with the https/http issue. But I think it might be OK.

dougmartin:
See my comment above yours - the reason I saved it in the thread at startup was the request object is not available in the model so it can't generate the url to save it in the model.  Look at all the breaking tests to see the problem in action.

scytacki:
But in your code I think you are saving the url in the `PendingPortalPublication` not the `PortalPublication`.  I think the `PortalPublication` is always created synchronously from a controller.

scytacki:
Here are a few places that create portal_publications:
https://github.com/concord-consortium/lara/blob/master/app/controllers/lightweight_activities_controller.rb#L194

https://github.com/concord-consortium/lara/blob/master/app/controllers/publications_controller.rb#L69

https://github.com/concord-consortium/lara/blob/master/app/controllers/import_controller.rb#L57

dougmartin:
Three tests are failing but I'm out of time this iteration to work on tech debt so it will have to wait until next time.

scytacki:
This last change of adding the auto_publish_url to activities and sequences won't work because the same activity or sequence can be published to two portals.

scytacki:
Since I'm in this code. I'll summarize what I'm seeing and my current recommended solution.  So right now we make a new PortalPublication each time the material is updated.  The creation of this new publication happens in an `after_update` callback.  Because this is happening in the callback the only way we came up with to access the `auto_publish_url` was through the thread.

My recommendation is that we change how PortalPublication is used.  There would just be one of these for each combination of Portal and Material.  Then we use the PendingPortalPublication model (or a new one), to record information about each time the material is published.  It makes sense to say "the material is published to multiple portals".  Or "which portals is the material published to". So it makes sense that there is just one PortalPublication per portal.  The PendingPortalPublication would become more like a PortalPublicationEvent.  

An hacky intermediate step would be to use the initial PortalPublication that is created when Publishable#portal_publish is called to store the URL.  The portal_publish method seems to always be called from a controller, so it should be possible for these calls to directly pass the auto_publish_url.  Then when later re-publications performed this url they would look at the first PortalPublication of the material for this portal in order to get the url.

PR #232, opened 2016-11-26
Authored state

pjanik:
This PR adds support of authored state for interactives. LARA Interactive API had to be extended (new messages: `authoredState`, `supportedFeatures`) and these changes are documented here: 
https://github.com/concord-consortium/lara-interactive-api/pull/5

Example interactive that saves authored state:
https://concord-consortium.github.io/table-interactive/?mode=authoring

To implement that in practice, I had to change a few things in LARA:

- Now iframe-save always tries to initialize iframePhone connection with iframe interactive and sends `initInteractive` message. State saving doesn't need to be enabled for that. We should probably rename and refactor iframe-saver in the future, as it does a lot more than iframe saving for a long time already.
- Implemented a new React component: `authoring/interactive_iframe`. It takes care of iframePhone communication with iframe. Parent components can just use properties and callbacks to interact with interactive, phone is not necessary.
- Modified ITSI authoring to use `authoring/interactive_iframe` component.
- Implemented `authoring/mw_interactive` component that is used in standard authoring mode and it also uses `authoring/interactive_iframe` to handle iframePhone communication.

Also, I've added a few tests:

- I was thinking for a while whether we could implement specs using real iframe talking to LARA / components. However then I realized that we have mock-iframe-phone which works quite well to test Interactive API. So, I used it in `authoring/interactive_iframe` specs. I couldn't think of any reasonable solution with real iframe + jasmine anyway.
- ITSI Authoring and standard authoring components specs don't use mock-iframe-phone, I test them against `authoring/interactive_iframe` interface (props and callbacks).
- Refactored a bit `iframe-saver` tests to use mock-iframe-phone instead of simple jasmine spies.

@scytacki, could you take a look at it when you get a chance? I guess you're the only person familiar with the whole context.


app/assets/javascripts/iframe-saver.coffee
```diff
@@ -9,11 +17,13 @@ class IFrameSaver
   constructor: ($iframe, $data_div, $delete_button) ->
     @$iframe = $iframe
     @$delete_button = $delete_button
+    @save_state = $data_div.data('save-state');
     @interactive_run_state_url = $data_div.data('interactive-run-state-url') # get and put our data here.
     @collaborator_urls = $data_div.data('collaborator-urls')
     @auth_provider = $data_div.data('authprovider') # through which provider did the current user log in
     @user_email = $data_div.data('user-email')
     @logged_in = $data_div.data('loggedin') # true/false - is the current session associated with a user
+    @authoredState = getAuthoredState($data_div) # state / configuration provided during authoring
```
scytacki: OK I finally figured this out. :) `save_state` is a boolean from the interactive, it is not the state from the interactive_run_state. I was reading it as "save*d* state".
 
It would be good to clear this up. However changing the name of `save_state` would require changing 27 references to save_state and dealing with the export and import code, so old activity exports would still import. 

I'd suggest just changing it in javascript and the dom so it is more clear, and we don't have to change the export and import code. Perhaps `enable_learner_state`.

pjanik: Generally you mean save_state vs saved_state confusion while reading the whole iframe-saver.coffee, right? Then right, it makes sense to me. Especially that now we have more states, authored, global, etc. I'll change it in javascript files.

scytacki: Yes I mean within iframe-saver.coffee.  And possibly in the interactive_run_helper.rb

pjanik:
@scytacki, I've added two new commits. 

The first ones addresses your comments. I went ahead and renamed save_state completely, including db field - it wasn't that bad. The only thing is that import code won't handle previously exported activities, but I'm betting it shouldn't be a real problem, right? And we need solid QA of interactives this time, so I think it's a good moment for this kind of changes. 

The second one makes learner_url optional and independent from learner_state. We're talking about that before. I decided it makes sense to do it now due to the same reason (interactives QA).

scytacki:
Looks great!

PR #233, opened 2016-12-15
Print format fixes

knowuh:
Small CSS fix for printing activities.

* if there is an preview image for an interactive show that in the printout.  Otherwise show a border box with "Interactive" in the center.

* The interactives URL should not be truncated in the print.

@scytacki  if you can have a look.

### DIFF
--- a/app/assets/stylesheets/print-styles.scss
+++ b/app/assets/stylesheets/print-styles.scss
@@ -29,7 +29,7 @@
   }
 
   .interactive-container {
-    .img-container, .interactive-print-text {
+    .interactive-print-text {
       display: block !important;
     }
   }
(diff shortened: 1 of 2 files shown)


scytacki:
I don't understand something. In the ITSI prints it is already showing the image for the interactive.  For example try the print link at the top of this activity: https://authoring.concord.org/activities/5645/single_page/550283ea-dc40-4312-a218-39df83e64641

How is that print working? It seems like from the code in this PR no interactive images were being shown in the print view.

knowuh:
Yes, it was printing in the case of interactives that had a 'click to run' image.  But it was printing the image on top of the image placeholder region -- the rounded rectangle border.   Dan wanted the rounded rectangle border thing to go away if there was an actual image.  

screenshot_12_15_16__4_32_pm

scytacki:
Ah, I understand now.

PR #234, opened 2016-12-19
Fix incorrect scroll position handling in full window mode

pjanik:
@scytacki, it seems to fix the problem in Firefox. I've also checked Chrome and Safari, but I haven't brought my iPad with me today to the coworking space. However if it works in OS X Safari, I hope it should be fine on iPad too.

[#136219369]


scytacki:
Looks good. I'll get it onto staging and add a note to make sure Evangeline tests it on iPad.

PR #236, opened 2016-12-23
Added user idle check in authoring pages [#136522673]

dougmartin:
This adds a new /api/v1/user_check endpoint that returns either null if the user is not logged in or the user object if they are logged in.  A modified third party library was used to perform the idle checks.  When the user comes back from being idle the user_check endpoint is called and if it fails then a blocking modal message is shown with a _blank targeted link to re-login.  Once the user logs back in and comes back to the edit page the user check is automatically called which should then dismiss the modal if they are logged in.

### DIFF
--- a/app/controllers/api/v1/user_check_controller.rb
+++ b/app/controllers/api/v1/user_check_controller.rb
@@ -4,6 +4,6 @@ class Api::V1::UserCheckController < ApplicationController
   skip_before_filter :verify_authenticity_token, :only => :index
 
   def index
-    render :json => {user: current_user}
+    render :json => {user: current_user ? current_user.attributes.slice('id', 'email', 'is_admin', 'is_author') : nil}
   end
 end
--- a/app/controllers/api/v1/user_check_controller.rb
+++ b/app/controllers/api/v1/user_check_controller.rb
@@ -1,0 +1,9 @@
+class Api::V1::UserCheckController < ApplicationController
+  layout false
+
+  skip_before_filter :verify_authenticity_token, :only => :index
+
+  def index
+    render :json => {user: current_user}
+  end
+end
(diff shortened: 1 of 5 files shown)


dougmartin:
@scytacki Let me know what you think of this approach.

scytacki:
@dougmartin have you checked what happens with this when LARA editing is embedded by the portal in an iframe. I suspect the re-login link will be confusing in this case. Since we don't really know which portal to send them to to login at I suspect the re login page shows them a list of portals to login with.  This will be confusing to ITSI teachers.
Perhaps you can identify (either via the parent window, or via a view variable) if the edit page is embedded and in this case just tell the user to reload the page.

dougmartin:
@scytacki Let's talk about this issue this morning.  I'd like to keep from having them reload the page as they would then lose their edits.

app/controllers/api/v1/user_check_controller.rb
```diff
@@ -0,0 +1,9 @@
+class Api::V1::UserCheckController < ApplicationController
+  layout false
+
+  skip_before_filter :verify_authenticity_token, :only => :index
+
+  def index
+    render :json => {user: current_user}
```
dougmartin: For future proofing in case sensitive info is added to current_user whitelist the current attributes to name, email, isAdmin, etc.

dougmartin:
@scytacki I whitelisted the user attributes and simplified the timeout message to "Your login session has timed out.  Reload this page to log back in and continue editing."

scytacki:
This looks good for now.  Internal authors might get frustrated because they will be using LARA outside of an iframe, so reloading the page will take them to a lara home screen, but I think this is still better than letting them think they can continue authoring.

scytacki:
Hmm I see that the build is failing though. It looks like we should update the version of Bundler and see if that fixes it: https://github.com/bundler/bundler/issues/3367  I'll give that a try on master since it seems to be having the same issue.

scytacki:
Updating bundler in the master branch fixed this issue.  @dougmartin can you merge master into this branch and we'll see if the tests pass?

dougmartin:
@scytacki It passed once I rebased to master and pushed.

PR #237, opened 2017-01-11
Added automatic fullwindow and no snapshots option support [#96093804] [#136598267]

dougmartin:


app/views/embeddable/labbook_answers/_lightweight.html.haml
```diff
@@ -5,15 +5,18 @@
...
-    != embeddable.prompt
-  %button.lb-action-btn{type: 'button'}
-    %span.lb-action-btn-label
-      %i.fa.fa-camera
-      = embeddable.action_label
-  %button.lb-album-btn{type: 'button'}
-    %span.lb-album-btn-label
-      = t('OPEN_ALBUM')
+
+  - if !embeddable.interactive.no_snapshots
```
scytacki: There is already code which is deciding to show or hide the labook. Can this check be moved into that code?  By doing this it should also take care of making sure the labbook doesn't show up in the reports.

scytacki:
Also there are some failing tests. I think it is because you can snapshot not just MwInteractives but ImageInteractives and video interactives too.

dougmartin:
@scytacki I moved the check for no_snapshots to the existing helper code and added support for the save_interactive_state flag in the model library

scytacki:
It will be redundant but you also need to update this method: http://github.com/concord-consortium/lara/blob/2a318de0ada7a8719411a578063c02f984a69093/app/models/embeddable/labbook.rb#L180
This file provides a bit more info about the various filtering of embeddables:
https://github.com/concord-consortium/lara/blob/master/documentation/embeddable-interactive-filtering.md#L1

dougmartin:
@scytacki I added the no_snapshot check to the labbook model

PR #238, opened 2017-01-11
Added cancel check to labbooks [#130837611]

dougmartin:


dougmartin:
@scytacki For review with https://github.com/concord-consortium/labbook/pull/2

app/assets/javascripts/labbook.coffee
```diff
@@ -74,6 +79,12 @@ class LabbookController
   showDialog: ->
     @$dialog.dialog('open')
 
+  onDialogClosing: (event) ->
+    if @implementsIFramePhone and @hasUnsavedState
```
scytacki: Instead of relying on the labbook to keep its state up-to-date, can you just have LARA ask the iframe for its state here, and block until it gets a response if the labbook implements iframe phone.  Then it will be the labbooks responsibility for telling the user about unsaved state . This will mean you'll need a way to handle dialogClosing I'm not sure if the dialog library supports that.

scytacki:
Can you add a comment in the labbook.coffee (or some other place), describing the protocol.  After talking with you my understanding is:
- labbook sends 'connected:true'
- if labbook goes to a page that doesn't need close protection then it sends 'connected:false'
- if labbook goes back to a page that needs close protection then it sends 'connected:true'
- if user clicks close and connected:true has been sent, then lara sends labbook 'canClose'
- labbook checks with user and sends back 'close' if the user approves

dougmartin:
@scytacki I added a big comment at the top of the file outlining the protocol: https://github.com/concord-consortium/lara/pull/238/commits/afc1809e0daa8bf8f1bcbdcf56a61c735c472d78

dougmartin:
@scytacki I renamed the variable.

PR #239, opened 2017-01-13
Add search and pagination to users page.

knowuh:
Adds pagination and searching to the admin users page.

[#137615501]

https://www.pivotaltracker.com/story/show/137615501


scytacki:
looks good 👍

PR #241, opened 2017-01-25
A student can upload an image from their desktop as an answer to an image question [#137911085]

dougmartin:


scytacki:
This looks good. The only reason I'm not merging it, is because we are getting ready to release v1.17.0, once that release is out then I'll merge this in.

PR #242, opened 2017-02-08
Add class hash and extend initInteractive [#138853633]

dougmartin:
Adds the public class hash and class info url to the run state, passed by the portal.

Also extends the InitInteractive call to pass
  * Class info url
  * Interactive Id
  * Interactive Name
  * Auth Info (same as returned in getAuthInfo)

### DIFF
--- a/app/controllers/lightweight_activities_controller.rb
+++ b/app/controllers/lightweight_activities_controller.rb
@@ -22,6 +22,11 @@ class LightweightActivitiesController < ApplicationController
   # These are the runtime (student-facing) actions, show and summary
 
   def show # show index
+    if params[:class_info_url]
+      @run.class_info_url = params[:class_info_url]
+      @run.save!
+    end
+
     authorize! :read, @activity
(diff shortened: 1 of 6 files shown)


scytacki:
If you agree that we don't need the public_class_hash, as I was suggesting in the Portal PR: https://github.com/concord-consortium/rigse/pull/288
Then it should be removed here too.

If an interactive wants a public unique ID for the class they can use the class_info_url. Or if they want a private ID then they can make a request to the class_info_url.

app/controllers/lightweight_activities_controller.rb
```diff
@@ -22,6 +22,11 @@ def index
   # These are the runtime (student-facing) actions, show and summary
 
   def show # show index
+    if params[:class_info_url]
+      @run.class_info_url = params[:class_info_url] if params[:class_info_url]
```
scytacki: this if is redundant.

scytacki:
👍

PR #243, opened 2017-02-10
Changes to support a portal connection

scytacki:
These changes make it possible change which port LARA binds too.
Also there is a compose overlay for connecting with the portal network.

More detail can be seen in this PR:
https://github.com/concord-consortium/rigse/pull/290


knowuh:
Awesome!

PR #244, opened 2017-02-21
Speed up question tracker report

scytacki:
Previously this was loading all of the answers from each of the requested runs. Then it would check if the answer was associated with one of the questions associated with the tracked question.

Now it does that with a database query.  It makes 2 assumptions:
- all of the questions associated with the tracked question are the same type of question.
- all question types define a `has_many :answers` which can be inspected to find the answer type and the foreign key in that answer.  
With these assumptions it can just look for Answers of that type that are associated with one of the questions, associated with the tracked question.


knowuh:
Looks great.

PR #245, opened 2017-02-23
Upgrade Ruby to 2.2.6

pjanik:
Automated tests are passing and I've done a bit of manual tests. This seems less risky than Portal upgrade, as I didn't have to update any of the most important gems.


pjanik:
I've created a new image `pjanik/docker-rails-ruby226`, updated Docker files, and deployed demo server:
http://ruby-22.lara.demo.concord.org/
So now it can be deployed to staging too.

The new image should be moved to concordconsortium org, but it seems I can't do it.

scytacki:
Great!

PR #246, opened 2017-03-07
Hide right logo in NGSS-Assessment theme

pjanik:
@ddamelin asked for this tiny change. It hides logo in the top-right corner.


ddamelin:
Does there need to be a space between .site-logo.logo-r

pjanik:
@ddamelin, no, because:

- `.site-logo .logo-r` selects `.logo-r` element which is a descendant of `.site-logo` element.
- `.site-logo.logo-r` selects an element which has two classes: `.site-logo` and `.logo-r`

The logo that you want to hide has those two classes at the same time.

PR #247, opened 2017-03-12
Generic interactive state

pjanik:
This PR adds support of a new answer type - "interactive". Now, every interactive that has learner state enabled will send this state to the Portal. Previously we were sending only reporting URL when it was available. Reporting URL is still supported and when it's available, it'll be send to the Portal instead of the interactive state JSON.

Note that the interactive state JSON that we send to the portal is actually an argument to the `initInteractive` call that will be made later by consumers of this state (e.g. report app). So, it follows LARA Interactive API format. It ensures that Portal and external apps don't need to know anything about this API and format. They'll just pass this state to the interactive later. Only LARA and Interactives care about it.

Also, there's one small, extension to the LARA Interactive API - "report" mode has been added (other modes are "runtime" and "authoring").

This PR is related to https://github.com/concord-consortium/rigse/pull/303.

### DIFF
--- a/app/assets/javascripts/components/authoring/mw_interactive.js.coffee
+++ b/app/assets/javascripts/components/authoring/mw_interactive.js.coffee
@@ -2,7 +2,7 @@
 
 modulejs.define 'components/authoring/mw_interactive',
 [
-  'components/authoring/interactive_iframe',
+  'components/common/interactive_iframe',
 ],
 (
   InteractiveIframeClass,
@@ -20,7 +20,6 @@ modulejs.define 'components/authoring/mw_interactive',
         saving: false
         message: ''
       }
-
 
     componentDidMount: ->
       window.addEventListener 'beforeunload', @onBeforeUnload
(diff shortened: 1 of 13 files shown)


ddamelin:
Does this mean that if I put in a Lab interactive that saves state and I check in LARA authoring that it has a report URL, that the teacher can see the saved state of the Lab interactive in the report?

pjanik:
@ddamelin, yes and no. :) This PR definitely makes it possible on LARA / Portal side. But you would actually need to check "state saving", not "report URL". Now every interactive that saves learner state is reportable too (or will be when this PR is merged). However, Lab would need to be updated too:

1. It would need to support one new iframe-phone call: `initInteractive`. (1h - 2h max)
2. I guess Lab doesn't serialize its model when it sends its state back to LARA. So, even now when you enable learner state saving, LARA won't really restore Lab model correctly the next time you open an activity. However, it might work with state that is saved within Interactive JSON (not model JSON), like tables, parameters etc. I guess we could just include model state within interactive state, but it's not backward compatible change. I think they're serialized separately because of the interactives page - there're separate interactive JSON and model JSON editors. I'd need to workaround it somehow. (0.5 day?)
3. Also, I can imagine Lab supporting authored state. It could be quite simple - authored state would be used if there's no learner state. If there's learner state, it'd be used instead. I could be useful for some easy tweaks of Lab interactives within LARA like moving atoms around, parameters tweaks etc. (0.5 day?)

Probably it'd be nice to have 1-1.5 day for those updates. If you find it useful, I'd ask @jengoree how to schedule it. I'd be happy to work on it, as I feel it'd be a nice feature.

pjanik:
I've just pushed two new commits to update summary page (student report). So, I think it's ready to review now.

PR #248, opened 2017-03-22
Remove remaining references to svg-edit

scytacki:
This would cause hard to find problems if the ENV variable was not configured right


config/application.rb
```diff
@@ -67,8 +67,6 @@ class Application < Rails::Application
     config.middleware.use Rack::Environmental,
                             :delayed_job => { :url => /delayed_job/,
                                               :style => :none},
-                            :svg_edit    => { :url => /svg-editor/,
```
knowuh: I can't figure out why this would have been here.

knowuh:
Thanks @scytacki  this looks good.  So great that you have the presence of mind to go back and remove unused code.  👍

config/application.rb
scytacki: I think svg edit added its assets in a weird way that required this.

PR #249, opened 2017-03-22
Dev env fixes

scytacki:
This fixes a problem caused by ruby 2.2 update

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,7 +23,8 @@ end
 
 group :development do
   gem "binding_of_caller", "~> 0.7"
-  gem "bullet", "~> 4.7"
+  # bullet 5.5 does not support AR 3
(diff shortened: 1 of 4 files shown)


pjanik:
Looks good. 👍  I was going to ask what was the problem, just out of curiosity. But I you explained that during standup if I remember correctly.

scytacki:
When importing an activity (only in dev environment), a 500 error would be returned. And there would be no info about it.  I fixed the exception handling on import so I could see the error (after messing around with log statements too long). The error was in active record.  The import code was using a simple `page.interactive_items.size` and that was causing this weird error.  After trying to debug it for awhile, then I googled it. And I found an issue reported that bullet caused this problem with ruby 2.2.  I didn't track it down but there was reference to using 'reflection vs. reflection()' so perhaps `reflection` is special in ruby 2.2 but that is just a hunch.

PR #250, opened 2017-03-24
Fix interactive printing

scytacki:
Remove the extra div with a image in the background, this doesn’t work when printing.
Fix the img-container layout so it is visible when the iframe is not.
Add margins so the printed interactives line up better with the questions.
Fix ITSI printing so the interactives don’t shift around in relation to questions, and they are centered.
Fix some typos in comments about page layout labels.
Make note of issue with bold text when using the print_blank path.
Add a Node script to generate 24 page activity to check various layouts and interactive settings
Slightly widening the questions in full width mode.
[#142314353]

### DIFF
--- a/app/assets/stylesheets/mw-runtime-base.scss
+++ b/app/assets/stylesheets/mw-runtime-base.scss
@@ -425,7 +425,10 @@ div.intra-page-nav {
   .question {
     margin: 20px;
     display: inline-block;
-    width: 45%;
+    // the questions sit in two columns in the full width layout.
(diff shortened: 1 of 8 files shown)


knowuh:
Looks good to me @scytacki  -- could you share a link to a pdf or something so @ddamelin can check?

scytacki:
Good suggestion.  I put a link to the old PDFs in the story: https://www.pivotaltracker.com/story/show/142314353 Since the code looks good, I'll get it deployed to staging and @ddamelin can check it there...

PR #251, opened 2017-03-28
Added new GEODE theme.

emcelroy:


scytacki:
I'll merge this in, but the addition of the `geode-footer` styles in the _runtime-footer.scss seems like it should be handled differently in the future. We don't have a good solution for it though, so it is fine for now.

PR #255, opened 2017-05-24
Save the portal publication data.

scytacki:
I'm running into a problem on authoring.staging with publishing an activity.  I didn't find a good way to track this down.  So I'm adding this recording of the publication data. 

In the future it could be useful to provide an admin with a way to view this publication data as well as the response from the portal after the publication.

Additionally in the future it could be useful for authors to compare the publication data to see what has changed from one publication to another. 

[#146040445]

### DIFF
--- a/app/models/portal_publication.rb
+++ b/app/models/portal_publication.rb
@@ -1,6 +1,7 @@
 class PortalPublication < ActiveRecord::Base
 
-  attr_accessible :portal_url, :response, :success, :publishable, :publication_hash, :publication_time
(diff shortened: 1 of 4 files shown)


scytacki:
@jsimpson-cc have you had a chance to look at this?

scytacki:
Right now there is no UI.

I did figure out after adding this, that the publication data can be also be recovered from the rails logs of the portal. I still think having it in the lara model makes it easier to track down problems though.

PR #256, opened 2017-05-25
Document additional steps for dory proxy config on linux.

jsimpson-cc:
Document steps for linux proxy config.


knowuh:
Looks good.

PR #257, opened 2017-05-31
Fix nil open response answers sent to portal.

jsimpson-cc:
[#143324403]

Do not send nil answer response text to portal.

### DIFF
--- a/app/models/embeddable/open_response_answer.rb
+++ b/app/models/embeddable/open_response_answer.rb
@@ -15,6 +15,15 @@ module Embeddable
     after_update :send_to_portal
     after_update :propagate_to_collaborators
 
+    #
+    # Only send to portal if answer text is not nil
+    #
+    def send_to_portal
+        if ! self.answer_text.nil?
+            super
+        end
+    end
+
     def self.by_question(q)
       where(:open_response_id => q.id)
     end
--- a/spec/support/shared_examples/answer.rb
+++ b/spec/support/shared_examples/answer.rb
@@ -32,6 +32,7 @@ shared_examples "an answer" do
   describe "send_to_portal" do
     describe "with a run" do
       it "should call run.send_to_portal(self, nil)" do
+        answer.answer_text = "foo"
         allow(answer).to receive_messages(:run => run)
         expect(run).to receive(:queue_for_portal).with(answer)
         answer.send_to_portal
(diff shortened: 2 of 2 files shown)


jsimpson-cc:
Sorry, need to fix spec tests.

scytacki:
This won't handle the case if someone calls: `Run#submit_answers_now`. However I've looked into this more and now I think your original approach was correct as long as we verify and possibly fix something in the portal.

Here is what I found.  LARA Answers have an answered? method, as far as I can tell this method is only used to determine completion of the Activity in LARA.  In the case of a open response, a blank answer will be not answered.  So even if a student answers it and then blanks it out, it will start out answered and then become unanswered. I think this metric of completion is correct: a student that answers a question and then blanks it out should be seen as having an incompletely activity.

This completion amount in LARA should match what the Portal reports for completion.  I haven't checked what the Portal uses for completion of open response answers, can you check and report back?  If the portal is not treating blank answers as incomplete then we should change that.

Once the completion metric is consistent, then it seems a good approach to have the Portal accept nil answer_text and treat it as an empty string. This means the Portal will be more robust and all the changes will be in the Portal.

It would be useful to add a LARA spec test to test some of the open_response_answer methods when the answer_text is nil.  And then around these tests can be documentation describing how the javascript code will send a nil answer_text if the answer is unanswered and the user clicks in and out again.

jsimpson-cc:
Saveables in portal also have an answered? method, which is not necessarily implemented the same as on the lara side. 
But they have a submitted? method. This is used in the report/util_learner classes to determine completion percent.

scytacki:
This unfortunately seems like a rat hole. I don't think it is good to modify submitted? because semantically it isn't right.  The best solution would be to add a new method completed? And switch the completion calculation to use that. 

At this point my suggestion is to just allow the portal to handle nil answer_text. And we will have to come back to the completion issue later. 

However if you really want to keep going on it. I'll support you in adding the new completed? method.  There are only 4 definitions of submitted so that side of things might be pretty simple.  But there are 24 references to submitted?.   And there is the issue of `#submitted_answer` if the answer is blank but it isn't a submittable question, it seems we should return something like 'not completed'.

jsimpson-cc:
I don't think all of that is necessary. Please see my latest push to portal.
https://github.com/concord-consortium/rigse/pull/323

I think it is sufficient to override #answered? in the subclasses that require it. This PR against LARA can be closed and both issues addressed in what is pushed to that branch on rigse.

jsimpson-cc:
Closing this to fix on portal side.
https://github.com/concord-consortium/rigse/pull/323

PR #262, opened 2017-07-06
Fix reporting too few submissions in sequence runlists.

knowuh:
Previous implementation assumed that there was only one submission per activity.  Activities may have several arg block sections.

### DIFF
--- a/app/services/dashboard_sequence_runlist.rb
+++ b/app/services/dashboard_sequence_runlist.rb
@@ -80,14 +80,18 @@ class DashboardSequenceRunlist
     if num_tries < 1
       return []
     end
-    submission = submissions.last
-    return {
-      page: submission.interactive_page_id,
-      pageIndex: submission.interactive_page.position,
-      tryCount: num_tries,
-      numQuestions: 4,
-      answers: submission_answers(submission)
-    }
+    submissions = submissions.group_by(&:interactive_page_id)
+    submissions.map do |page_id, sumbissions|
+      num_tries = sumbissions.length
+      submission = sumbissions.last
+      {
+        page: page_id,
+        pageIndex: submission.interactive_page.position,
+        tryCount: num_tries,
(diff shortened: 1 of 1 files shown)


knowuh:
@scytacki if you get a moment, I would appreciate a quick review this. 

The previous implementation would only return one answer per activity run.  

In this case, we return one answer per page.  It might be that we will have to revisit this later if we want to have more than one argumentation block per page.  But this matches assumptions baked into dashboard reporting.

app/services/dashboard_sequence_runlist.rb
```diff
@@ -80,14 +80,18 @@ def run_answers(run, submissions_created_after)
...
-    submission = submissions.last
-    return {
-      page: submission.interactive_page_id,
-      pageIndex: submission.interactive_page.position,
-      tryCount: num_tries,
-      numQuestions: 4,
-      answers: submission_answers(submission)
-    }
+    submissions = submissions.group_by(&:interactive_page_id)
+    submissions.map do |page_id, sumbissions|
```
scytacki: It is confusing that the loop variable is the same as the variable being looped over.
Perhaps the loop variable could be changed to `page_submissions`

knowuh:
Thanks, that was pretty confusing.  Nice catch.

app/services/dashboard_sequence_runlist.rb
```diff
@@ -81,15 +81,15 @@ def run_answers(run, submissions_created_after)
       return []
     end
     submissions = submissions.group_by(&:interactive_page_id)
-    submissions.map do |page_id, sumbissions|
-      num_tries = sumbissions.length
-      submission = sumbissions.last
+    submissions.map do |page_id, page_sumbissions|
```
scytacki: submissions is misspelled.

knowuh:
Third times a charm 🤞

PR #265, opened 2017-08-14
Create new themes

ddamelin:
Please take a look to make sure production.rb file was modified correctly and in a way that will add the new ipums-terra them.

Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,8 +138,8 @@ If you are going to deploy to a server with load balancing enabled (production),
 
 ## Docker and docker-compose for developers:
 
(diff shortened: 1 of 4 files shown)


ddamelin:
Oops I realize that I make this branch off of the another when where I had commits to update the readme files. I'll close that pull request, and we an just use this one.

ddamelin:
Anyone have ideas about why the build of this branch is failing? Travis gets to Jasmin server started, and then just hangs, eventually terminating. @knowuh @scytacki ?

PR #266, opened 2017-09-08
Fix view work links for CODAP interactives

scytacki:
I finished this a while ago and then forgot to make a PR for it.

This also fixes the completion state for interactives that don't have a report_url.
[#150113706]
[#150113578]

### DIFF
--- a/app/models/interactive_run_state.rb
+++ b/app/models/interactive_run_state.rb
@@ -60,11 +60,15 @@ class InteractiveRunState < ActiveRecord::Base
     #   It's useful if state can be saved in the URL or is kept by the interactive itself (e.g. CODAP / docstore)
     # - Otherwise, interactive state JSON is sent to the Portal. Later, the same state will be provided to teacher report
     #   and sent to the interactive using LARA Interactive API.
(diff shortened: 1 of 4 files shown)


pjanik:
I guess I wrote the original logic there. I haven't worked with CODAP documents too often, so it's very likely I had missed this use case. Changes look good to me, but it seems tests failed. I'm not entirely sure what's the current status of tests (if it's related or random fail), so I'm not clicking merge.

jsimpson-cc:
I'm still not very familiar with the LARA codebase so I suppose this is okay. That test failure looks like what we see with Portal on travis using trusty now. There's a fix for the db access issue on portal's master branch but there might be other issues after that is resolved.

jsimpson-cc:
If these tests are using selenium-webdriver, that will be the next issue after fixing db access since trusty runs Firefox 48+, you'll need geckodriver. I have a branch on portal which starts to address this but there is still work to do. On portal see branch 150757054-fix-travis if you'd like more info there. Or use the .travis.yml workaround to go back to the Ubuntu precise image.

PR #267, opened 2017-10-12
Update Shutterbug to v1.0.0

pjanik:
Update Shutterbug to v1.0.0. API is backward compatible, so I didn't have to change much. I tested it locally and it works.

Note that we would have to update `SHUTTERBUG_URI` ENV variable to `https://fh1fzvhx93.execute-api.us-east-1.amazonaws.com/production`. Or leave it blank to so Shutterbug JS lib will use the default one (the same URL).


pjanik:
Hm, some tests failed, but it doesn't seem related. Is it possible that these are some random fails? I haven't been working on LARA for a long time, so not sure what's the current tests status.

scytacki:
Is it hard to keep using the rails assets approach? I thought that was pulling auto generated asset packages from npm. But perhaps it is more complex than that.

scytacki:
It looks like the test are not random.  The build before passed and these two builds both failed the same tests.  My guess is that the failures are related to the jquery assets gem being removed as part of this work.

pjanik:
> Is it hard to keep using the rails assets approach? I thought that was pulling auto generated asset packages from npm. 

It's using Bower, not NPM. I've removed it from Shutterbug 1.0.0 on purpose, as we don't use it anymore in any new project. I could add it again and publish to Bower, but frankly speaking I can't see any advantage. We have manual copies of JS libraries in LARA anyway. And Bower doesn't seem to be popular these days, especially in the enviroment we usaully use (React, webpack). So, I'd avoid adding confusion and forcing developers to remember about publish to NPM and Bower all the time.

scytacki:
Yes it makes sense to drop Bower support. OK we'll go with the included js file for now, and when we Rails updated we can start using npm for this.  There is still the issue of the failing tests, but Piotr needs to get back to some other work so I'm going to see if another developer can wrap this up.

pjanik:
Tests are fixed. The problem was that new Shutterbug code wasn't ES5 compatible. PhantomJS error wasn't very helpful and I have no idea why it was failing on `$.waypoint` call. But I made sure ES5 was the culprit.

Then I had issues with Jasmine tests that didn't want to start, but upgrade to v2.8 seems to help.

PR #269, opened 2017-10-17
Added navigation iframe api endpoint [#151917842]

dougmartin:


app/assets/javascripts/save-on-change.coffee
```diff
@@ -179,21 +179,29 @@ class @ForwardBlocker
     @binding.bind 'navigate_away', (e,opts) =>
       @navigate_away(opts.click_element, opts.action_to_perform)
 
-  prevent_forward_navigation_for: (blocker) ->
-    if blocker not in @blockers
-      @blockers.push blocker
+  get_blocker_index: (blocker) ->
+    firstBlocker = (b for b in @blockers when b.blocker is blocker)[0]
+    @blockers.indexOf(firstBlocker)
```
scytacki: This is confusing. There is an array called "blockers", but its values are actually objects that have a field named `blocker`, and a field named `disabledForwardNavMessage`.   I think calling the array `blockers` makes sense. I think the `blocker` field is actually the dom element, so I'd call that `element`.

app/assets/javascripts/save-on-change.coffee
```diff
@@ -179,21 +179,29 @@ class @ForwardBlocker
...
+
+  get_latest_disabled_forward_nav_message: ->
+    (b.disabledForwardNavMessage for b in @blockers when b.disabledForwardNavMessage?).reverse()[0]
```
scytacki: The name `disabledForwardNavMessage` is confusing to me. It sounds like a flag.  I think just `message` works here.  Assuming these objects are called Blockers then this is the message of the Blocker which makes sense to me.

app/assets/javascripts/save-on-change.coffee
```diff
@@ -179,21 +179,29 @@ class @ForwardBlocker
...
+    index = @get_blocker_index(blocker)
     if index > -1
       @blockers.splice(index,1)
       @update_display()
 
   navigate_away: (click_element, action_to_perform) ->
     if @block_for_element(click_element)
       $('.question').addClass('did_try_to_navigate')
-      modalDialog(false, t('PLEASE_SUBMIT'))
+      message = @get_latest_disabled_forward_nav_message() or t('PLEASE_SUBMIT')
```
scytacki: Instead of this, why not have all blockers have a message. This way the question objects would just pass in `t('PLEASE_SUBMIT')` as their message. Then the 'get_latest...._message' could be changed to just get the latest blocker object and the message would be its message.

scytacki:
@jsimpson-cc I'm tagging you so you know this is going into LARA.  These "blockers" are used to prevent forward navigation. They are currently used by the ArgBlocks and Prediction questions. ArgBlocks provide feedback to the student, so the forward navigation is blocked until the user submits their answers.  Prediction questions use them because we want students to submit their answer which then gets locked (they can't change it), before they can go onto the next page.

The PR allows interactives to do the same thing. So an interactive can block the forward navigation until the user has 'completed' it.

ddamelin:
I'm interested in the mechanism, as we have been considering a feature like this for IS 2.

scytacki:
One thing missing from the code here, is support for 'click to play' interactives.  The way the code in this PR works, these 'click to play' interactives will not block the forward nav until they are actually loaded.  Supporting that would require more changes than Doug had time for on his current project. If IS2 interested that would be great then we might be able to wrap this up better.

dougmartin:
I updated the code last week but didn't have time to test it until this morning.

scytacki:
@dougmartin it looks like the jasmine tests are failing here: https://travis-ci.org/concord-consortium/lara/builds/294902145?utm_source=github_status&utm_medium=notification But perhaps that is a random failure?  It seems most of the time the jasmine tests don't even run.

app/assets/javascripts/save-on-change.coffee
```diff
@@ -173,34 +173,34 @@ class @ForwardBlocker
...
+    index = @get_blocker_index(element)
     if index > -1
       @blockers.splice(index,1)
       @update_display()
 
   navigate_away: (click_element, action_to_perform) ->
     if @block_for_element(click_element)
       $('.question').addClass('did_try_to_navigate')
-      message = @get_latest_disabled_forward_nav_message() or t('PLEASE_SUBMIT')
+      message = @get_latest_message() or t('PLEASE_SUBMIT')
```
scytacki: Since every blocker should have a message now. The `or ...` can be removed. This way the save on change is completely independent of the prediction and c-rater code.

PR #271, opened 2017-11-15
Data games theme and fixes

emcelroy:
Added a new theme for Data Games, and fixed general problems with carousel questions and the sticky header feature.


kswenson:
The notion that class selectors should be preferred over id selectors is certainly out there (cf. https://csswizardry.com/2011/09/when-using-ids-can-be-a-pain-in-the-class/).

jsimpson-cc:
Thanks Kirk. I wonder if that article is more about defining large sets of css styles rather than obtaining attribute info of a known element? Do you think it makes the case that the jquery selector `$('parent > child')` is more "reusable" than `$('#some-id')`? Would using an id have avoided the original bug?

Anyway I don't mean to derail this PR with this question. I'll approve. I can merge unless we want to wait on @scytacki 's approval.

emcelroy:
Sorry I missed this conversation till now. In general, I use IDs, or augment classes with IDs, in cases like this. I’d actually prefer that over the $(‘parent > child’) solution. It’d be a little less complex and slightly clearer IMO.

I may be wrong, but from what I’ve seen, IDs are infrequently used by most developers at CC. So I thought I’d go along with that mode in the LARA code.

I’m happy to change this to use an ID though if nobody has a strong objection.

jsimpson-cc:
> I’m happy to change this to use an ID though if nobody has a strong objection.

No strong opinion, I was just asking a question. And I didn't mean to hold up this PR.
Did @scytacki also want to review and approve this?

scytacki:
@emcelroy is most of the code in `dg-runtime-base.scss` a copy of some other file? 
If so could you extend some base file instead of copying all of this?  We've been trying to keep themes simple. If they get complex then Evangline will need to an activity using each theme each time we update LARA. Most of the themes just change colors and perhaps a footer image.

A super simple one: https://github.com/concord-consortium/lara/blob/master/app/assets/stylesheets/theme-has-astro.scss

One with more customizations:
https://github.com/concord-consortium/lara/blob/master/app/assets/stylesheets/theme-ipums-terra.scss

This is probably the most complex one:
https://github.com/concord-consortium/lara/blob/master/app/assets/stylesheets/theme-ngss-assessment.scss

I realize that these three all based on `cc-runtime`. Does the data games theme need to be different than those?

emcelroy:
@scytacki: Ah, sorry I didn't quite get how themes worked exactly. I think I've made the necessary changes. 

The new theme is meant to extend mw-runtime. There are only a couple customizations. 

Does it look better now?

scytacki:
Looks great!

eireland:
I like IDs! Makes for easier selector search for automating tests.
Minimally, unique classes also work.

Evangeline

PR #274, opened 2018-01-20
Render more than one interactive on a single page (in mult-page layout)

pjanik:
The first, easier solution of multiple interactives being displayed on a single page. 
This PR removes `show_multiple_interactives?` as it doesn't seem to be any reason to keep it. Also, it fixes authoring mode where multiple interactives were not displayed correctly. Interactives are not pinned, as it could make it impossible to see the bottom one when the screen height is smaller than the sum of interactives height.

I've tested generic theme and HAS water. It seems to work fine in 40-60 and full-width mode. 30-70 also works, but this layout has some issues in itself (I've checked that in the production branch).

### DIFF
--- a/app/models/interactive_page.rb
+++ b/app/models/interactive_page.rb
@@ -88,15 +88,6 @@ class InteractivePage < ActiveRecord::Base
     interactives.select { |i| !i.is_hidden }
   end
 
-  # can this page display multiple interactives?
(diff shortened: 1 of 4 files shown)


scytacki:
Looks good. I made a PT story for this specific change so I could track it during the release process: https://www.pivotaltracker.com/story/show/154535325

PR #275, opened 2018-01-26
Multiple interactives on one page, interleaving with questions

pjanik:
@scytacki, @knowuh, I'm opening this PR with the latest changes, but it's not ready to merge. The current status is:

1. All the interactive types should be supported (MW/Iframe, Image, Video).
2. Basic authoring should work (and authoring UI is tweaked so things look reasonable).
3. Basic runtime should work. 
4. Copy, import, and export should work.
5. I've managed to have all the RSpec tests passing (started with ~50 failures).
6. I've tried Labbook and image questions + interactives in a few scenarios (especially page copying), but I didn't have much time to test it.

~~TODO~~ (DONE)
1. Fix Jasmine tests.
2. Test various cases, untypical scenarios, ITSI layout (that might be broken), and other edge cases. I guess it'll produce a bunch of new points in TODO list.
3. Remove `InteractiveItem` class (it might break more tests, but it would be useful to do for sure).
4. Write migration that replaces all the existing `InteractiveItems` with `PageItems` (and `section="interactive_box"`).

### DIFF
--- a/app/models/interactive_page.rb
+++ b/app/models/interactive_page.rb
@@ -28,11 +28,7 @@ class InteractivePage < ActiveRecord::Base
   # See https://www.pivotaltracker.com/story/show/60459320
   validates :text, :sidebar, :html => true
 
-  # InteractiveItem is a join model; if this is deleted, it should go too
-  has_many :interactive_items, :order => :position, :dependent => :destroy, :include => [:interactive]
-
-  # Like InteractiveItems, PageItems are join models, so they should not
-  # survive the deletion of associated instances of InteractivePage.
+  # PageItem is a join model; if this is deleted, it should go too
   has_many :page_items, :order => [:section, :position], :dependent => :destroy, :include => [:embeddable]
 
   # Interactive page can register additional page sections:
--- a/app/models/interactive_page.rb
+++ b/app/models/interactive_page.rb
@@ -18,9 +18,7 @@ class InteractivePage < ActiveRecord::Base
 
   EMBEDDABLE_DISPLAY_OPTIONS = ['stacked','carousel']
 
-  INTERACTIVE_TYPES = [{ :name => 'Image',  :class_name => 'ImageInteractive' },
-                       { :name => 'Iframe', :class_name => 'MwInteractive' },
-                       { :name => 'Video',  :class_name => 'VideoInteractive' }]
+  INTERACTIVE_BOX = 'interactive_box'
 
   validates :sidebar_title, presence: true
   validates :layout, :inclusion => { :in => LAYOUT_OPTIONS.map { |l| l[:class_val] } }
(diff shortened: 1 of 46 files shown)


scytacki:
@pjanik A few things to check with this:
- Does LARA send these items correctly to the portal when publishing?
- Does LARA send the answers to the items correctly when sending answers?
- Does the question numbering in the LARA report and portal report make sense? Some interactives save state, some don't, I believe the current numbering reflects this. And there was a lot of tweaks to fix the numbering as related to interactives.
- When looking at the ITSI stuff, check the way that LabBooks are linked with interactives. And then the labbook buttons were rendered below their interactive regardless of where the labbook item was in the assessment list.  I don't remember the details, but that seems like something that this change could break.

scytacki:
@dougmartin you should also look at this PR to see how it affects the CollabSpace's use of the LARA export to generate its 'prompts'.

pjanik:
@scytacki, I think it's ready now.

> Does LARA send these items correctly to the portal when publishing?
> Does LARA send the answers to the items correctly when sending answers?

I checked following cases:

1. I created an activity using master branch, ran DB migration and checked if the activity still works and if a student can still update answers.
2. I created a new activity using this branch, published it and tried to run it as a student.

In both cases it worked fine, I could see answers in the report.

> Does the question numbering in the LARA report and portal report make sense? Some interactives save state, some don't, I believe the current numbering reflects this. And there was a lot of tweaks to fix the numbering as related to interactives.

Yes, it works fine. Interactives that don't save state are always skipped in the report. Interactives that do save state and provide report url, will look like questions and have a question number in the report. In LARA, they don't have question header and number. But it works the same right now - an interactive in the interactive box doesn't have header and number, but it increases numbers of the subsequent questions (so it matches report numbering). Interactive that are part of the interactive box will be always at the end of the page report (that's old behavior too). Ones that are mixed with questions will be mixed with questions in the same way in the report.

> When looking at the ITSI stuff, check the way that LabBooks are linked with interactives. And then the labbook buttons were rendered below their interactive regardless of where the labbook item was in the assessment list. I don't remember the details, but that seems like something that this change could break.

It seems to work fine.
I tried to test various caes, but well, I guess this PR requires a lot of QA testing anyway. 

--- 

### A quick explanation how it works

The main change is that `InteractiveItems` are converted into `PageItems`. We still support interactive box. However, now it's implemented using page item `section`. We added sections when we're working on Arg Block. Now it's a bit more useful. Section let you organize / layout embeddables in a custom way. In this case, it's used to render one central interactive. Other things and underlying classes stayed pretty similar, so this changes mostly page rendering.

I removed `InteractiveItem` class, but I did **not** remove interactive_items DB table. This makes DB migration a bit safer and easier to revert (we could still do it even if we deleted interactive_items table completely, but this is even more safe). Once things are tested and work well, we can remove it.

First post lists the most important changes (TODO items are done now).

app/models/interactive_page.rb
```diff
@@ -264,32 +257,17 @@ def export
...
-    self.class.registered_additional_sections.each do |s|
-      additional_section = {name:s[:name],section_embeddables:[]}
-      self.section_embeddables(s[:name]).each do |embed|
-        section_embeddable_hash = helper.wrap_export(embed)
-        additional_section[:section_embeddables] << section_embeddable_hash
+      if embed.respond_to? :linked_interactive
+        embeddable_hash[:linked_interactive] = embed.linked_interactive.present? ? helper.wrap_export(embed.linked_interactive) : nil
       end
-      page_json[:sections] << additional_section
+      page_json[:embeddables] << { embeddable: embeddable_hash, section: embed.page_section }
```
scytacki: I think this format change will break what @dougmartin  is using to generate "prompts" for Collabspace. I think that is OK.  But we probably want to add a version number to the export, so Doug or other systems can check this, so they know how to handle the exported file.

pjanik:
@scytacki, I've added a version number to activity export JSON in the last commit. Is this enough for now?
We could theoretically support old format during LARA import, but I'm wondering if it makes sense. I'm not entirely sure now, but I think that when I was initially working on this, I was asking and it didn't seem necessary. The question is whether we have some sets of old export JSONs that are still useful.

scytacki:
Yes I think the version number is enough for now. That way @dougmartin can use it in his content conversion app.  I don't think we have any important activities that we need to be able to import. If we do then we can make a script that will transform the old format to the new one.

PR #276, opened 2018-02-02
Create "External Scripts" in Lara. Model Feedback

knowuh:
This commit adds two models:
 * ApprovedScripts              - Scripts that admins allow authors to add to page.
 * Embeddable::ExternalScript   - A script in the page added by author.

external_script.js library defines a very rough API for intializing
external scripts.  More work needs to be done defining this API in the
README.

Create "Interactivity Feedback" embeddable assessment item.

[#154444435]

https://www.pivotaltracker.com/story/show/154444435

### DIFF
--- a/app/views/embeddable/external_scripts/_lightweight.html.haml
+++ b/app/views/embeddable/external_scripts/_lightweight.html.haml
@@ -23,8 +23,8 @@
         name: '#{embeddable.name}',
         scriptUrl: '#{embeddable.url}',
         embeddableId: '#{embeddable.id}',
-        config: '#{embeddable.configuration}',
-        div: '#{outputId}'
+        config: '#{embeddable.configuration.gsub(/'/, "").gsub(/\s+/i, " " )}',
+        div: $('#{outputId}')
       }
       ExternalScripts.register(ScriptConstructor, env);
     }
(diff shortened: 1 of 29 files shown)


scytacki:
It would also be possible to use `document.currentScript` to figure out the URL of the current script. Then we wouldn't need to store the identifier in the LARA database. The identifier would just the the URL to the script. But `document.currentScript` isn't supported in IE, and IE is still 2% of LARA users.

app/views/embeddable/external_scripts/_lightweight.html.haml
```diff
@@ -0,0 +1,25 @@
+- runtimeDiv = "output-#{embeddable.id}"
+= content_for :external_scripts do
```
scytacki: Is this going to work with multiple scripts are on the page?
Will rails join together all of the content_for blocks for the external scripts, or does it just show one?
I'm seeing conflicting info in stack overflow, it looks like this might have changed from one rails version to the next.

knowuh: Yes, this keeps appending items into the header using yields.  This is something we used a lot in the early days of the portal.  I just verified it's still in the rails docs:  http://guides.rubyonrails.org/layouts_and_rendering.html#using-the-content-for-method

And I was able to verify that output does contain multiple script tags when added in my local authoring setup.

scytacki:
Looks good to me.

PR #279, opened 2018-02-27
Load 1st available interactive state from linked interactives chain if the directly linked one is empty

pjanik:
It improves a case when students are skipping activity pages. E.g.:
- a student does some work on page 1
- then he skips page 2 where the interactive is linked to page 1
- he opens page 3 instead where the interactive is linked to page 2

Without this fix, he'd see an empty interactive. This PR ensures that he'll see his work from page 1.

### DIFF
--- a/app/models/interactive_run_state.rb
+++ b/app/models/interactive_run_state.rb
@@ -114,7 +114,7 @@ class InteractiveRunState < ActiveRecord::Base
     (interactive.respond_to? :linked_interactive) && !interactive.linked_interactive.nil?
   end
 
-  def find_linked_interactive_state
+  def linked_state
     # Note that this method will return the first available state of the linked interactives *chain*.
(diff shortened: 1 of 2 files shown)


pjanik:
Ops, missing 9 at the end of PT ID. The correct link is:
https://www.pivotaltracker.com/n/projects/736901/stories/155524259

pjanik:
One more commit added:
1. Added `raw_data` check. It fixes a case when user opens a page without opening interactive (and InteractiveRunState is created with null data) and then goes to the next page.
2. Merged two methods together, as prev `linked_state` didn't make much sense.

scytacki:
I just realized that this is kind of quirky behavior in the case where an interactive is merging the state itself like the prediction graph. But I think it is fine to keep it this way. Lets see if that quirkiness causes problems before trying to fix it.

PR #280, opened 2018-03-06
 Extend LARA Interactive API, so interactive can pick the most recent linked state

pjanik:
This PR extends Interactive API, so interactives are able to pick the most recent data from linked states chain. Changes:
- `allLinkedStates` array in `initInteractive` message
- `interactiveStateUpdatedAt` timestamp in `initInteractive` message
- `interactiveStateCreatedAt` timestamp in `initInteractive` message (not necessary, but just to be consistent)
- support of `touch` argument passed in `interactiveState` message (it only updates timestamp)
- fix / better support of `null` argument passed in `interactiveState` (so interactive can reset its state)

@scytacki, most of these changes are necessary no matter if we use `touch` message in document-server or not. I've decided to add `touch` and provide a bunch of timestamps, as it's relatively straightforward and makes implementation on the interactive side way easier. It works for doc-server `autolaunch` page and it might work well for other interactives that would like to support similar behavior in the future. Especially if interactive stores all data in LARA. doc-server `launch` page will require separate implementation (it would anyway), but I still hope we can depreciate that page.

Related document-server PR: https://github.com/concord-consortium/document-store/pull/26

BTW, this PR is based on https://github.com/concord-consortium/lara/pull/279, so it includes its changes too.

### DIFF
--- a/app/models/interactive_page.rb
+++ b/app/models/interactive_page.rb
@@ -195,8 +195,8 @@ class InteractivePage < ActiveRecord::Base
     self.save!(:validate => false) # This is the part we need to override
   end
 
-  def index_in_activity
-    lightweight_activity.visible_pages.index(self) + 1
+  def page_number
+    lightweight_activity.visible_page_ids.index(self.id) + 1
   end
 
   def duplicate(interactives_cache=nil)
--- a/app/models/interactive_page.rb
+++ b/app/models/interactive_page.rb
@@ -195,6 +195,10 @@ class InteractivePage < ActiveRecord::Base
     self.save!(:validate => false) # This is the part we need to override
   end
 
+  def index_in_activity
+    lightweight_activity.visible_pages.index(self) + 1
+  end
+
   def duplicate(interactives_cache=nil)
     interactives_cache = InteractivesCache.new if interactives_cache.nil?
     new_page = InteractivePage.new(self.to_hash)
--- a/app/models/interactive_run_state.rb
+++ b/app/models/interactive_run_state.rb
@@ -123,7 +123,7 @@ class InteractiveRunState < ActiveRecord::Base
   end
 
   # Returns all linked states. 1st element in the array is an interactive directly linked to given one.
-  def all_linked_states(host = nil)
+  def all_linked_states(protocol = nil, host = nil)
     result = []
     current_interactive = interactive
     while (current_interactive.respond_to? :linked_interactive) && !current_interactive.linked_interactive.nil?
@@ -133,18 +133,18 @@ class InteractiveRunState < ActiveRecord::Base
       # It's not necessary to provide some metadata even for empty linked states, although it doesn't cost much
       # and it can be useful for interactive itself to know that there are some linked interactives,
       # but not run by student yet.
-      linked_state_info = { interactive_id: current_interactive.id, data: nil, created_at: nil, updated_at: nil }
+      linked_state_info = { interactive_id: linked_interactive.id, data: nil, created_at: nil, updated_at: nil }
       if state
         linked_state_info[:data] = state.raw_data
         linked_state_info[:created_at] = state.created_at
         linked_state_info[:updated_at] = state.updated_at
         linked_page = state.interactive.interactive_page
         linked_activity = linked_page && linked_page.lightweight_activity
-        linked_state_info[:page_index] = linked_page && linked_page.index_in_activity
+        linked_state_info[:page_number] = linked_page && linked_page.page_number
         linked_state_info[:page_name] = linked_page && linked_page.name
         linked_state_info[:activity_name] = linked_activity && linked_activity.name
         if host
-          linked_state_info[:interactive_state_url] = state.interactive_state_url(host)
+          linked_state_info[:interactive_state_url] = state.interactive_state_url(protocol, host)
         end
       end
       result.push(linked_state_info)
@@ -177,14 +177,14 @@ class InteractiveRunState < ActiveRecord::Base
     arg ? original_json(arg) : answer_json
   end
 
-  def to_runtime_json(host)
+  def to_runtime_json(protocol, host)
     hash = self.as_json
     hash[:has_linked_interactive] = has_linked_interactive
     hash[:linked_state] = linked_state
     hash[:run_remote_endpoint] = run_remote_endpoint
     hash[:all_linked_states] = all_linked_states(host)
-    hash[:interactive_state_url] = interactive_state_url(host)
-    hash[:page_index] = page && page.index_in_activity
+    hash[:interactive_state_url] = interactive_state_url(protocol, host)
+    hash[:page_number] = page && page.page_number
     hash[:page_name] = page && page.name
     hash[:activity_name] = activity && activity.name
     hash.to_json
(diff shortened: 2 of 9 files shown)


app/models/interactive_run_state.rb
```diff
@@ -114,18 +114,39 @@ def has_linked_interactive
...
+    result = []
+    current_interactive = interactive
+    while (current_interactive.respond_to? :linked_interactive) && !current_interactive.linked_interactive.nil?
+      linked_interactive = current_interactive.linked_interactive
+      runs = run.sequence_run ? run.sequence_run.runs : [run]
+      state = InteractiveRunState.where(run_id: runs.map(&:id), interactive_id: linked_interactive.id).first
+      # It's not necessary to provide some metadata even for empty linked states, although it doesn't cost much
+      # and it can be useful for interactive itself to know that there are some linked interactives,
+      # but not run by student yet.
+      linked_state_info = { interactive_id: current_interactive.id, data: nil, created_at: nil, updated_at: nil }
```
scytacki: Should the interactive_id be the `linked_interactive.id` instead?
If that is true, then it might be clearer to create a new method on InteractiveRunState that just returns the state_info hash.

pjanik: Hm, not sure if I understand. Do you mean just the key name (`interactive_id` vs `linked_interactive_id`) or some logic (since you used `.id` here)?
Actually, this ID not necessary and used by anything, but I just added it so there's any form of some unique ID provided in case the interactive needs to work more with those states.

pjanik:
This branch had just failing tests. I've fixed them now, and also added a note to readme how to run Jasmine tests (using the web browser, not jasmine:ci, but it seems good enough for development for now).

I think one of the previous branches had PR branch failing and push working (or the other way). I guess it was actually one of our random issues with tests. Sometimes Jasmine server gets stuck and doesn't start. E.g. even in this relatively old, `master` build:
https://travis-ci.org/concord-consortium/lara/builds/341611216
screen shot 2018-03-06 at 13 18 01

Actually, I remember working on Jasmine issues a long time ago and I couldn't fix this issue. The old version had some other problems, the new one had this one. I guess we just assumed that it fails rarely enough. It might be worth checking again, but it seems to be a separate issue.

pjanik:
Added one more commit which adds interactive state URLs to linked states array metadata. Document server uses those URLs to load previews.

pjanik:
A few more commits added to support new features in the document server. It should be ready for review again.

app/models/interactive_run_state.rb
```diff
@@ -114,22 +122,46 @@ def has_linked_interactive
...
       state = InteractiveRunState.where(run_id: runs.map(&:id), interactive_id: linked_interactive.id).first
-      if state && state.raw_data
-        return state.raw_data
-      else
-        # Try next interactive in chain.
-        current_interactive = linked_interactive
+      # It's not necessary to provide some metadata even for empty linked states, although it doesn't cost much
+      # and it can be useful for interactive itself to know that there are some linked interactives,
+      # but not run by student yet.
+      linked_state_info = { interactive_id: current_interactive.id, data: nil, created_at: nil, updated_at: nil }
```
scytacki: Shouldn't the id here be the `linked_interactive.id`?

app/models/interactive_page.rb
```diff
@@ -195,6 +195,10 @@ def set_list_position(index)
     self.save!(:validate => false) # This is the part we need to override
   end
 
+  def index_in_activity
+    lightweight_activity.visible_pages.index(self) + 1
+  end
+
```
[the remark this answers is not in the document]
scytacki: If you do change this to page_number, then that name should trickle down into the json in the interactive_run_state.

app/models/interactive_run_state.rb
```diff
@@ -184,4 +225,8 @@ def self.default_answer(conditions)
   def add_key_if_nil
     self.key = InteractiveRunState.generate_key if self.key.nil?
   end
+
+  def interactive_state_url(host)
+    host + Rails.application.routes.url_helpers.api_v1_show_interactive_run_state_path(key: self.key)
+  end
```
scytacki: A slightly better approach, that abstracts the construction of the url is to do it this way:
`Rails.application.routes.url_helpers.api_v1_show_interactive_run_state_url(key: self.key, protocol: protocol, host: host_with_port)`
You can see this approach used a few times in the portal for example: https://github.com/concord-consortium/rigse/blob/e1449b04ca831dc023f678e1cdd57ed3f1f5bf57/app/models/portal/clazz.rb#L299

app/models/interactive_run_state.rb
```diff
@@ -145,8 +177,17 @@ def to_json(arg=nil)
...
+    hash = self.as_json
+    hash[:has_linked_interactive] = has_linked_interactive
+    hash[:linked_state] = linked_state
+    hash[:run_remote_endpoint] = run_remote_endpoint
+    hash[:all_linked_states] = all_linked_states(host)
+    hash[:interactive_state_url] = interactive_state_url(host)
+    hash[:page_index] = page && page.index_in_activity
+    hash[:page_name] = page && page.name
+    hash[:activity_name] = activity && activity.name
+    hash.to_json
```
scytacki: Since this is roughly the same as what is returned for each linked_state_info. It would be better to unify these two somewhat.  One difference is data vs raw_data. Another is that this top level interactive state includes all of the properties while the linked state only selects a few. And finally this one is recursive while the other one isn't. I probably missed some differences. But here is a suggestion based on those differences:

- use `raw_data` for the linked interactive state too 
- it would be safer to explicitly define which properties of the interactive run state are sent down, so limiting this to just created_at, updated_at, raw_data, and interactive_id seems fine to me. However it should be backward compatible, so perhaps the current doc-store interactive expects some of the other properties (run_id, learner_url, or key).
- passing a `include_linked_states` param to this method would solve the recursive issue

app/models/interactive_run_state.rb
pjanik: It should, thanks! Fixed.

app/models/interactive_page.rb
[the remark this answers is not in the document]
pjanik: Good point, done.

app/models/interactive_run_state.rb
pjanik: I've changed it. I had used the previous approach as that's what we use in other places in LARA (and requires one less argument to be passed around).

app/models/interactive_run_state.rb
pjanik: Done. You should see that in the last commit. It was a bit tricky, as `to_runtime_json` is used both by API and `initInteractive` post message. I've tried to include reasonable params that were enforced either by `initInteractive` format, API specs or common sense.
I've also refactored `initInteractive` message to make use of the more consisten approach.

pjanik:
I've addressed all the comments (more detailed answers are in outdated, inline threads). Thanks!

scytacki:
Looks great!

scytacki:
Here is a story so I don't forget this when releasing LARA: https://www.pivotaltracker.com/story/show/156024882

PR #281, opened 2018-03-06
capitalization error was causing inconsistent authoredState

scytacki:
All of the interactive that I could find were already handling both strings and object
forms of the authoredState. So this fix should not break anything.

During runtime the authoredState was being sent to the interactive as a object
even with this bug. This is because the jquery data method is returning an
object not a string at least in Chrome. I'm not sure if this is a feature of jquery or of
Chrome. The server is sending down `data-authored-state` attribute, but in the
DOM this attribute is not present. Instead there is a dataset property on the element.

[#155754828]

This might break some tests. Lets see...


pjanik:
Looks good. 👍 I guess jQuery tries to parse that automatically when it detects 'application/json' type (or when we set it explicitly).

PR #282, opened 2018-03-21
Responsive layout, interactive API: aspectRatio, new interactive pinning code

pjanik:
Changes:

1. Extension to LARA Interactive API: an interactive can set an `aspectRatio` property in `extendedSupport` message. LARA will respect that and set the aspect ratio to the desired value. It works for all the layouts. If interactive doesn't send anything, we still use width/height ratio provided by authors.

2. Responsive layout - a new type of layout where interactive tries to take as much horizontal space as possible and the full page width is used.

3. Completely new interactive pinning code. The old one wasn't working great (interactive was jumping a bit at the beginning + it was problematic to make it work with a bit more complicated CSS rules used by responsive layout). It was using some library (jquery.waypoints, removed now), but still, lots of custom code was necessary. I think that completely custom solution is better, as we have full control over things that happen there. And I think that the new approach is a bit more correct/generic. It puts a placeholder in place of the pinned element, so page layout should not be affected. The old code was trying to work around that using custom CSS for various LARA layouts. The main reason for this change was to make the responsive layout work with interactive pinning (it was one of the biggest problems). Another advantage of the new implementation is that now multiple interactives can be pinned if their total height is smaller than the height of the window (so we don't need to scroll to see everything). It works automatically. This part would require solid Q&A. There are many, many edge cases when we use the responsive layout, e.g. you can scroll down so interactive is pinned and start resizing the browser window. I've tried to support all that.

4. Cleanup / reimplementation of interactive-sizing. I've removed automated tests, even though I wrote new ones. They were failing while used with PhantomJS and I'd say they didn't provide much value. This is all very visual and might also be browser specific (I hope it's not).

5. Tiny cleanup of full-width layout styles (as I started from there while working on responsive one).

It's only one commit, as I was working locally on that and those changes are overlapping. I could have tried to divide it into separate commits, but decided it's not worth it.

[#155515297]
[#155514666]


scytacki:
I was trying to deprecate extendedSupport. It is documented here: http://concord-consortium.github.io/lara-interactive-api/docs/implementing-interactive/#getextendedsupport
Can the aspectRatio be passed in supportedFeatures instead?

scytacki:
I can't track down why I wanted to deprecate it. I think because it was redundant with the 'supportedFeatures' message.  But I don't remember why we added 'supportedFeatures' instead of sticking with 'extendedSupport'.   Perhaps it is because the interactive can send 'supportedFeatures' whenever it wants, but the interactive is supposed to wait to send 'extendedSupport' until it receives the 'getExtendedSupport' message.

At one point I think I had the idea that the interactive would need to send 'supportedFeatures' first before LARA sent initInteractive . This way the interactive could tell LARA which version of the iframe API it used. Beyond this, I was hoping to build this kind of thing into the initial connection message sent by iframePhone. 

Anyhow, I'm just providing context.  If we need to have both 'supportedFeatures' and 'extendedSupport' for some reason, that is OK with me now.

app/assets/stylesheets/cc-runtime-base.scss
```diff
@@ -174,11 +174,6 @@ div.activity-mod>div.site-width {
...
 }
 
-// *********************************************
-//  7. Sticky Interactives
-// *********************************************
-@import "partials/runtime-sticky";
-
 // *********************************************
 //  8. Interactive + Questions
 // *********************************************
```
scytacki: Probably ought to update the number of the sections.

app/views/interactive_pages/_interactive.html.haml
```diff
@@ -1,5 +1,5 @@
 -# Do not pin multiple interactives, as it means that the user might not be able to see the bottom ones
 -# if his screen is shorter that the height of all the interactives
-.interactive-mod{class: (page.visible_interactives.count > 1 || layout == 'l-full-width') ? nil : 'pinned'}
+.interactive-mod{class: layout == 'l-full-width' ? nil : 'pinned'}
```
scytacki: Probably should update the comment above this lines so it matches the code.

app/views/interactive_pages/_show.html.haml
```diff
@@ -11,8 +11,8 @@
     - if page.visible_interactives.length > 0 && page.show_interactive
       = render partial: 'interactive_pages/interactive', locals: {page: page, layout: layout}
 
-  .questions-mod.ui-block-2{ :class => page.embeddable_display_mode == 'carousel' ? 'jcarousel' : '' }
-    - if page.show_info_assessment
+  - if page.show_info_assessment
+    .questions-mod.ui-block-2{ :class => page.embeddable_display_mode == 'carousel' ? 'jcarousel' : '' }
```
scytacki: What if the layout is 60-40, but the user choses to not show the info_assessment block.  Currently LARA will still keep the interactive at 60% width. I don't think this matters if the result is a 60% interactive or a 100% interactive, but I just wanted to make sure this change doesn't break the page in that case.

app/assets/stylesheets/cc-runtime-base.scss
pjanik: Done.

app/views/interactive_pages/_interactive.html.haml
pjanik: Thanks, fixed.

app/views/interactive_pages/_show.html.haml
pjanik: Yes, I was testing it. And double checked now. Interactive will still take 60 or 70% space only. I think it makes sense, if the author wants full width then the full-width layout is there (or responsive).

pjanik:
Thanks. I fully agree on `supportedFeatures`. I used `extendedSupport` as document-store was using it and perhaps I forgot about `supportedFeatures` existence. I've made this change. I've added support of `reset` to `supportedFeatures`, so I could fully migrate to doc-store autolaunch page to `supportedFeatures`.

Also, I've updated interactive-related React components so they support aspect ratio too. So, interactive that provides aspect ratio will be rendered correctly in regular authoring and ITSI authoring (tested that). Perhaps not necessary at the moment, but might be useful in the future. It could be a simple and useful change in Lab, since most of the interactives actually define optimal aspect ratio in the Interactive JSON. They could easily send that to LARA.

pjanik:
You can see a demo / screencast here: 
https://www.pivotaltracker.com/story/show/155515297/comments/187431831

PR #284, opened 2018-03-21
this fixes an issue with the app not being able to access the lara test database

scytacki:
This wasn't necessary for @pjanik, but when I tried to start up lara I was seeing errors.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -23,9 +23,6 @@ services:
   db:
     image: mysql:5.6
     environment:
-      MYSQL_DATABASE: lara
-      MYSQL_PASSWORD: master
-      MYSQL_USER: master
       MYSQL_ROOT_PASSWORD: xyzzy
     command: mysqld --character-set-server=utf8 --collation-server=utf8_general_ci
     volumes:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,10 +6,11 @@ services:
       dockerfile: Dockerfile-dev
     command: docker/dev/run.sh
     environment:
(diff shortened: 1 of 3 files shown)


pjanik:
Honestly, I don't really understand that. A few lines below, in DB container, we specify master/master again. `docker/dev/run_spec.sh` exports root/xyzzy explicitly. There are a few other places where user/password is defined too. 😉 

I've tried that change locally, rebuild my app container, and everything still works. So, I'm fine with merging, but I can't say I'm following what's going on. 😉

scytacki:
I didn't see that `docker/dev/run_spec.sh` configuring this itself.  I'm now pretture sure the error I was seeing was actually harmless.  It believe it was due to this line: 
https://github.com/concord-consortium/lara/blob/master/docker/dev/run.sh#L30 
I removed the duplicate references to usernames an passwords that aren't needed now.

Also I removed the comments that explain why using the root user is a required. 
Basically the DB docker container has some automatic configuration code. It looked at the DB
MYSQL_DATABASE, MYSQL_PASSWORD, and MYSQL_USER environment variables and creates a database if it doesn't exist. and then creates a user that only has access to that database.

By using the root user and password in docker-compose it simplifies things.

@pjanik can you check one more time that these new changes aren't breaking anything for you?

scytacki:
I'm going to go a head and merge this in because I'm pretty sure it is safe.

pjanik:
@scytacki, sorry, somehow I missed that on Friday. It works fine and simpler better now, thanks.

PR #287, opened 2018-03-23
Added DB_NAME to docker-compose.yml

dougmartin:
Without the DB_NAME defined the app can't connect to the database.


scytacki:
We just removed that setting, because it causes the default database.yml file to use the same database for development and test. And that can cause the test code to erase the development database. Without that set, then the default is to use `lara` for development and `lara_test` for test.  If you can't connect without it, then I'd suspect you have an older or customized database.yml that is different than database.sample.yml.

PR #288, opened 2018-03-28
Fix "Undo all work" button, especially in combination with "Click to play"

pjanik:
It might be hard to guess what has changed reading the code, so a bit longish description: 

The main goal was to fix issues with "Undo all my work" button in combination with "Click to play". The button was covered by "Click to play" overlay and didn't really work, as the previous implementation was expecting iframe to be present in DOM for this button to work ("Click to play" creates iframe DOM element only if the user actually clicks the overlay image).

To fix this, I just always show this button if the given interactive supports state saving. We don't really need iframe to be loaded, as we can easily delete interactive data anyway (it just resets a record in DB).

Interactive can tell LARA that it doesn't support "Reset" button using LARA Iframe API. This case is a bit problematic together with "Click to play", as LARA won't receive this message until the user starts interactive. But I think the best solution is just to avoid using such interactives together with "Click to play". Also, I think there are not too many interactives that don't want this button (I can only think of button-mode doc-server CODAP launcher which will be probably depreciated soon). And I can't really see why it'd be useful in the future. We can always reset a record in LARA DB to restore things to a clean state, why not.

Another change is that I removed some poor logic that tried to show or hide this button depending on whether interactive state is already saved or not. It makes it consistent with "Click to play" case which doesn't have this info yet. It means that button will be always there and you can reset "empty" state. I don't think it's a big deal. And might actually help if interactive gets broken or something like this. Also, the previous implementation was imperfect anyway - button was shown only after the page was reloaded, not after the state was actually saved. So, it wasn't clear that work can be reset. Now users have access to it straight away.

[#152171695]
[#151762176]

### DIFF
--- a/app/assets/javascripts/interactive-manager.coffee
+++ b/app/assets/javascripts/interactive-manager.coffee
@@ -19,13 +19,12 @@ class InteractiveManager
     $data = $interactive.find('.interactive_data_div')
     # Delete data button is rendered outside interactive container, just next to it.
     $delete_button = $interactive.siblings('.delete_interactive_data').first()
-    # IFrameSaver does a lot more than only saving.
-    # It also initializes interactive and provides LARA Interactive API.
-    # Create IframeSaver even if there's no iframe yet (it might happen when "Click to play" is being used).
-    # It ensures that delete data button still works.
-    new IFrameSaver($iframe, $data, $delete_button)
 
     if $iframe.length != 0
+      # IFrameSaver does a lot more than only saving.
+      # It also initializes interactive and provides LARA Interactive API.
+      new IFrameSaver($iframe, $data, $delete_button)
+
       if globalIframeSaver
         globalIframeSaver.addNewInteractive $iframe
       LoggerUtils.logInteractiveEvents($iframe)
--- a/app/assets/javascripts/interactive-manager.coffee
+++ b/app/assets/javascripts/interactive-manager.coffee
@@ -18,7 +18,7 @@ class InteractiveManager
     $iframe = $interactive.find('iframe[src]')
     $data = $interactive.find('.interactive_data_div')
     # Delete data button is rendered outside interactive container, just next to it.
-    $delete_button = $interactive.siblings('.delete_interactive_data').first()
+    $delete_button = $("#delete-int-data-#{$data.data('interactive-id')}")
 
     if $iframe.length != 0
       # IFrameSaver does a lot more than only saving.
--- a/app/assets/javascripts/interactive-manager.coffee
+++ b/app/assets/javascripts/interactive-manager.coffee
@@ -17,17 +17,17 @@ class InteractiveManager
   _register: ($interactive) ->
     $iframe = $interactive.find('iframe[src]')
     $data = $interactive.find('.interactive_data_div')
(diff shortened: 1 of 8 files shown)


scytacki:
My concern with this change is the ITSI activities. I think there are a few that are using the table and more that hopefully will be soon.  Carolyn likes to keep the UI simple and consistent. All of the interactives in ITSI are click to play. Can you setup or find an example in authoring.staging so we can look at what this button looks like there?  The solution might just be to check this UI in ITSI and make sure it looks fine.

How about this approach:
We remove the 'reset' option from the iframe api all together, and just add another option in authoring. The default is to show the reset button, but an author can choose to hide it.

The use case I'm thinking of is when the interactive is super simple like a table.  In this case having an undo all of my work button is confusing. The table will seem like it is part of the activity just like an open response question.  And the student can delete their work just by clearing the cells.  Having this option is authoring does put more of a burden on the activity authors, but it doesn't seem too bad of a burden.

pjanik:
@scytacki, I've updated this PR.

@ddamelin wanted to hide "Undo all my work" button until "Click to play" interactive in started. Instead of implementing a special case for "Click to play" interactives, I decided to implement it in a generic way that will have pretty much the same effect for "Click to play". 

The "Undo all my work" button is now hidden on the page load. It will be shown in two cases: 

1. When LARA detects that there's already saved state (e.g. user was working with this interactive earlier). It happens asynchronously, right after the page is loaded, so in practice, the button will be shown almost immediately.
2. When a user starts playing with the interactive and LARA saves new state.

When "Click to play" is enabled, 1 doesn't happen unless you actually open interactive, so button won't be visible. Once you start the interactive, the button will be shown either immediately (if there's some previous state) or after user makes the first change. 

Note that this is similar to the original implementation, but it should work better (I tried to simplify it and the original implementation required a page reload to show the delete button).

Also, I've added a new checkbox to authoring page. Author can decide whether this button should be visible or not. Screenshots:
screen shot 2018-03-29 at 16 57 31
screen shot 2018-03-29 at 16 57 54

This option is enabled by default, to be backward compatible.

I removed "reset: true/false" support from LARA Interactive API "supportedFeatures" message handler and completely removed "getExtendedSupport" and "extendedSupport" messages + handlers.

app/assets/javascripts/interactive-manager.coffee
```diff
@@ -17,8 +17,8 @@ class InteractiveManager
   _register: ($interactive) ->
     $iframe = $interactive.find('iframe[src]')
     $data = $interactive.find('.interactive_data_div')
-    $delete_button = $interactive.find('.delete_interactive_data')
-    $click_to_play = $interactive.find('.click_to_play.shown')
+    # Delete data button is rendered outside interactive container, just next to it.
+    $delete_button = $interactive.siblings('.delete_interactive_data').first()
```
scytacki: Does this work correctly with multiple interactives on a page?

pjanik:
Thanks, good catch. The last commit fixes multi interactives case. I was considering this case, but I thought `$.siblings` worked in a different way. I could easily fix it using `$.nextAll`, but switched to using IDs to make it less dependant on DOM structure.

PR #289, opened 2018-03-28
Add logging when a run is sent to the portal

scytacki:
This log info should show up in the worker log when running on production.

While setting up the ngss assessment portal in ECS. I had an issue where learner data was not being sent from LARA to the portal. It turned out to be an issue with an old sequences run that had not been updated to https. I never really reconstructed what exactly happened, and it took a lot of time to track down. Hopefully with this new logging, it will be easier to get a handle on issues like this. 

I'd note that the portal does log its side of this in its request log. However in this case there wasn't anything useful showing up in the portal logs. And the portal log did not contain info about the LARA run id and key that was sending the data.

The formatting change is so the log is just a single line instead of split into multiple lines with bunches of spaces in between.

[#156348385]


scytacki:
Looks like this broke some tests. Please hold...

PR #291, opened 2018-03-30
Fix arg block rendering

pjanik:
This PR fixes arg block rendering. I introduced the bug while I was working on the responsive layout. Test page:
https://authoring.staging.concord.org/activities/628/pages/2807/

The important change is in the HAML file. Actually, it restores it to what it was before. Scott, I remember you're a bit suspicious about it back then.

I've also tested:
 - responsive layout with questions
 - responsive layout without any questions
 - responsive layout with arg block (unlikely this combination is gonna be used, but just so it's not broken)

[#155515297]


scytacki:
👍

PR #292, opened 2018-04-03
Replace an old jeditable rich text editor with TinyMCE

pjanik:
The general idea was to remove an ancient jeditable wysiwyg editor and replace with TinyMCE that was already used by ITSI. That involved:

1. Upgrade of TinyMCE version (slightly different look).
2. New React components used in authoring mode (text editor).
3. Refactoring of ITSI components and mixins, so the code can be shared between ITSI and regular authoring components.
4. Cleanup of various TinyMCE configs. E.g. ITSI, Tracked Questions, and so on. They seemed unnecessary. I've tried to reduce it to only two: regular and minimal. Minimal is used in places where the width is limited (questions edit dialogs). The basic one is pretty much an old ITSI config (and ofc still used by ITSI).
5. Jeditable lib was also to edit interactive page title in place. I've implemented simple React component that does the same, so Jeditable could be removed.
6. Jeditable lib removed.
7. A better approach to sidebar title editing. Now, when you click "edit", it will edit both title and content.
8. I've made sure that custom HTML doesn't break page and sidebar titles (we've seen that during demo). It will be escaped and html tags displayed, instead of being interpreted as before.

Screenshots:
1. Introduction:
    screen shot 2018-04-03 at 12 21 37
2. Sidebar editing:
    screen shot 2018-04-03 at 12 21 52
3. Page title:
    screen shot 2018-04-03 at 12 22 11
4. Question edit dialog (minimal TinyMCE config):
    screen shot 2018-04-03 at 15 06 49
5. Activity edit form (I made it a bit wider):
    screen shot 2018-04-03 at 15 06 24

There are more places where TinyMCE is used, but these above are perhaps the most important.

### DIFF
--- a/app/views/embeddable/open_responses/_form.html.haml
+++ b/app/views/embeddable/open_responses/_form.html.haml
@@ -15,9 +15,10 @@
       .ul
         = render :partial => 'shared/edit_prediction', :locals => { :f => f }
 
-    :javascript
-      $(function() {
-          $('.wysiwyg').wysiwyg({"controls": { "html": { "visible": true }, "h1": { "visible": false }, "h2": { "visible": false }, "h3": { "visible": false } } });
-      });
   = f.submit "Cancel", :class => 'close'
   = f.submit "Save", :class => 'embeddable-save'
+
+  :javascript
+    $(function() {
+      initTinyMCE('.wysiwyg', window.TinyMCEConfigMinimal);
+    });
--- a/app/views/embeddable/open_responses/_form.html.haml
+++ b/app/views/embeddable/open_responses/_form.html.haml
@@ -3,13 +3,13 @@
   = f.label "Heading"
   = f.text_field :name
   = field_set_tag 'Prompt' do
-    = f.text_area :prompt, :rows => 5, :class => "wysiwyg"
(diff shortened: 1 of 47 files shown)


app/views/embeddable/open_responses/_form.html.haml
```diff
@@ -15,9 +15,10 @@
...
-    :javascript
-      $(function() {
-          $('.wysiwyg').wysiwyg({"controls": { "html": { "visible": true }, "h1": { "visible": false }, "h2": { "visible": false }, "h3": { "visible": false } } });
-      });
   = f.submit "Cancel", :class => 'close'
   = f.submit "Save", :class => 'embeddable-save'
+
+  :javascript
+    $(function() {
+      initTinyMCE('.wysiwyg', window.TinyMCEConfigMinimal);
```
scytacki: It looks like this would apply the minimal config to all `.wysiwyg` elements on the page. What happens when there are some elements that should be minimal and some that should be basic?  Or are there never both in the DOM at the same time?

pjanik:
@scytacki, thanks, I've addressed two issues. I hoped ITSI would be okay with code editor too. And I've used different class/selector for minimal config (`wysiwyg-minimal`). I started with per-page initialization and added the general one, which uses `.wysiwyg` selector, later.

PR #293, opened 2018-04-06
TinyMCE improvements and fixes

pjanik:
A few semi-related changes:

1. TinyMCE assets compilation should be fixed now (dockerhub build problem).
2. Code editor added to the minimal TinyMCE config.
3. New options in TinyMCE image dialog - float left/right. It's done using custom classes:
     screen shot 2018-04-05 at 18 25 10
4. Page introduction layout vs floating images bug fixed. Layout was broken when there were floating images in the page introduction and no elements after it.


scytacki:
👍

PR #294, opened 2018-04-06
Precompile tinymce-content.css, add margins to floating images

pjanik:
... and update font-size so the editor matches better the final rendering in the runtime mode.


scytacki:
👍

PR #299, opened 2018-05-25
the updated_at time of Multiple Choice Answers was not being set correctly

scytacki:
For some reason the update callback is triggered, but the update_at time is not set.

[#157824597]


scytacki:
@knowuh this looks ready to review as well. All tests are passing, and the change is very minor.

PR #303, opened 2018-06-10
Allowed portal firebase token api endpoint [#157488572]

dougmartin:


dougmartin:
@scytacki In case you didn't see this push I added a security check and a rspec controller test.

PR #304, opened 2018-06-19
add sync container, update docker-compose file version

scytacki:
this also updates the bundler container

This is basically the same as the changes made to rigse

### DIFF
--- a/Dockerfile-dev
+++ b/Dockerfile-dev
@@ -3,7 +3,7 @@ FROM concordconsortium/docker-rails-base-ruby-2.2.6:1.1
 #
 # Install wait-for-it to support docker-volume-sync
 WORKDIR /usr/local/bin
-RUN wget https://raw.githubusercontent.com/vishnubob/wait-for-it/master/wait-for-it.sh && \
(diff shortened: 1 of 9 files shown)


Dockerfile-dev
```diff
@@ -1,4 +1,10 @@
-FROM concordconsortium/docker-rails-base-ruby-2.2.6
+FROM concordconsortium/docker-rails-base-ruby-2.2.6:1.1
+
+#
+# Install wait-for-it to support docker-volume-sync
+WORKDIR /usr/local/bin
+RUN wget https://raw.githubusercontent.com/vishnubob/wait-for-it/master/wait-for-it.sh && \
```
dougmartin: I think it might be safer to use the blob id instead of the master tag, ie https://raw.githubusercontent.com/vishnubob/wait-for-it/db049716e42767d39961e95dd9696103dca813f1/wait-for-it.sh

scytacki: Good point. I changed that, and also updated the portal: https://github.com/concord-consortium/rigse/pull/514

PR #305, opened 2018-06-19
Fix the page item positions

scytacki:
An earlier migration caused there to be duplicate positions

[#158473908]


scytacki:
turns out using update_attributes was not good. That triggers both callbacks and validations.  Triggering callbacks is dangerous because acts_as_list will reorder the list during the callback which can result in the list changing out from underneath while we iterate it.  The solution is to use update_column instead which bypasses callbacks and validations.

PR #306, opened 2018-06-22
Add is_featured field to reportable questions + update specs

pjanik:
[#158190868]

This PR adds is_featured to reportable items, updates UI and specs.


pjanik:
> It looks like the only portal_hash method that is currently tested is the Xhtml object. Maybe you could make a test that went through a whole set of objects and checked if they returned 'is_featured' correctly. 

If I look at the correct test, it seems it only tests "to_hash". Xhtml is never sent to Portal I guess, so there's no portal_hash.

Anyway, I've added tests of #portal_hash to embeddables. I think it makes more sense that one test that is focused on is_featured.

PR #309, opened 2018-07-03
Cleanup _list_embeddables.html.haml code

pjanik:
Reorganize rendering of embeddables. Previous change in https://github.com/concord-consortium/lara/pull/307 was meant to remove unnecessary `.embeddable-interactive` CSS class and avoid duplication of CSS classes geneartion. However, I didn't notice that I was creating unnecessary divs in some cases. Changes in this PR:

1. Don't create empty `div.question` tags for Labbooks in ITSI layout
2. Include `Embeddable` in interactives, mostly to provide `show_in_runtime?` implementation, but it also seems to make sense in general. Interactives are now real embeddables.

### DIFF
--- a/app/models/embeddable.rb
+++ b/app/models/embeddable.rb
@@ -74,10 +74,4 @@ module Embeddable
   def index_in_activity(activity)
     activity.reportable_items.index(self) + 1
   end
-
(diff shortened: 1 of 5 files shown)


scytacki:
👍

PR #310, opened 2018-07-04
Let authors pick interactive for the image question

pjanik:
[#158702375]

I pretty much extracted necessary functionality from Labbook, added to a separate module and reused it in ImageQuestion. I did a similar thing with tests (added `shared_examples`).

I've also added a warning if ImageQuestion is set to shutterbug, but without referencing any interactive:
screen shot 2018-07-04 at 15 40 52

Migration ensures that new DB attributes will be set to the first interactive on a given page. I've used update_column to avoid any callbacks and so on.

### DIFF
--- a/app/models/attached_to_interactive.rb
+++ b/app/models/attached_to_interactive.rb
@@ -1,0 +1,53 @@
+module AttachedToInteractive
+  def self.included(clazz)
+    clazz.class_eval do
+      # That requires model that includes this module to have interactive_id (integer) and interactive_type (string) attributes.
+      belongs_to :interactive, :polymorphic => true
+
+      attr_accessible :interactive_select_value
+      attr_writer :interactive_select_value
+
+      before_validation :parse_interactive_select_value
+    end
+  end
+
+  NO_INTERACTIVE_LABEL =   I18n.t('NO_INTERACTIVE')
+  NO_INTERACTIVE_VALUE = "no-interactive"
+  NO_INTERACTIVE_SELECT = [NO_INTERACTIVE_LABEL, NO_INTERACTIVE_VALUE]
+
+  def possible_interactives
+    page ? page.interactives : []
+  end
+
+  def interactives_for_select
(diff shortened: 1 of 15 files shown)


app/models/attached_to_interactive.rb
```diff
@@ -0,0 +1,53 @@
+module AttachedToInteractive
+  def self.included(clazz)
+    clazz.class_eval do
+      # That requires model that includes this module to have interactive_id (integer) and interactive_type (string) attributes.
+      belongs_to :interactive, :polymorphic => true
+
+      attr_accessible :interactive_select_value
+      attr_writer :interactive_select_value
```
scytacki: It seems a little strange that this module is adding the `belongs_to` for the interactive. But it isn't adding the `attr_accessible` for interactive_type and interactive_id

db/migrate/20180703131647_add_interactive_to_image_question.rb
```diff
@@ -0,0 +1,23 @@
+class AddInteractiveToImageQuestion < ActiveRecord::Migration
+  def up
+    add_column :embeddable_image_questions, :interactive_id, :integer
+    add_column :embeddable_image_questions, :interactive_type, :string
+
+    Embeddable::ImageQuestion.find_each(batch_size: 10) do |iq|
```
scytacki: To be safe this file should define `Embeddable::ImageQuestion` and all of the other models it is using. That way if we remove this class in the future the migration can still run.   An example is here: https://github.com/concord-consortium/lara/blob/a58e866b9a954b72002fd0d44b34c1f5447bcafe/db/migrate/20180619214822_fixup_page_item_positions.rb#L1-L9 Unfortunately in this case that requires several models. But I think it is worth the effort. It also makes it more it explicit which relationships of the models are being used by the migration.

scytacki:
👍

scytacki:
Oh, but it still has merge conflicts.

PR #313, opened 2018-07-09
Description properties update

pjanik:
https://www.pivotaltracker.com/story/show/158870433
https://www.pivotaltracker.com/story/show/158870340
https://www.pivotaltracker.com/story/show/158870718

1. "Activity Description" -> "Text for index page" (only in UI)
2. "Sequence Description" -> "Text for index page" (only in UI)
3. The sequence abstract should be removed, but we can't do it as long as the Portal expects it. We still need to show it in the authoring UI, as that is the only (correct) way to manage abstract value in old Portals. I've added a note for authors that this property is depreciated.

We still need to send all the properties to Portal. I've added some comments to make it clear that they are there only for backward compatibility. Once all the Portals are upgraded to version 1.31 or bigger, we can clean up these things. Here's a PT story that describes necessary cleanup:
https://www.pivotaltracker.com/story/show/158910688

Related Portal PR: https://github.com/concord-consortium/rigse/pull/523

### DIFF
--- a/app/views/sequences/_form.html.haml
+++ b/app/views/sequences/_form.html.haml
@@ -52,7 +52,7 @@
       %br
       = f.text_area :description, { :class => 'wysiwyg', :cols => 60, :rows => 10 }
     .field
-      = f.label :abstract, "Abstract [DEPRECIATED] Works only with old Portals"
+      = f.label :abstract, "Abstract [DEPRECATED] Works only with old Portals"
       %br
       = f.text_area :abstract, { :class => 'wysiwyg', :cols => 60, :rows => 10 }
     .actions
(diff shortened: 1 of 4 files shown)


app/views/sequences/_form.html.haml
```diff
@@ -48,11 +48,11 @@
       = f.check_box :is_official
 
     .field
-      = f.label :description
+      = f.label :description, "Text for index page"
       %br
       = f.text_area :description, { :class => 'wysiwyg', :cols => 60, :rows => 10 }
     .field
-      = f.label :abstract
+      = f.label :abstract, "Abstract [DEPRECIATED] Works only with old Portals"
```
kswenson: "DEPRECATED" 😉

pjanik: Ops 🙄 Thanks! Fixed.

scytacki:
👍

PR #314, opened 2018-07-13
Precipitating change theme fix

emcelroy:
I incorrectly added the project logo to the original theme via style rules. This removes those rules.

The theme on LARA production will need to be updated to point to the project logo once this fix is deployed there. I can take care of that part.


scytacki:
Related PT story: https://www.pivotaltracker.com/story/show/159088454

PR #318, opened 2018-08-08
Reimplement copying

pjanik:
@scytacki, I'm opening this PR so you can take a look at the code and it's easier for me to address your comments tomorrow. I'll fix and add tests tomorrow.

Related to this PR: https://github.com/concord-consortium/rigse/pull/538

The old duplication method still exists so old Portals are supported. This one seems to be simpler to understand. The same code is reused for Activities and Sequences. The only thing that I don't like is this manual call to `portal_publications.create(...` and generally that we "publish" resource in a bit different way than we used to before. Also, `portal_publications` structure is designed to include response details, sucess or failure status, and so on. We can't provide most of these information here. I wouldn't say it's a big problem, though.

I've tested a use case when user never visted LARA before and it seems to work. It's created in remote_dulicate and when he visits LARA, the correct account seems to be recognized. However that might be a good use case for Evangeline to double check on staging too.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -351,17 +351,17 @@ class ApplicationController < ActionController::Base
       user = "Anonymous"
     end
 
-    className = resource.class.name
+    className = resource && resource.class.name
     if className == "LightweightActivity" || className == "Sequence"
       message = "#{user}, you don't have permission to #{action} the #{className} with id : #{resource.id}"
     else
       message = "#{user}, you don't have permission to access this page."
     end
 
-    data = {
+    {
       :user => user,
       :className => className,
-      :id => resource.respond_to?(:id) ? resource.id : 'n/a',
+      :id => resource && resource.respond_to?(:id) ? resource.id : 'n/a',
       :message => message
     }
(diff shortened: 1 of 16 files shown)


pjanik:
@scytacki, thanks. I've thought about some refactoring there too. I have implemented pretty much your idea but tried to simplify a few things (no arguments passed to the block, instead I'm using variables from outer context).

scytacki:
Looks good. Duplicate 'serialize_for_portal' seems like a good trade off to avoid additional complexity.

The only thing left are some tests. :)

scytacki:
Oh I guess the tests are still broken. It looks like they need to updated with the `source_type`

pjanik:
Yes, I'm going to add some tests hopefully and fix broken ones, but I started from Portal.

pjanik:
@scytacki, I've fixed tests and added a new one. Also, I slightly changed how params are handled. The previous pattern (manual JSON parsing) was copied from some old code and it didn't make sense + caused a lot of problems when I wanted to write RSpec controller tests.

scytacki:
👍

PR #319, opened 2018-08-10
Fixed routing error for url helper for newly created runs [#159208846]

dougmartin:
Adds a check for the `run.id` being nil as an additional fix for this PR: https://github.com/concord-consortium/lara/pull/315

The `print_blank` controller method used `Run.new()` to create the @run variable which meant the `id` was nil - this caused the `api_v1_get_firebase_jwt_url(run.id)` helper to break.


dougmartin:
@scytacki The PR build passes but the push build keeps timing out.

PR #320, opened 2018-08-29
Initial documentation work

knowuh:
@pjanik 

This ongoing work is just to clean up & document what is happening with external-scripts. I also started to produce some tests for the existing API -- but I didn't get to build them out yet.

### DIFF
--- a/app/assets/javascripts/external-scripts.js
+++ b/app/assets/javascripts/external-scripts.js
@@ -1,22 +1,34 @@
 ExternalScripts = {
-  _constructors: {},
+  _classes: {},
   _scripts:[],
   _script_labels:[]
 };
 
-ExternalScripts.notRegistered = function(label) {
+ExternalScripts._notRegistered = function(label) {
   console.error("No external script registered for " + label);
 };
 
-ExternalScripts.externalScriptError = function(e, other) {
+ExternalScripts._externalScriptError = function(e, other) {
   console.group("External Script Error");
   console.error(e);
   console.dir(other);
   console.groupEnd();
 };
 
+/**************************************************************
+@function init
+This method is called to initialize the external scripts
+Called from the LARA partial `_lightweight.html.haml`
+@arg {string} label - the the script identifier.
+@arg {any} ctx - context object {} for the script, includes
+      scriptLabel -- same as label above
+      scriptUrl -- where the script was loaded from.
+      embeddableId -- the embeddableId of the script.
+      config -- author configuration.
+      div -- the invisible div on the page for this script to render.
+**************************************************************/
 ExternalScripts.init = function(_label, ctx) {
-  var constructor = this._constructors[_label];
+  var constructor = this._classes[_label];
   var config = {};
   var script = null;
   if (typeof constructor == 'function') {
@@ -30,43 +42,28 @@ ExternalScripts.init = function(_label, ctx) {
       this._script_labels.push(_label);
     }
     catch(e) {
-      this.externalScriptError(e, ctx);
+      this._externalScriptError(e, ctx);
     }
   }
   else {
-    this.notRegistered(_label);
+    this._notRegistered(_label);
   }
 }
 
-ExternalScripts.config = function(_label, ctx) {
-  var constructor = this._constructors[_label];
-  var script = null;
-  if (typeof constructor == 'function') {
-    if (typeof constructor.config == 'function'){
-      try {
-        script = constructor.config(ctx);
-      }
-      catch(e) {
-        this.externalScriptError(e, ctx);
-      }
-    }
-    else {
-      console.log("No configuration for " + _label);
-    }
-  }
-  else {
-    this.notRegistered(_label);
-  }
-}
-
-
-ExternalScripts.register = function(_label, _constructor) {
-  if(typeof _constructor == 'function') {
-    if(this._constructors[_label]) {
+/**************************************************************
+ @function register
+ Register a new external script as `label` with `_class `
+ @arg label - the identifier of the sctipt
+ @arg Class - the Plugin Class being associated with the identifier
+ @example: `ExternalScripts.register('debugger', Dubugger)`
+ **************************************************************/
+ExternalScripts.register = function(_label, _class) {
+  if(typeof _class == 'function') {
+    if(this._classes[_label]) {
       console.error("Duplicate script constructor for label " + _label);
     }
     else {
-      this._constructors[_label] = _constructor;
+      this._classes[_label] = _class;
     }
   }
 };
@@ -75,11 +72,18 @@ ExternalScripts.nameForScript = function(script) {
   return(script.name || "(unknown)");
 }
 
+
+/**************************************************************
+ @function try
+ @arg {string} functionName - name of the function to invoke
+ @arg {any[]} args - remaining arguments
+ @example `ExternalScripts.try('handleEvent', evt, logger)`
+ **************************************************************/
 ExternalScripts.try = function() {
   // Convert arguments to an array we can safely manipulate:
   // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments
-  const args = (arguments.length === 1 ? [arguments[0]] : Array.apply(null, arguments));
-  const funcName = args.shift();
+  var args = (arguments.length === 1 ? [arguments[0]] : Array.apply(null, arguments));
+  var funcName = args.shift();
   for(var i=0; i < this._scripts.length; i++) {
     var script = this._scripts[i];
     var name = this.nameForScript(script);
@@ -89,15 +93,25 @@ ExternalScripts.try = function() {
         script[funcName].bind(script).apply(script,args);
       }
       else {
-        this.externalScriptError("script " + name + " doesn't respond to  " + funcName);
+        this._externalScriptError("script " + name + " doesn't respond to  " + funcName);
       }
     }
     catch (exception) {
-      this.externalScriptError(exception, {name: name, funcName: funcName});
+      this._externalScriptError(exception, {name: name, funcName: funcName});
     }
   }
 }
 
+/**************************************************************
+ @function handleLogEvent
+ Tell registered scripts to invoke `handleLogEvent` if they can.
+ *******************************************************~*******/
 ExternalScripts.handleLogEvent = function(evt, logger) {
   this.try('handleEvent', evt, logger);
 };
+
+// see jest/external-scripts.spec.js
+// where we consume this module
+if (typeof module !== 'undefined') {
+  module.exports = ExternalScripts
+}
--- a/app/assets/javascripts/external-scripts.js
+++ b/app/assets/javascripts/external-scripts.js
@@ -92,9 +92,6 @@ ExternalScripts.try = function() {
         // script[funcName].apply(script,args) was insufficient …
         script[funcName].bind(script).apply(script,args);
       }
-      else {
-        this._externalScriptError("script " + name + " doesn't respond to  " + funcName);
-      }
     }
     catch (exception) {
       this._externalScriptError(exception, {name: name, funcName: funcName});
--- a/app/assets/javascripts/external-scripts.js
+++ b/app/assets/javascripts/external-scripts.js
@@ -25,7 +25,6 @@ Called from the LARA partial `_lightweight.html.haml`
       scriptUrl -- where the script was loaded from.
       embeddableId -- the embeddableId of the script.
       config -- author configuration.
-      div -- the invisible div on the page for this script to render.
 **************************************************************/
 ExternalScripts.init = function(_label, ctx) {
   var constructor = this._classes[_label];
--- a/docs/external-scripts-api.md
+++ b/docs/external-scripts-api.md
@@ -1,0 +1,30 @@
+## ExternalScripts (original)
+
+* Admins must create a new `ApprovedScript` record.  The record includes the
+    * URL to the remote script
+    * An internal label for reference.
+* ExternalScripts must conform to this (developing) API:
+    * They must call `ExternalScripts.register("label", ExternalScriptClass)` immediately;
+    * `new ExternalScriptClass(config, ctx)` must return an instance of an ExternalScript
+        * `config` is an object of your choice.  Author's can edit this as json when adding a script instance.
+        * `ctx` is the runtime context at page render time. It includes handy propertie for your script:
+        `name`,`scriptLabel`, `scriptUrl`, `embeddableId` and `div`. Perhaps more data will be added to the context soon.
+    * The API for a `ExternalScriptClass` is still being worked out.  Right now it only needs to implement these methods:
+        `handleEvent({event: string, time: number, parameters: {}})`. These are logEvents as per `logger.js`.
+    * see `external-scripts.js`, `external_script.rb`, `app/views/external_scripts/_lightweight` for more information.
+
+----
+
+## NOTES
+- Scripts are inialized when theirview partial `external_scripts/_lightweight` is rendered.
+- This is how scripts can be bound to specific dom elements.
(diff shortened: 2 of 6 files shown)


docs/external-scripts-api.md
```diff
@@ -0,0 +1,30 @@
...
+        * `ctx` is the runtime context at page render time. It includes handy propertie for your script:
+        `name`,`scriptLabel`, `scriptUrl`, `embeddableId` and `div`. Perhaps more data will be added to the context soon.
+    * The API for a `ExternalScriptClass` is still being worked out.  Right now it only needs to implement these methods:
+        `handleEvent({event: string, time: number, parameters: {}})`. These are logEvents as per `logger.js`.
+    * see `external-scripts.js`, `external_script.rb`, `app/views/external_scripts/_lightweight` for more information.
+
+----
+
+## NOTES
+- Scripts are inialized when theirview partial `external_scripts/_lightweight` is rendered.
```
kswenson: "Scripts are initialized when their view partial"

app/assets/javascripts/external-scripts.js
```diff
@@ -89,15 +93,25 @@ ExternalScripts.try = function() {
         script[funcName].bind(script).apply(script,args);
       }
       else {
-        this.externalScriptError("script " + name + " doesn't respond to  " + funcName);
+        this._externalScriptError("script " + name + " doesn't respond to  " + funcName);
```
pjanik: Do we need to log an error?
I guess we could just check if the plugin implements this method and do nothing if it's missing.
That way we won't force scripts to implement dummy `handeEvent`. I'd change it even here, as it won't be breaking change.

docs/lara-plugin-api.md
```diff
@@ -0,0 +1,129 @@
...
+/****************************************************************************
+@function openPopup: Ask lara to add a new popup window
+@arg {IPopupOptions} popupOptions
+@returns A parent DomNode node where component can render content
+****************************************************************************/
+LARA.openPopup(popupOptions: IPopupOptions) : DomNode
+
+
+/****************************************************************************
+@function openPopup: Ask lara to add a new popup window
```
pjanik: I guess it's here by accident.

jest/jest.config.js
```diff
@@ -0,0 +1,5 @@
+
+module.exports = {
+  "bail": true,
+  "verbose": true
+}
```
pjanik: I'm all in for modern tools, but I'm wondering if we should add a new one at this point.
I guess these tests can be written using the existing Jasmine setup.

It would be great to migrate them all together to a technology that we choose. Otherwise, it might be yet another thing to configure and run, or it might die if other developers don't use it (as we'd have jasmine vs jest choice now). Also, I know I was talking about using Jest, but Kirk made a really good point about Cypress using Mocha. That's a significant argument - we could pretty much limit all the syntax and matches to mocha/chai style assuming that we also use Cypress. Constant switching between two styles within one project can be annoying.

docs/lara-plugin-api.md
```diff
@@ -0,0 +1,129 @@
...
+****************************************************************************/
+LARA.addSidebar(sidebarOptions: ISidebarOptions) : DomNode
+
+
+/****************************************************************************
+@function openPopup: Ask lara to add a new popup window
+@arg {IPopupOptions} popupOptions
+@returns A parent DomNode node where component can render content
+****************************************************************************/
+LARA.openPopup(popupOptions: IPopupOptions) : DomNode
```
pjanik: I think @scytacki was voting for keeping everything async or Promise-based.
Personally, I'd keep it like that - simple. If for some reason this needs to be async in the future, we can create a new method and/or update API. Making **everything** returning promises would make the plugin code pretty strange and painful to work with. We'll have tons of nested `.then()` statements or async/await thingies. 
That's my gut feeling / opinion, but I remember this note from @scytacki, so we can use this thread to discuss it.

pjanik: On a related note, I'd rather return a hash (e.g. `{ container: HTMLElement }`) even if we can return basic type. It means that we can extend this API and response without being backward incompatible. And in this case, it's pretty likely that we might provide some handles to close the popup or reposition it, and so on.

scytacki: Would this alternative work:
The IPopupOptions contains a React Class name, and a set of properties.  Then openPopup doesn't need to return anything.  And this could be used by something in an iframe or worker using a LARA provided iframe component, with a property which was a url to the iframe content.

scytacki: This means plugins need to provide react components, but I'm OK with that.

docs/lara-plugin-api.md
```diff
@@ -0,0 +1,129 @@
...
+LARA exports.
+
+```typescript
+
+/****************************************************************************
+// @function addSidebar: Ask lara to add a new sidebar window
+// @arg {ISidebarOptions} sidebarOptions
+// @returns A parent DomNode where component can render content
+****************************************************************************/
+LARA.addSidebar(sidebarOptions: ISidebarOptions) : DomNode
```
scytacki: If my suggestion below about providing a react class and properties works, then it should be used here too.

Also I had considered it might be better for this to be a plugin configuration instead of a API method. So when an admin adds the plugin to lara they specify:
- the url for the javascript
- optional text for a sidebar
- optional image for a sidebar
Then LARA configures the sidebar on the server side, and provides the dom element in the initialization configuration to the plugin. This is less flexible though, the plugin might not want to show the sidebar on every page. So lets try the API approach first and see how well it works.

knowuh:
I am fine with leaving `external-scripts.js` as is for now, and creating a new `plugins.js`. I just wanted to clean things up enough in `external-scripts.js` before moving on.

docs/lara-plugin-api.md
pjanik: Hm, passing React component class (or class name) and set of properties feels a bit opinionated to me. I'd rather pass any HTML element. The plugin can prepare this HTML element whatever way it wants. Using document.createElement, using jQuery, or using React.

Actually, that was my very first proposal. However, after discussing that with @knowuh, we thought that React would fail when we try to render a component into a "detached" HTML container (not part of the page tree yet, as it will be done by LARA later). But I've just tried that and it all works fine:

```typescript
// That's the plugin part:
const container = document.createElement("div");
ReactDOM.render(
  <GlossaryPopup
    word="eardrum"
    definition="An eardrum is a membrane, or thin piece of skin, stretched tight like a drum."
    askForUserDefinition={true}
    onUserDefinitionsUpdate={newUserDefinition}
  />,
  container
);
// now, send container to LARA, e.g.:
// LARA.openPopup({ content: container });
// (...)
// Later, done by LARA:
document.body.appendChild(container);
// it works ;)
```

There's one small drawback of the rendering HTML when elements are detached. You can't measure elements, as they all have 0 width and height (e.g. if you need to measure some dynamic text label). Lab works exactly that way. Lab components use jQuery to render themselves and later some main controller adds them into right spot. This measurment problem is the only one I had while doing it. However, it's very rare that you need it. And even if you do, you can always attach your element to body, measure, and detach (that's what I did in Lab when I had).

So, if we want to make this API simpler, I'd vote for this approach.

docs/lara-plugin-api.md
pjanik: I think that the additional advantage of API approach is that plugins will be 100% self-contained.  Fewer things to configure on staging, production, locally, etc. E.g. if the developer is testing some plugins locally and needs to set them up first, he doesn't need to know that this one requires some sidebar config. The plugin will do it itself providing right text and icon.

docs/lara-plugin-api.md
knowuh: I liked Piotr's sample code, so I have extended his example, trying to pseudo-code a complete implementation, using two different `openPopup` signatures, one in which we provide a DomNode, and one where we get a DomNode back.

These basically look identical to me. The question is are we more worried about handling async return values from `openPopup()` or are we more worried about edge cases in layout behavior when we need to measure things?  I guess I think both are low-risk, and don't care much either way.

1. Pass in the content to the call to openPopup:

```
const definitions = {
  'word': {text: 'word definition'},
  'eardrum': {text: 'An eardrum is a membrane, or thin piece of skin, stretched tight like a drum.'}
  // …
}

const openPopup = (event) => {
  // hand-waving …
  const word = event.target.innerText
  const definition = definitions[word].text
  const {x,y} = event.target.position;

  const container = document.createElement("div")

  ReactDOM.render(
    <GlossaryPopup
      word={word}
      definition={definition}
      askForUserDefinition={true}
      onUserDefinitionsUpdate={newUserDefinition}
    />,
    container
  );

  const options = {
    title: "Glossary",
    position: {screenX: x, screenY: y},
    content: container,
    draggable: true,
    onClose: () =>  { /* ? …  */ }
  }

  LARA.openPopup(options)
}

const words = Object.keys(definitions)
const replace = '<span data-decorationtype="glossary-definition>$1</span>"'
LARA.decorateContent({words: words, replace: replace, listeners: [{wordClick: openGlossaryPopup}]})

```

2. Get the container from the return value of openPopup:

```

const definitions = {
  'word': {text: 'word definition'},
  'eardrum': {text: 'An eardrum is a membrane, or thin piece of skin, stretched tight like a drum.'}
  // …
}

const openGlossaryPopup = (event) => {

  // hand-waving …
  const {x,y} = event.target.position
  const word = event.target.innerText
  const definition = definitions[word].text

  const options = {
    title: "Glossary",
    position: {screenX: x, screenY: y},
    draggable: true,
    onClose: () =>  { /* ? …  */ }
  }

  const container = LARA.openPopup(options)

  ReactDOM.render(
    <GlossaryPopup
      word={word}
      definition={definition}
      askForUserDefinition={true}
      onUserDefinitionsUpdate={newUserDefinition}
    />,
    container
  );


}

const words = Object.keys(definitions)
const replace = '<span data-decorationtype="glossary-definition>$1</span>"'
LARA.decorateContent({words: words, replace: replace, listeners: [{wordClick: openGlossaryPopup}]})

```

app/assets/javascripts/external-scripts.js
```diff
@@ -1,22 +1,34 @@
...
+@function init
+This method is called to initialize the external scripts
+Called from the LARA partial `_lightweight.html.haml`
+@arg {string} label - the the script identifier.
+@arg {any} ctx - context object {} for the script, includes
+      scriptLabel -- same as label above
+      scriptUrl -- where the script was loaded from.
+      embeddableId -- the embeddableId of the script.
+      config -- author configuration.
+      div -- the invisible div on the page for this script to render.
```
pjanik: This seems to be broken in the current implementation. It's jQuery object called with a wrong selector (ID without #). It seems no existing scripts use it, so I'd simply remove it from docs. Glossary Plugin will use it just for demo purposes, before we switch to the new Plugin API.

In the future version, I'd rather pass visible div if a plugin is a page-level plugin. So it can stay invisible (by not rendering anything) or actually render something. Or simply don't pass any container it at all, as for now we don't have any use case for it, so it doesn't make sense to guess requirements.

app/assets/javascripts/external-scripts.js
```diff
@@ -30,43 +42,28 @@ ExternalScripts.init = function(_label, ctx) {
...
-}
-
-
-ExternalScripts.register = function(_label, _constructor) {
-  if(typeof _constructor == 'function') {
-    if(this._constructors[_label]) {
+/**************************************************************
+ @function register
+ Register a new external script as `label` with `_class `
+ @arg label - the identifier of the sctipt
```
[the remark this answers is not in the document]
knowuh: The label is used as a key. When a script registers itself it uses this value. LARA uses this value both to figure out which scripts to put on the page, but also what constructors to call. We could probably devise a different mechanism to coordinate this. We could for example use javascript ClassNames -- and just let the script manage it. But since there is the possibility of name-space collisions, we probably want to the admin to manage it in some way?

In short, I agree this seems redundant on the surface, but I believe some mechanism like this would be required.

app/assets/javascripts/external-scripts.js
knowuh: You are right.  

I had thought that it was in use by this [model-feedback script](https://github.com/concord-consortium/model-feedback/blob/master/src/lara-approved-scripts/feedback-view.ts)  -- but I see now that  it is not.

app/assets/javascripts/external-scripts.js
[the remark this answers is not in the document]
pjanik: Right, looking at the current implementation, I can see what you mean. However, I think it might not be necessary in general.

An author selects a script using pulldown menu. It uses a "name" (which is different from "label"). So, LARA knows which script to add to the page (by adding <script> tag).

Then, plugin calls `ExternalScript.register(MyPluginClass)` providing a constructor to LARA. LARA saves this constructor. At this point, LARA could immediately initialize the plugin too and it should solve the problem. 
Or we do it later - LARA simply iterates over all saved constructors and initializes all the plugins. But we'd need to make sure that it happens after they are all added to the page (perhaps including them in <head> and calling that at the end of <body>, or something like that). Not saying it's super important, but I feel it's possible to simplify that a bit.

docs/lara-plugin-api.md
pjanik: I agree there's no significant difference between them, and I'm fine with both.

I'd say I'd slightly prefer option 1, as content goes together with other options like title, position, etc. So, everything you need to do to open a popup is contained in this one options hash.

The second approach is a bit spread out. You set title and position using options, but then render content using returned element. So, you need to remember what goes where. It'd be more visible if there're more containers returned - e.g. a container for the icon, header container, and so on.

Also, both options can exist together if we ever need it, so API can be easily extended either way.

knowuh: Let us go with option 1 then.  I was against it at first, but have been won over.

But one other thing I am generally worried about (which is probably what @scytacki  was trying to warn us about) is that we can have nested mounted React components.  So if the popup container ever becomes a child of a React mounted component, we might be in trouble. 

Is this something we should be worried about, or can we ignore it for now?

scytacki: I haven't read the discussion above carefully. But here is my 2 cents:

I was mainly trying to find a way so it would be easy to use the same API from an iframe or worker. So having a hard rule that everything needs to async, would help with that.  But it also means that you can't pass dom elements and listeners around.  

In addition to this it would be useful to plan for a fully React LARA UI that was a SPA.  This is where I'd like LARA to go. And it would be great if we could get there by pulling out components from LARA and making them interactives or plugins.  And then those same interactives and plugins could work both in the SPA LARA and the current LARA.  It sounds like from the last comment that, this would not work because the glossary could not embed itself inside of a already formed React tree.

But given all of this, I think it is best to get started building option 1 as you decided. And if we have time we can consider a refactor that sets us up for both iframes and a React SPA.

pjanik: I think option 1 and 2 can exist together. React + props could be handled too. I don't think it would complicate LARA and plugins will have a choice.

> I was mainly trying to find a way so it would be easy to use the same API from an iframe or worker.

This still seems impossible with React component name + props approach if I understand you correctly. We can't pass component class thorugh iframe, and even if we pass name as a string, this component implementation won't be available on the LARA side.

The only option I can see here is that iframe plugin could send content as a string like ```{ content: "<iframe src='some-component.concord.org/view-a.html'/>" }``` and LARA could interpret this string (trivial using jQuery). Actually, it would be yet another option that LARA could support without much work.

> So if the popup container ever becomes a child of a React mounted component, we might be in trouble.

That's a good question. It feels like a possible problem. However, I don't know if it would break things or not. I guess it should be pretty easy to test and see what happens.

pjanik:
I'm going to merge it as it'll be useful for my work. I think most of the discussions are resolved, and we can still continue them in other places. I think that API will change quite a lot during development and we shouldn't keep this PR open until there's a final version. :wink:

PR #321, opened 2018-09-03
Upgrade React to v16.4.2

pjanik:
[#160204811]

Most of the scripts are included by putting a library in assets and requesting it in application.js and runtime.js. It's not super fancy approach, but it's straightforward. And I think it's better than trying to include some gems or using rails assets. The modern world seems to use https://github.com/rails/webpacker already, so I'd keep as simple as possible, so it's easier to migrate later (gems quite often have some config and lots of unnecessary features, e.g. react-rails).

I had to add React, ReactDOM, create-react-class helper, core-js, and react-test-utils (but that's added only to Jasmine tests). Main changes:

1. React.DOM factories removed -> I've implemented a custom helper in react-factories.js
2. React.createClass removed -> create-react-class
3. React.findDOMNode removed -> ReactDOM.findDOMNode
4. New refs syntax and handling.
5. core-js provides polyfills. Mostly needed for Jasmine tests, but might help with some older browsers too (if we care about them).
6. Updated Jasmine tests.

Actually, the biggest pain and most of the time was eaten by Jasmine tests. Which are painful to run and slow. People seem to assume that testing libraries are imported using CommonJS / node (e.g. try to find UMD build of react-test-tools). Plus, I think the `jasmine:ci` mode is impossible to run locally due to a bug in jasmine-gem.

I briefly tested:
1. ITSI authoring
2. Interactives authoring and runtime (table-interactive is a good test, as it supports authored state).
3. Tracked questions UI.

and they seemed fine, but @eireland should focus on these features when she gets to test this change.

Two side notes:

1. webpacker looks pretty good, it's included by default in Rails 5.1, setup seems easy and it seems to work together with assets pipeline. I think a migration wouldn't be that bad if we were only using rails 4.2+. I was even tempted to try it before I noticed 4.2+ requirment. It would enable us to write modern JS. Every line of JS code that we add to LARA at this point will become a tech debt soon. 

2. Same problems with Jasmine tests. @knowuh tried Jest and I think we should go that direction. It requires some config, but if we keep JS stuff separate and not to complex, hopefully it will be possible to integrate that one day with Webpacker.


db/schema.rb
```diff
@@ -172,10 +172,10 @@
     t.text     "prediction_feedback"
     t.boolean  "is_hidden",                        :default => false
     t.text     "hint"
-    t.boolean  "is_full_width",            :default => false
-    t.boolean  "show_in_featured_question_report", :default => true
+    t.boolean  "is_full_width",                    :default => false
     t.integer  "interactive_id"
     t.string   "interactive_type"
+    t.boolean  "show_in_featured_question_report", :default => true
```
knowuh: There is frequently weird churn in `schema.rb` where the changes are simply whitespace, or the order in which fields appear.  I am not sure what we can do about it.  Usually I review the changes, and if its whitespace only, I just replace my `schema.rb` with  whatever is checked in.

scytacki:
Looks good to me too.

db/schema.rb
pjanik: I do the same. I usually try not to commit schema.rb with whitespace changes only. But here I had so many files to commit that I didn't notice it probably.

PR #323, opened 2018-09-04
Authors can add Plugins to an activity

knowuh:
Includes modeling and views for authors to add plugins to their activities.

After consulting with @scytacki I decided to use simple remote form pattern for this feature instead of react components.  We recognize that react might be a better long term solution, but continuing the existing authoring patterns would be sufficient for this.

Pages initialize the activity plugins on every page, except in the case of single page  activities, in which case the Plugins are only initialized once.

There was an old bug in a ajax function in `common.js`. We never noticed it because this function  wasn't actually every invoked, because none of our remote forms used the `data-replace` attribute, so updating it was low-risk.  I verified authoring  and runtime continued  to work as expected.

### DIFF
--- a/app/views/plugins/_show.haml
+++ b/app/views/plugins/_show.haml
@@ -1,0 +1,19 @@
+- runtimeDiv = "output-#{plugin.id}"
+= content_for :external_scripts do
+  %script{src:plugin.url}
+
+.output{id:runtimeDiv}
+
+:javascript
+  // Begin script for #{plugin.name}
+  $(document).ready( function() {
+    env = {
+      name: '#{plugin.name}',
+      scriptLabel: '#{plugin.label}',
+      scriptUrl: '#{plugin.url}',
+      pluginId: '#{plugin.id}',
+      config: '#{ j plugin.author_data }',
+      div: $('##{runtimeDiv}')
+    }
+    ExternalScripts.init('#{plugin.label}', env);
+  });
--- a/app/assets/javascripts/common.js
+++ b/app/assets/javascripts/common.js
@@ -7,7 +7,7 @@ $(function () {
         .data('type', 'html')
         .on('ajax:success', function (event, data) {
             var $this = $(this);
-            $($this.data('replace')).html(data.html);
+            $($this.data('replace')).html(JSON.parse(data).html);
             $this.trigger('ajax:replaced');
         });
 });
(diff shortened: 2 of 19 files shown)


app/views/plugins/_show.haml
```diff
@@ -0,0 +1,19 @@
...
+
+:javascript
+  // Begin script for #{plugin.name}
+  $(document).ready( function() {
+    env = {
+      name: '#{plugin.name}',
+      scriptLabel: '#{plugin.label}',
+      scriptUrl: '#{plugin.url}',
+      pluginId: '#{plugin.id}',
+      config: '#{ j plugin.author_data }',
```
pjanik: `j plugin.author_data` is it a typo or some magic Rails helper (JSON parsing)?
If so, should we agree on one common format of Plugin state (both authored and user)? I'd assume it's always `string` or `JSON`. A string can be saved in JSON and JSON can be stringified, so it doesn't matter to me. Although I'd make sure we use one format so we don't have to make checks everywhere. So far I've added user state as `string` to LARA API docs.

app/views/plugins/_show.haml
```diff
@@ -0,0 +1,19 @@
+- runtimeDiv = "output-#{plugin.id}"
+= content_for :external_scripts do
+  %script{src:plugin.url}
+
+.output{id:runtimeDiv}
```
pjanik: Could we render that in some reasonable place and make sure it's not hidden? I guess plugin-output could be a more descriptive class name too.
E.g. I could imagine that to be at the end of the activity, so a plugin can render some status / output without much effort (even for debugging).

app/views/plugins/_show.haml
knowuh: Yes, this `j` is rails magic.  

It escapes text for including in javascript.  It has a better, longer, and more descriptive function name `escape_javascript` , which  I will switch to here to better document what is happening.

I will also  update the `_show.html.haml` for the ExternalScripts to use this `escape_javascript` .  I had a half-baked regex I hand-rolled previously.

app/views/plugins/_show.haml
knowuh: Yes -- For sure.  I was just copying the pattern we had as a first pass.  I  will be making further commits to improve this.

knowuh:
You are absolutely correct:  

We will be moving everything to the new API soon.  I just wanted to drop in something that used the old API first.

pjanik:
@knowuh, forgot about it then, but it'd be nice to have some tests here I guess. :wink:

PR #324, opened 2018-09-05
Update LARA.addPopup implementation and make all LARA dialogs use LARA.addPopup

pjanik:
[#160096504]

I think the most important part is API changes. I've renamed LARA.openPopup to LARA.addPopup as now it doesn't have to open it automatically. Also, it returns the popup controller (pretty simple). This works better for some old-school, jQuery based implementations that rely on existing DOM content. 

Then, I've replaced all the jQuery.dialog uses with LARA.addPopup. The small problem is that we don't have tests covering this UI. I've tried to manually test all of that. In some cases, there are small style changes, but the huge benefit is that now things will be more consistent, we'll have this API tested, and it'll be easier to replace implementation if necessary and still keep all the dialogs looking the same.

Updated dialogs (it might be useful for @eireland to note them):
- Labbook
- Image question (including errors, e.g. network error)
- Multiple choice answer "check answer" dialog
- Share report dialog (visible in ITSI them)
- Unauthorized dialog (can be triggered by `$(document).trigger(`unauthorized`);`


app/assets/javascripts/runtime.js
```diff
@@ -41,7 +41,6 @@
 //= require video
 //= require interactive-sizing
 //= require embeddable-carousel
-//= require image_question
```
knowuh: Just noticed `image_question` was removed. 

I guess it was superseded by `image_question_drawing_tool`?  Maybe it is something that should have been removed a while ago?

pjanik: Exactly, it should have been removed long time ago. That was an old version that was using svg-edit editor.

PR #325, opened 2018-09-05
Lara plugin changes.

knowuh:
* Adds version to approved_script model.  Version 1 expects old ExternalScript API.  Version 2 expects new LARA Plugin Api.

* Adds registration methods in `lara-api.js`
* Adds API for decorating text in `lara-api.js`
* Renames labels in the UI to be more "plugin"-ish and less "External Script"-ish.

TODO: The way the text-decoration library was added was simply with a javascript include tag.  This must of course be cleaned up.

@pjanik 
This entire PR will require more cleanup later, but I wanted to get it in your hands before I had to leave this evening, incase you have some time to review / test.

### DIFF
--- a/app/assets/javascripts/lara-api.js
+++ b/app/assets/javascripts/lara-api.js
@@ -149,5 +149,119 @@ window.LARA = {
   ****************************************************************************/
   saveUserState: function (pluginInstance, state) {
     console.log('Plugin', pluginInstance, 'wants to save a state:', state);
+  },
+
+
+  /****************************************************************************
+  @function decorateContent: Ask LARA to decorate authored content (text / html)
+  @arg {string[]} words - a list of case-insensitive words to be decorated
+      can use limited regex
+  @arg {string} replace - the replacement string.
+      Can include '$1' representing the matched word.
+  @arg {IEventListeners} listeners - one or more { type, listener } tuples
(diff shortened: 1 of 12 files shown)


app/assets/javascripts/lara-api.js
```diff
@@ -149,5 +149,119 @@ window.LARA = {
...
+  @arg {IRuntimContext} runtimeContext - context for the plugin
+
+  // Its likely we dont need most of these context values. TBD.
+  Interface IRuntimeContext {
+    name: string;         // Name of the plugin
+    url: string;          // Url from which the plugin was loaded
+    pluginId: string;     // Active record ID of the plugin scope id
+    authorData: string;   // The authored configuration for this instance
+    learnerData: string;  // The saved learner data for this instance
+    div: DomNode;         // reserved DomNode for the plugin in ther runtime.
```
pjanik: I guess that's HTMLElement in TypeScript. :) No need to fix it now, but just for the future reference. I'm assuming that one beautiful day we can really copy-paste these interfaces and use them.

pjanik:
btw, did decoration worked for you? I can see underline, but click handlers are ignored. It seems text-decorator library is a bit broken when it comes to event handlers.

PR #327, opened 2018-09-07
Revert "Include text-decorator using assets pipeline"

knowuh:
OK -- This is PR is only here for two reasons: 

1. To see if travis passes again
2. For a discussion about what to do about asset pipeline being so antiquated.

The problem was that sprockets was dying on 'const' keywords in `text-decorator-1.0.0.umd.js`  

javascript console errors also reported that the uglify gem (also part of asset pipeline) could be configured to transpile with the correct options:  `Uglifier.new(harmony: true)` -- which actually we configure correctly AFIK in `env/production.rb`.   Perhaps it would be possible to configure `env/test.rb` and `env/development.rb` to do something similar?

I did an experiment where I compiled `text-decorator-1.0.0.umd.js` to ES5 using babel CLI, but then I got other console error messages that seemed to indicate that the exports were not being handled correctly. 

I looked at the `text-plugin` build to see if I could configure it to build umd in ES5 style.  I couldn't figure out how to get that to work with babel/rollup/typescript.  This might be a good direction to look.

But I think ultimately the problem is that asset pipeline is getting out of date.  We have talked about switching before, but I am not sure how big a lift that would be in the case of LARA.

This reverts commit 4a5a6b811c596aa486ec38c1704f6131101aed8c.


pjanik:
There were es5 and umd versions of text-decorator, I picked umd as es5 looked like non-umd. We need umd + es5 apparently. It might be already there or we might need to change rollup config, but I'm also not familiar with it. It should be doable for sure.

Re assets pipeline, since things work fine in the browser, I think it's more about ancient version of PhantomJS / test environment. I haven't noticed  errors in the browser console, but I might have missed them.

Webpacker gem looks like the very reasonable replacement (already shipped with new Rails versions). It would greatly help with testing too and architecture (since we could finally use proper module loading). But it requires Rails 4.2.

pjanik:
This has been already fixed by https://github.com/concord-consortium/lara/pull/328

PR #330, opened 2018-09-11
Plugin learner state saving

knowuh:
This PR adds routes and controllers to save the PluginLearnerState.

A few strange thing:  

* The API changed.  In the GlossaryPlugin the function name for saving plugin data changed to `saveLearnerState` which actually seems like a better name.  So I changed it in LARA to match.
 
* When the plugin is registered and initialized, we save some configuration details in `__LaraPluginContext ` of the Plugin instance that is returned.  This is a gross violation of many expectations and norms.  Eventually I would prefer that we initialize the Plugin with customized API object instead.

It covers this PT Story: [#160096573](https://www.pivotaltracker.com/story/show/160096573)

> A Plugin can save and load user state (on page load) using the Lara Plugin API.

### DIFF
--- a/app/assets/javascripts/lara-api.js
+++ b/app/assets/javascripts/lara-api.js
@@ -1,6 +1,5 @@
 
 window.LARA = {
-  PluginsApi: function() { return window.Plugins} ,
 
   /****************************************************************************
    @function addPopup: Ask LARA to add a new popup window
@@ -163,13 +162,7 @@ window.LARA = {
    @returns Promise
   ****************************************************************************/
   saveLearnerPluginState: function (pluginId, state) {
-    var pluginsApi = window.Plugins;
-    if (pluginsApi && typeof pluginsApi.saveLearnerPluginState === 'function') {
-      return pluginsApi.saveLearnerPluginState(pluginId, state);
-    }
-    return new Promise( function(resolve, reject) {
-      reject('pluginApi not defined. saveLearnerPluginState failed.')
-    });
+    return Plugins.saveLearnerPluginState(pluginId, state);
   },
 
   /****************************************************************************
@@ -208,10 +201,6 @@ window.LARA = {
    @example: `LARA.registerPlugin('debugger', Dubugger);
    **************************************************************/
   registerPlugin: function(label, _class) {
-    var pluginsApi = window.Plugins;
-    if(pluginsApi && pluginsApi.registerPlugin){
-      return pluginsApi.registerPlugin(label, _class);
-    }
-    return false;
+    return Plugins.registerPlugin(label, _class);
   }
 };
--- a/app/assets/javascripts/lara-api.js
+++ b/app/assets/javascripts/lara-api.js
@@ -1,4 +1,7 @@
+
 window.LARA = {
+  PluginsApi: function() { return window.Plugins} ,
+
   /****************************************************************************
    @function addPopup: Ask LARA to add a new popup window
    @arg {IPopupOptions} popupOptions
@@ -142,30 +145,30 @@ window.LARA = {
   },
   /****************************************************************************
   @deprecated saveLearnerState
-  @see savePluginState
+  @see saveLearnerPluginState
   ****************************************************************************/
   saveLearnerState: function (pluginId, state) {
     var deprication =
-      'saveLearnerState is depricated, use `savePluginState` instead'
+      '⚠️ saveLearnerState is depricated, use `saveLearnerPluginState` instead'
     console.warn(deprication);
-    this.savePluginState(pluginId, state);
+    this.saveLearnerPluginState(pluginId, state);
   },
 
   /****************************************************************************
-   @function savePluginState: Ask LARA to save the users state for the plugin
+   @function saveLearnerPluginState: Ask LARA to save the users state for the plugin
    @arg {string} pluginId - ID of the plugin trying to save data, initially passed to plugin constructor in the context
    @arg {string} state - A JSON string representing serialized plugin state.
    @example
-    LARA.savePluginState(plugin, '{"one": 1}').then((data) => console.log(data))
+    LARA.saveLearnerPluginState(plugin, '{"one": 1}').then((data) => console.log(data))
    @returns Promise
   ****************************************************************************/
-  savePluginState: function (pluginId, state) {
+  saveLearnerPluginState: function (pluginId, state) {
     var pluginsApi = window.Plugins;
-    if (pluginsApi && typeof pluginsApi.savePluginState === 'function') {
-      return pluginsApi.savePluginState(pluginId, state);
+    if (pluginsApi && typeof pluginsApi.saveLearnerPluginState === 'function') {
+      return pluginsApi.saveLearnerPluginState(pluginId, state);
     }
     return new Promise( function(resolve, reject) {
-      reject('window.Plugins not defined. savePluginState failed.')
+      reject('pluginApi not defined. saveLearnerPluginState failed.')
     });
   },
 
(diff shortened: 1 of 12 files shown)


app/models/plugin.rb
```diff
@@ -11,6 +11,11 @@ class Plugin < ActiveRecord::Base
   delegate :url,   to: :approved_script, allow_nil: true
   delegate :version, to: :approved_script, allow_nil: true
 
+  after_initialize :generate_rare_key
```
pjanik: Is it desired that we automatically set plugins to have shared state? I don't really know, just thinking loudly.
If we always generate UUID (that can be overwritten in case of need), we could have one lookup strategy less - `PluginAndRun` seems redundant in this case. Unique `shared_learner_state_key` would handle it.

app/assets/javascripts/plugins.js
```diff
@@ -23,12 +23,13 @@ window.Plugins = {
...
-    pluginId: string;     // Active record ID of the plugin scope id
-    authorData: string;   // The authored configuration for this instance
-    learnerData: string;  // The saved learner data for this instance
-    div: HTMLElement;     // reserved HTMLElement for the plugin output
+    name: string;               // Name of the plugin
+    url: string;                // Url from which the plugin was loaded
+    pluginId: string;           // Active record ID of the plugin scope id
+    authorData: string;         // The authored configuration for this instance
+    learnerData: string;        // The saved learner data for this instance
+    pluginUserStatePath: string // Where to save / load data from for this instance.
```
pjanik: I wouldn't really pass it to the plugin, so they can't mess with it. E.g. someone can write plugin posting requests manually and we won't be able to change paths, underlying implementation, etc. If we force plugin to use JS API, we have more flexibility.

app/assets/javascripts/plugins.js
```diff
@@ -40,6 +41,10 @@ window.Plugins = {
         plugin = new constructor(runtimeContext);
         this._plugins.push(plugin);
         this._pluginLabels.push(label);
+        // FIXME: We need some way to costomize the save & load paths.
+        // This isn't great. Maybe we change how we store plugins so that
+        // we can save our own copy of `runtimeContext` without messing with the plugin.
+        plugin.__LaraPluginContext = runtimeContext;
```
pjanik: What about something like this:
```javascript
initPlugin: function(label, runtimeContext, pluginUserStatePath) {
// ...
   this._learnerStatePaths[plugin] = pluginUserStatePath;
  // or  this._learnerStatePaths[runtimeContext.id] = pluginUserStatePath;
}
```

and then you'd have access to it later.

app/assets/javascripts/lara-api.js
```diff
@@ -142,13 +142,37 @@ window.LARA = {
...
    @returns Promise
+   TOOD:
+    * Rename to savePluginState (because of plugin instance argument)
+    * Add more generic function saving unspecified resources … ?
+    * Use Promise polyfill for IE support ?
   ****************************************************************************/
-  saveUserState: function (pluginInstance, state) {
+  saveLearnerState: function (pluginInstance, state) {
     console.log('Plugin', pluginInstance, 'wants to save a state:', state);
+    var context = pluginInstance.__LaraPluginContext
```
pjanik: If this functionality was delagated to `Plugins` helper, we could use its internal state to get the path that we need (or any other context).

knowuh:
016204d – resolves some issues brought up in discussion around a cleaner LARA JS api.  We no longer pass path info into the plugins, we no longer modify plugins when they register themselves, and we have the plugin handle the data saving.

knowuh:
> maybe some tests of the rails code that checks various lookup methods.

I created [this spec](https://github.com/concord-consortium/lara/blob/fd59fb63a7b1def77eaba3270cc7590b33f3377d/spec/models/plugin_learner_state_spec.rb) before I wrote `plugin_learner_state.rb` -- it helped drive the design of that class.  I could provide more tests, but I this at least it captures the lookup methods.

More coverage is needed for the [controller spec](https://github.com/concord-consortium/lara/pull/330/files#diff-e0430c5eedf8c748d75ebfe1e35bddd2) for sure.  Similarly I think we need to protect those routes.

Jest tests for `lara-api.js` and `plugin.js` are on their way.

pjanik:
> I created this spec before I wrote plugin_learner_state.rb -- it helped drive the design of that class. I could provide more tests, but I this at least it captures the lookup methods

Ah, I forgot about them and that's why I was asking. They look nice and it's enough for sure.


Re `saveLearnerState` => `savePluginState`. I'm concerned about two things:
1. I think it's nice if save method corresponds to name of the state that we pass in the context. So, `saveLearnerState` means that you'll receive that as `context.learnerState`. If we go that way, I'd rename `context.learnerState` to `context.pluginState`. But it feels a bit wrong to me, as it's state of the plugin + user combination, not just plugin.
2. I think that what we're actually saving now is a learner-specfic state. I can definitely see use cases for a generic plugin state (not to be bound to any specific user) and `savePluginState` method too. But it feels to me that they could be separate methods. E.g. I could use savePluginState to count how many users provided their own definiton of a given word in a very specifc glossary and log that or display in runtime. I'd call this kind of state a "plugin state". So, something bound to an instance, but not user.

In general, I'd rather use user/learner/student word in this API and assume that we might implement separate `savePluginState` method in the future.

app/assets/javascripts/lara-api.js
```diff
@@ -140,15 +143,33 @@ window.LARA = {
...
-  saveUserState: function (pluginInstance, state) {
-    console.log('Plugin', pluginInstance, 'wants to save a state:', state);
+  saveLearnerPluginState: function (pluginId, state) {
+    var pluginsApi = window.Plugins;
+    if (pluginsApi && typeof pluginsApi.saveLearnerPluginState === 'function') {
+      return pluginsApi.saveLearnerPluginState(pluginId, state);
+    }
+    return new Promise( function(resolve, reject) {
+      reject('pluginApi not defined. saveLearnerPluginState failed.')
+    });
```
pjanik: Is there any scenario when this can happen (PluginApi not available)? I guess it would be some total collapse of LARA JS env. Actually, it might be even beneficial if everything fails, some errors are thrown, don't get handled, and finally Rollbar reports them to us. 

In general, I'm not sure if the nice handling of unexpected disasters is desired. It might silence any signs that they happened. Plugin might not care about this Promise and the fact it got rejected. I'm not too concerned about it, as I think it's unlikely this line of code will be ever executed tho. Just an observation. :wink:

app/assets/javascripts/lara-api.js
```diff
@@ -1,4 +1,7 @@
+
 window.LARA = {
+  PluginsApi: function() { return window.Plugins} ,
```
pjanik: Seems unused.

app/assets/javascripts/lara-api.js
```diff
@@ -140,15 +143,33 @@ window.LARA = {
...
+    var deprication =
+      '⚠️ saveLearnerState is depricated, use `saveLearnerPluginState` instead'
+    console.warn(deprication);
+    this.saveLearnerPluginState(pluginId, state);
+  },
 
   /****************************************************************************
-   @function saveUserState: Ask LARA to save the users state for the plugin
-   @arg {ILaraPluginRef} pluginInstance - The plugin trying to save data
+   @function saveLearnerPluginState: Ask LARA to save the users state for the plugin
```
pjanik: Nice, now I get what was the intention. saveLearnerState wasn't really related to the plugin. This seems perfect, thanks.

app/assets/javascripts/lara-api.js
knowuh: Oh -- yes thanks.  Left over from some other experiment.

knowuh:
@pjanik -- in 53a11eb I cleaned up the api a little, and moved some tests into `plugins.spec.js`

If I can sneak more time into this project I might move towards adding Plugins to pages.

Thanks for another look.

pjanik:
Looks great. I was thinking that actually some tests mocking Ajax calls would be great too and I can see them now. Thanks :wink:

PR #332, opened 2018-09-13
Implement LARA.addSidebar

pjanik:
[#160096518]

This PR adds implementation of `LARA.addSidebar`. It follows patterns started by `LARA.addPopup`. Sidebar implementation is partially based on our old, static markup from HAML and old styles. I've implemented short screens handling. Client code just can't specify height - docs describe how it works. I hope it's pretty clear and it ensures that we we won't have issues with different screen sizes. I've added Jest tests.

Again, I've moved implementation that to some helper module, so lara-api.js doesn't grow to thousands of lines and doesn't have a mix of unrelated helper functions in one file. I'm assuming that lara-api is just set of well-documented, well-tested, quite often unrelated utilities that can be used either by the internal code or plugins.


app/assets/javascripts/sidebar.js
```diff
@@ -1,29 +1,124 @@
...
+      $sidebar.append($handle);
+      $sidebar.append($body);
+      $('body').append($sidebar);
+
+      // Add event handlers.
+      var isOpen = function () {
+        return $sidebar.hasClass('expanded')
+      };
+      $handle.add($closeBtn) // .add creates a set of elements, so we can apply click handler just once
+        .on('click.sidebarToggle', function () {
```
knowuh: I have a feeling I am about to learn something about jQuery here.  

What is this string 'click.sidebarToggle'?  It looks like a custom event, but I guess I am not sure how it works.

Is there something special about the string?  I can't seem to find other references in the source?

I couldn't find it in jQuery docs, but by playing with it in the console, it seems exactly the same as `on('click')`

Maybe we can have other elements listen for the special event `click.sidebarToggle`

spec/controllers/interactive_page_controller_spec.rb
```diff
@@ -186,11 +186,11 @@
 
     it 'does not show page areas which are not selected to be shown' do
```
knowuh: Maybe we can use the it title from line 164 (`calls LARA.addSidebar content on pages which have authored sidebar`
to here, and then delete that spec It seems redundant because this one tests both cases.

Also It wasn't obvious to me at first that the Page factory was creating pages with sidebars by default.

app/assets/javascripts/sidebar.js
pjanik: This is a namespace. It's quite similar in other even handling libraries (EventEmitter2 etc).

It's most useful when you want to cleanup event handlers. If you just do `$elem.off('click')`, it would remove all the click handlers, even ones added by other scripts. So, instead you can call `$elem.off('click.sidebarToggle')` to remove handlers that were added by you only. Or, `$elem.off('.sidebarToggle')` to remove all handlers with a given namespace (click, touch, mouseover, and so on).

I added it here, as this function used to be called multiple times for a single element and I had to call `.off` first. Since the implementation changed and I never use this namespace, yes, it's redundant now.

pjanik:
Thanks. I've removed jQuery namespace, improved rspecs as you suggested and reverted unnecessary schema change.

PR #336, opened 2018-09-21
160129986 responsive layout question column 320px

knowuh:
As per this PT story: https://www.pivotaltracker.com/story/show/160129986

This PR is a simple style change for responsive layout.   We change the fixed-width question column from 250px to 320px in 850bd59.

I also added an content overflow fix in 84fff53.

It wasn't part of the request story, but it was simple to fix, and seemed related.

This is such a small PR I assigned it to both @scytacki  and @dougmartin 

But seriously -- if you can think of any reason a question-prompt should be allowed to overflow, let me know.  That is the main reason I am looking  for a review at all.

### DIFF
--- a/app/assets/stylesheets/partials/_responsive-layout.scss
+++ b/app/assets/stylesheets/partials/_responsive-layout.scss
@@ -46,7 +46,7 @@
   .question {
     display: block;
     vertical-align: top;
-    width: 250px;
+    width: 320px;
   }
 
   .embeddables {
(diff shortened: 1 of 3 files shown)


scytacki:
I'll go ahead and merge this and get it deployed so we can try it out on staging. @dougmartin if you have time to take a look I'd also be curious if you can think of a reason we'd want overflow to not be hidden.

PR #338, opened 2018-10-02
Fixed size of question column is now 365px.

knowuh:
This story originally asked for 320px width questions column in the responsive layout.

But after looking at it, 360px is the desired width.

[#160129986]
https://www.pivotaltracker.com/story/show/160129986


knowuh:
One line review @scytacki  ?

PR #341, opened 2018-10-09
Changes to Interactive Iframe Layout

knowuh:
## Interactive Authoring form changes: 
* Dropdown select for the aspect_ratio_method
* Hide native_width & native_height fields when using "DEFAULT" or "MAX" `aspect_ratio_method`, but make visible for "MANUAL" `aspect_ratio_method`.
* I18n for aspect ratio method names.

## Update the scaling logic in `interactive-sizing.js`

* Take into account `aspect_ratio_method`
* Limit the iFrame interactive height to the visible viewport and scale width accordingly to maintain aspect ratio.

PT Stories:
* https://www.pivotaltracker.com/story/show/160601693
* https://www.pivotaltracker.com/story/show/160601386

### DIFF
--- a/app/assets/javascripts/interactive-sizing.js
+++ b/app/assets/javascripts/interactive-sizing.js
@@ -5,17 +5,22 @@ function interactiveSizing () {
     var $iframe = $(this);
     var aspectRatio = $iframe.data('aspect-ratio');
     var resizeMethod = $iframe.data('aspect-ratio-method');
-    var currWidth = $iframe.width();
+    var magicNumber = 125;
+    var maxHeight = window.innerHeight - magicNumber;
 
     if(resizeMethod === 'MAX') {
-      magicNumber = 125;
-      newHeight = window.innerHeight - magicNumber;
-      $iframe.attr('height', newHeight);
+      $iframe.height(maxHeight);
+      $iframe.attr('width', '100%');
     }
     else {
-      $iframe.attr('height', currWidth / aspectRatio);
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/interactive-sizing.js
```diff
@@ -4,16 +4,31 @@ function interactiveSizing () {
   function setSize () {
     var $iframe = $(this);
     var aspectRatio = $iframe.data('aspect-ratio');
-    var currWidth = $iframe.width()
-    $iframe.attr('width', '100%');
-    $iframe.attr('height', currWidth / aspectRatio);
+    var resizeMethod = $iframe.data('aspect-ratio-method');
+    var magicNumber = 125;
```
scytacki: Different themes could change the size of the header or other things. Is this magic number necessary? Is it the same way the pinning code computes the maxHeight?

scytacki:
Regarding specially handling of the question block, perhaps it is good when the questions are full-width, for the MAX option to work so the interactive is a large as possible for a student to see all of it without scrolling.  However when it is in the question block and the layout is 40-60.  Using MAX would be really weird, but I guess we can let the author make that mistake.

I now realize I was wrong with the sticky height stuff. That line was just comparing the height of the interactive to the height of the window.  And in that case it is using a magic number of 75. It that case it is also using the jQuery height. I'm not sure if jQuery height is different than a basic innerHeight.

So based on all of this I'm going to merge and deploy this to staging so we can demo and get reaction.

PR #343, opened 2018-10-23
160836352 teacher edition

knowuh:
[DL/NP] Initial LARA Teacher Edition.

<hr/>

[#160836352]
https://www.pivotaltracker.com/story/show/160836352

## Full Story:

Add a new mode to LARA for launching a teacher edition of an activity and sequence. Initially this edition will just include a banner saying 'Teacher Edition'. This banner needs to show on every single page of the activity. This includes sequence and activity home pages.

In the future stories this will be appended with:
- open response recommended answers
- html teacher notes
- interactive  teacher notes

It seems best to implement this using a URL parameter or path. And this parameter or path needs to be carried onto every page that is linked in an activity. 

## Implementation Details:

1. Banner display is very simplistic until final styling has been decided;
1. Added a technique to decorate links to pass through "white-listed" query params in the url's -- may need to be extended or revisited with follow on stories;
1. A particular complication was handling the `redirect_to` logic -- in the run controller -- to correctly continue to pass the pertinent URL data in addition to the white-listed params; and,
1. A fairly extensive Capybara spec test has been implemented for most of the obvious happy-paths.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -198,7 +198,7 @@ module ApplicationHelper
   end
 
   def teacher_content
-    params['mode'] == 'teacher-view'
+    params['mode'] == 'teacher-edition'
   end
 
   def pass_white_list_params(url_or_path, whitelist=default_param_whitelist)
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -190,4 +190,13 @@ module ApplicationHelper
     }.html_safe
   end
 
+  def pass_white_list_params(url_or_path)
+    mode = params[:mode]
+    if (url_or_path and mode)
+      "#{url_or_path}?mode=#{mode}"
+    else
+      url_or_path
+    end
+  end
+
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -197,6 +197,10 @@ module ApplicationHelper
       ["mode"]
   end
 
+  def teacher_content
+    params['mode'] == 'teacher-view'
+  end
+
   def pass_white_list_params(url_or_path, whitelist=default_param_whitelist)
     # Construct query string from the contents of a url and those parameters
     # that are whitelisted, passing thruough the ones in the whitelist and
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -91,7 +91,7 @@ module ApplicationHelper
   def default_footer
     <<-EOF
       <p class="footer-txt">
-        Copyright © 2018 <a href="http://concord.org/">The Concord Consortium</a>. 
+        Copyright © 2018 <a href="http://concord.org/">The Concord Consortium</a>.
         All rights reserved. This activity is licensed under a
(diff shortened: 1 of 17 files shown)


app/helpers/application_helper.rb
```diff
@@ -190,4 +190,35 @@ def editable_field (object, property, options={})
...
 
+  # The default is hardcoded, here, in place, for the time being. If and
+  # when the functionality is extended, the white-list could be located in
+  # a more reasonable location in the code-base.
+  def default_param_whitelist
+      ["mode"]
+  end
+
+  def teacher_content
+    params['mode'] == 'teacher-view'
```
scytacki: I think the name should be `teacher-edition`.  
`teacher-view` could be confused with what teacher's see when previewing activities in the portal.

app/views/shared/_sequence_next_prev.html.haml
```diff
@@ -3,11 +3,11 @@
 .intra-sequence-nav
   .next-activity{ :class => ('disabled' if next_activity.nil?) }
     - if next_href
-      = link_to 'Next Activity >', sequence_activity_path(sequence, next_activity), { :title => next_activity.name }
+      = link_to 'Next Activity >', runnable_activity_path(sequence, next_activity), { :title => next_activity.name }
```
scytacki: This looks broken. `runnable_activity_path` doesn't seem to take a sequence as a param.

app/views/shared/_sequence_next_prev.html.haml
```diff
@@ -3,11 +3,11 @@
...
     - else
       Next Activity >
   .last-activity{ :class => ('disabled' if previous_activity.nil?) }
     - if last_href
-      = link_to "< Last Activity", sequence_activity_path(sequence, previous_activity), { :title => previous_activity.name }
+      = link_to "< Last Activity", runnable_activity_path(sequence, previous_activity), { :title => previous_activity.name }
```
scytacki: This looks broken. `runnable_activity_path` doesn't seem to take a sequence as a param.

knowuh: Right, [`runnable_activity_path`](https://github.com/concord-consortium/lara/blob/master/app/helpers/lightweight_activity_helper.rb#L12) doesn't accept a sequence.  Sort of weird that this didn't blow up more noticeably.

`runnable_activity_path` It depends on an instance variable `@sequence` being set in the [app controller](https://github.com/concord-consortium/lara/blob/master/app/controllers/application_controller.rb#L108)

Oops.

knowuh:
Hello @scytacki  --  bd128e8 corrects the invocation of `runnable_activity_path` removing the sequence parameter (see discussion...) and adds a small view partial test for the happy path rendering.

There is already a reasonable [`runnable_activity_path` helper test](https://github.com/concord-consortium/lara/blob/master/spec/helpers/lightweight_activity_helper_spec.rb#L11).

I would have liked to extend our capybara test to exercise this, but there are some challenges to that, most notably that the controllers use session variables, and capybara rack driver doesn't seem to save session data.

PR #347, opened 2018-10-25
Aspect Ratio dropdown menu displays aspect ratio description.

knowuh:
Select Option was setting label attribute, but not the child content to the setting description. 

For some reason this works in Chrome. ¯\_(ツ)_/¯

Thanks for catching this @dougmartin


scytacki:
👍

PR #348, opened 2018-10-26
Update `run-ci.sh` script for continuously running rspec.

knowuh:
This commit should fix a problem where Guard was unable to run rspec tests because of confusion interactions of spring  and bundler.

@DALove1025 -- please have a look at this and see if it works in your local development environment.


scytacki:
👍

PR #353, opened 2018-11-15
Fix wrapping of embeddables, let authors wrap the next interactive on the same page

pjanik:
1. Cleanup of `Embeddable#page` implementation (1st commit)
2. Fixed numbering / labeling of the attachable embeddables. Now, I always use index based on the position of embeddable on the given page. Before, in some places (authoring view, after the form is closed) I was using "index_in_activity" which works only for questions/reportable_items only. So, if user tried to wrap the non-reportable item, it was crashing LARA edit page. It was also causing some mismatch between labels. Now, the same method is used in pulldown and authing view.
3. I've added a possibility to wrap the next embeddable on the same page. It was very simple so I went ahead. I intentionally reused existing DB fields, so we don't have to run migrations back and forth if we don't like this idea for some reason (and it works fine). I think it's pretty simple and it makes authoring way easier. When I was setting up the example on staging yesterday, it was pretty hard to find the right question when we had 15 of them + a bunch of plugins that also get listed in the pulldown. I've addes some tests of this feature too.

### DIFF
--- a/app/models/attached_to_embeddable.rb
+++ b/app/models/attached_to_embeddable.rb
@@ -15,6 +15,34 @@ module AttachedToEmbeddable
   NO_EMBEDDABLE_VALUE = "no-embeddable"
   NO_EMBEDDABLE_SELECT = [NO_EMBEDDABLE_LABEL, NO_EMBEDDABLE_VALUE]
 
+  # Model can be attached to the next embeddable present on a given page.
+  # It will be calculated dynamically when needed (see embeddable method below).
+  # That way, an author can use embeddable item positioning to wrap other embeddables / attach them to each other.
+  NEXT_EMBEDDABLE_LABEL = I18n.t('NEXT_EMBEDDABLE')
+  NEXT_EMBEDDABLE_VALUE = "next-embeddable"
+  NEXT_EMBEDDABLE_SELECT = [NEXT_EMBEDDABLE_LABEL, NEXT_EMBEDDABLE_VALUE]
+
+  def embeddable
+    if attached_to_next_interactive
+      embeddables = page.embeddables
+      self_index = embeddables.index(self)
+      embeddables[self_index + 1]
+    else
+      super
+    end
+  end
+
+  def attached_to_next_interactive
+    embeddable_type == NEXT_EMBEDDABLE_VALUE
+  end
+
+  def attached_to_embeddable
+    # Note that it's not enough to check embeddable value. If this item is the last one in the page and is attached
+    # to the next interactive, #embeddable will return nil. However, we still should consider this item as attached
+    # to something (e.g. in authoring forms).
+    attached_to_next_interactive || !!embeddable
+  end
+
   def possible_embeddables
     # Do not let embeddable attach to itself
     page ? page.embeddables.select { |e| e != self } : []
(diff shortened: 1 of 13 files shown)


app/models/attached_to_embeddable.rb
```diff
@@ -15,40 +15,88 @@ def self.included(clazz)
...
+  # Model can be attached to the next embeddable present on a given page.
+  # It will be calculated dynamically when needed (see embeddable method below).
+  # That way, an author can use embeddable item positioning to wrap other embeddables / attach them to each other.
+  NEXT_EMBEDDABLE_LABEL = I18n.t('NEXT_EMBEDDABLE')
+  NEXT_EMBEDDABLE_VALUE = "next-embeddable"
+  NEXT_EMBEDDABLE_SELECT = [NEXT_EMBEDDABLE_LABEL, NEXT_EMBEDDABLE_VALUE]
+
+  def embeddable
+    if attached_to_next_interactive
+      embeddables = page.embeddables
```
scytacki: Can you think of a way to do this without loading the actual embeddable objects?  I'm worried that on long pages this will be slow if it happens a lot. If embeddables is using 'acts_as_list' then it might have a next method that efficiently finds the next one.  But I'd guess the embeddables list is not a simple acts_as_list association.

app/views/interactive_pages/_list_embeddables.html.haml
```diff
@@ -12,8 +12,9 @@
           -# e variable is actually either embeddable (for interactives, plugins, etc.) or embeddable answer
           -# (open response, multiple choice and all the question types). If the latter, get a real question object.
           - question = e.respond_to?(:question) ? e.question : e
-          .question{ class: css_class, id: "question-#{question.embeddable_id}" }
-            .question-container
+          .question{ class: css_class, id: question.embeddable_dom_id }
+            -#This CSS class is important, as it might be referenced by plugins and passed to a plugin instance.
+            .embeddable-container
```
pjanik: This might feel risky. However, note that both `id: "question-#{question.embeddable_id}"` and `.question-container` labels have been added recently (wrapping plugin work), in this PR: https://github.com/concord-consortium/lara/pull/351

scytacki:
@pjanik Did you check what happens with interactive pinning? How about the responsive layout work, does the wrapping get adjusted along with the interactive?

app/models/attached_to_embeddable.rb
pjanik: That's a good point. I should be able to come up with something more efficient.

pjanik: @scytacki, fixed here:
https://github.com/concord-consortium/lara/pull/353/commits/6d460934faae3d026881ea24c24eb7f26a5ab7af
Example SQL query:
```
PageItem Load (0.8ms)  SELECT `page_items`.* FROM `page_items` WHERE `page_items`.`embeddable_id` = 26 AND `page_items`.`embeddable_type` = 'Embeddable::EmbeddablePlugin' LIMIT 1
  PageItem Load (0.6ms)  SELECT `page_items`.* FROM `page_items` WHERE (`page_items`.`interactive_page_id` = 538 AND position > 14) ORDER BY page_items.position ASC LIMIT 1
  Embeddable::Xhtml Load (0.7ms)  SELECT `embeddable_xhtmls`.* FROM `embeddable_xhtmls` WHERE `embeddable_xhtmls`.`id` = 16 LIMIT 1
```
So, it's loading page_item, then the next one, and finally the next embeddable. Seems reasonable.

PR #356, opened 2018-11-27
Precipitating change theme

emcelroy:


scytacki:
I was a little surprised at the number of changes in the theme.  We try to keep those minimal so there is less likely to be a problem if we make a change to the code and we don't check that it works with every theme.  But I think it is best to merge this and get it on staging so we can see it. Then we might be able to find a way to reduce the number of customizations required for this theme.

emcelroy:
@scytacki Maybe it would've been better to put at least some of the changes into the mw-runtime-base theme file? There are some funky issues in the navigation bar I was trying to alleviate in particular. There were also some pretty minor changes I made to bring things a little more inline with Concord's new color palette and general look and feel. 

I was reluctant to make changes that could affect other themes that use mw-runtime-base, but maybe that would be OK as long as they're relatively minor tweaks?

scytacki:
There are only 3 themes using mw-runtime-base: data-games, MW, and now precipitating change.
I don't think anyone cares about the MW.  And I think you worked with Frieda to make data-games, so if you see general ways to improve the mw-runtime-base I think it would fine to do. We just need to test data-games.

PR #360, opened 2018-12-03
Configure GA using  ENV vars.

knowuh:
Instead of hard-coding GA information in the view templates, use the same technique and ENV vars used by portal: GOOGLE_ANALYTICS_ACCOUNT

NOTE: A new GA account was setup for "Development Sites"
so that we can use a GA token for staging servers.

[#161965279]

https://www.pivotaltracker.com/story/show/161965279


scytacki:
👍

PR #361, opened 2018-12-05
Precipitating change theme

emcelroy:
This should fix the problem noted here: https://www.pivotaltracker.com/story/show/162394330/comments/197253712

### DIFF
--- a/app/assets/stylesheets/theme-precipitating-change.scss
+++ b/app/assets/stylesheets/theme-precipitating-change.scss
@@ -137,7 +137,7 @@ h3 {
               display: inline-block;
               padding: 0;
 
-              &.am-report, &.am-close { display: none; }
+              &.am-home, &.am-report, &.am-close { display: none; }
 
               a {
                 color: #007c8b;
--- a/app/assets/stylesheets/theme-precipitating-change.scss
+++ b/app/assets/stylesheets/theme-precipitating-change.scss
@@ -91,15 +91,15 @@ h3 {
     }
   } // end .logo-r
 
-@media print {
-  header:first-child {
-    -webkit-print-color-adjust: exact;
-    color-adjust: exact;
+  @media print {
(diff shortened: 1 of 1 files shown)


scytacki:
This change is kind of redundant.  The home button is already hidden, it is just that the styles in this theme are more specific than the existing style that is hiding the button.  It would be safer to find a way so that existing style is not overridden.  Otherwise if we add another button and try to hide it globally it will start showing up in precipitating change.

image

scytacki:
oops I realize my screenshot above is not right.  But I still think this what is happening because of the pre-existing css that is supposed to hide `am-home`

scytacki:
Doh. Strike that, the screenshot is correct above. It is just that I had the `span` selected instead of the li.   The place where this existing style is here: https://github.com/concord-consortium/lara/blob/08782828b3d0c723231964ec0ad2e1ad823311c3/app/assets/stylesheets/activity-menu.scss#L44

emcelroy:
@scytacki: OK, thanks. I've added another commit that makes the style rules in the Precipitating Change theme less specific. That should fix the issue.

That specificity was a result of me trying to use nesting in SASS to follow the structure of nested elements in the HTML. This fix makes the SASS a little more of a jumble. I'm sure there's a way that could be mitigated, but I don't have time to put a lot of thought into this right now unfortunately. Hopefully I can revisit it someday. If you think it's a significant problem though, let me know.

scytacki:
I think this is fine. 

If you do come back to it, I think it would be best if you can look at the data-games and MW theme and see if they would benefit from many of these changes. If so, then moving these styles in to the `mw-runtime-base` file (or higher if it makes sense), would be best.

If we have this many customizations in each theme in LARA it will make testing LARA really hard. The tester will need to test every feature in every theme. So it would be best if we could move most of these changes into 'parent' files.

PR #362, opened 2018-12-11
Added Geniventure theme.

emcelroy:


scytacki:
Looks good.

I do wonder if some of these customizations could be rolled into the `theme` mixin. It could take additional color parameters. The tricky part is trying to figure out sematic labels for how these colors are used.  So instead of saying really generically "theme color a", or really specifically "this is the color for the selected background of the pagination links", we'd can say this is the color for selected buttons.

But in any case I think it is fine to get this merged and up on production and then hopefully soon we can look at this as well as your changes for the PC theme and see if we can consolidate things before we end up with lots theme specific customizations.

PR #367, opened 2019-01-07
Connectedbio theme

emcelroy:


scytacki:
Looks good.

Can you set aside some time to review the 18 'cc-runtime' themes and see if there is a way you can unify their style definitions somewhat?  There are some overlaps. It seems like if you pass more variables to the `_cc-theme-mixin` mixin, then that could reduce the amount of code in the individual theme files.

emcelroy:
Sure. I'll try to get to that some time soon.

PR #371, opened 2019-01-11
Theme style consolidation

emcelroy:
@scytacki: Regarding consolidating theme styles (https://github.com/concord-consortium/lara/pull/367#issuecomment-452785039), does this look like the right path?

For style rules in a theme that could be applied universally, I just remove them from the themes and update related rules in the base SASS files.

For rules that should remain optional per theme, I add params to the theme mixin if it's likely something multiple themes would want as an option.

These are a bit sketchy, but the first commit is an example of #1. The second commit is an example of #2.

I'll aim to do more of 1 than 2, and will leave some custom rules in the theme files whenever it seems like adding yet another param to the mixin would be of limited use.


app/assets/stylesheets/theme-geniventure.scss
```diff
@@ -1,92 +1,55 @@
...
+$col-dim-gray: #4c4a42;
+$col-sea-green: #a7be99;
+$col-dark-sea-green: #7d8d75;
+$col-orange: #f88c22;
+
+$theme-params: (
+  bgColorOrUrl: url('/assets/geniventure-bg/mission-control.jpg'),
+  theme-color-a: $col-dark-olive-green,
+  theme-color-b: $col-orange,
+  activity-menu-display: none
```
scytacki: Have you checked if this activity-menu-display option is working in this case?
My guess is that the 'splat' approach of `@include theme($theme-params...);` just takes the theme-params turns it into an array and then uses each value as a parameter. So my guess is it the activity-menu-display ends up in the right-logo-display parameter.

scytacki:
This looks good to me. @emcelroy do you think this is ready to merge in, or are you doing more work on it?  If it is done can you remove the 'WIP' from the title of the PR.

app/assets/stylesheets/theme-geniventure.scss
emcelroy: Yeah, I checked. I've just double-checked to be sure. It works correctly.

As I understand it, and as it appears to function, so long as you specify the right parameter name, the mixin will assign the corresponding value to that parameter. The order doesn't matter and you can leave out these new parameters without it breaking things. It seems pretty flexible and I thought it'd be a good method to use so you don't have to follow a set order and include values for things you're not concerned with, especially if we keep adding more parameters. Does that make sense?

emcelroy:
Yeah, I think this is done for now and ready for merging. 

I did try to work on making the navbar colors customizable without requiring so many custom style rules in the individual theme files, but I don't think it's going to be easy and it's not something I could work out this week or next.

PR #376, opened 2019-01-18
Add portal endpoints to Lara Plugin context

knowuh:
These URLS are required for the plugin to authenticate with firebase

[#163325279] -- Add Class Info URL
[#163325255] -- Add firebase JWT Fetching URL

https://www.pivotaltracker.com/story/show/163325255

I was looking for clever ways to put this into the LARA API instead of the initialization context, but the Plugin API doesn't know anything about the specific plugin, its configuration, or the run, it would have been problematic.  I documented and tested the additional the Plugin Context values

@scytacki or @dougmartin  could you have a look?

### DIFF
--- a/app/assets/javascripts/plugins.js
+++ b/app/assets/javascripts/plugins.js
@@ -33,13 +33,35 @@ window.Plugins = {
     loadPath: string;
   }
 
+  Interface IEmbeddableContext {
+    // TBD, can be almost anything, details of the wrapped embeddable, eg:
+    // Serialized form of the embeddable, eg:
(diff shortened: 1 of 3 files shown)


knowuh:
I will need to make another commit to provide the plugins a way to fetch learner state from the interactive.

scytacki:
@knowuh and @dougmartin is there a debugging or testing plugin that can be used to verify these changes are working?  I think I remember there was, if so can you add it to the PT stories above so it is easy to test this after it is deployed.

PR #377, opened 2019-01-25
Added click to play dom element id to plugin parameters if click to play enabled [#161772107]

dougmartin:


scytacki:
I think this is a fine way to start, but I would like to abstract it a bit more so plugins do the least amount of direct dom work as possible and instead use javascript apis provided by LARA.

Could you put the new field in an experimental parent in the API?  So the name of the new API field would be something like `experimental.clickToPlayId`

PR #380, opened 2019-02-25
Add custom layout widths to accommodate interactives

emcelroy:


scytacki:
Lets give this a try. But probably we need to make some other changes to LARA to accommodate these requirements better since changing themes is not so easy to get released.

emcelroy:
PT story: https://www.pivotaltracker.com/story/show/164286193

PR #381, opened 2019-02-28
Added check for over quota in question rater [#164118854]

dougmartin:
NOTE: The `ARG_BLOCK.RESUBMIT_ANSWER_ERROR` was removed as it didn't really make sense for any of the error cases as it said to resubmit.

### DIFF
--- a/app/models/c_rater/feedback_functionality.rb
+++ b/app/models/c_rater/feedback_functionality.rb
@@ -23,8 +23,9 @@ module CRater::FeedbackFunctionality
       client_id:    ENV['C_RATER_CLIENT_ID'],
       username:     ENV['C_RATER_USERNAME'],
       password:     ENV['C_RATER_PASSWORD'],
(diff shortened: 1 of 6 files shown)


knowuh:
@dougmartin  the PR looks good.  We should see what is going on with the CI failures though.

dougmartin:
@knowuh Build failures are fixed.  One was a error message string not matching (I had edited it after creating the test) and the others were due to not passing the new api_key parameter in the crater config in the tests.

scytacki:
@dougmartin if we merge will LARA still be able to work with the old CRater? 
It looks like the API key is required.  If that key is set to something bogus will the old CRater still accept the request?

knowuh:
@scytacki -- I updated this PR to address our concern about sending 'x-api-key' headers to C-Rater which might not accept them.  

Also fixes a case where `#c_rater_enabled?` was returning false (even for c-rater) when the `api_key` wasn't set.

PR #384, opened 2019-03-06
Guard: Don't run all tests by default

knowuh:
This is for developers convenience.

Instead of running the entire test suite, manually re-run suite when needed.

We were getting bogged down doing TDD because every time  a focus test passed, the entire suite would re-run.

### DIFF
--- a/Guardfile
+++ b/Guardfile
@@ -3,7 +3,7 @@
 
 guard :rspec,
   cmd: 'bundle exec spring rspec --tag ~slow --tag ~js',
-  # To run all tets, just hit "<return>" in the console.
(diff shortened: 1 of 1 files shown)


Guardfile
```diff
@@ -3,8 +3,9 @@
 
 guard :rspec,
   cmd: 'bundle exec spring rspec --tag ~slow --tag ~js',
-  all_after_pass: true,
-  all_on_start: true do
+  # To run all tets, just hit "<return>" in the console.
```
scytacki:
```suggestion
  # To run all tests, just hit "<return>" in the console.
```

knowuh:
Typo Fixed!  :)

PR #386, opened 2019-03-08
Copy all feedback_item propterties when copying.

knowuh:
fixes 🐞 [#164446076]

> C-RATER: Show student feedback, even when using cached feedback value

We weren't copying `feedback_text` which had the
side-effect of preventing the client-side javascript
from showing feedback for unchanged items.

see this [pt story](https://www.pivotaltracker.com/story/show/164446076)

### DIFF
--- a/app/models/c_rater/feedback_functionality.rb
+++ b/app/models/c_rater/feedback_functionality.rb
@@ -53,15 +53,14 @@ module CRater::FeedbackFunctionality
   end
 
   def copy_of_feedback(prev_feedback)
-    feedback_item = CRater::FeedbackItem.new({
-    status: prev_feedback.status,
-    answer: self,
-    answer_text: self.answer_text,
-    score: prev_feedback.score,
-    feedback_text: prev_feedback.feedback_text,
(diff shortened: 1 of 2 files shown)


knowuh:
@dougmartin -- @scytacki is reviewing, but would appreciate your eyes too.

app/models/c_rater/feedback_functionality.rb
```diff
@@ -57,7 +57,9 @@ def copy_of_feedback(prev_feedback)
     status: prev_feedback.status,
     answer: self,
     answer_text: self.answer_text,
-    score: prev_feedback.score
+    score: prev_feedback.score,
+    feedback_text: prev_feedback.feedback_text,
+    response_info: prev_feedback.response_info
```
scytacki: It seems like the indentation is off here. 
Also you could use `create!` instead of `new` followed `save!`

knowuh: @scytacki  we followed your advice in 4e886b2

PR #387, opened 2019-03-12
164441051 crater resubmit limit 4

DALove1025:
[NP/DL] This PR addresses the PT story [Resubmit limit](https://www.pivotaltracker.com/story/show/164441051) that will limit a the number of times the student may send auto-scoring requests.

Note: Test coverage of this work is deferred until we can import activities with arg-blocks -- which currently appears to be broken. See [LARA Import code skips Arg Block sections](https://www.pivotaltracker.com/story/show/164584021).

Several things are worth noting about this code:

* The manifest constant MAX_ATTEMPTS was changed to MAX_ERROR_RETRIES so to reduce the conflict with the concept of MAX_STUDENT_SUBMISSIONS.

* A completely new activity will have the submit button labeled "Submit"; but once the student has submitted something, the button will be relabeled with "Resubmit" and the number of retries remaining to the student, from that point on, even when the page is refreshed and the reset count is reinitialized to 4.

* To avoid conflicting messages, from the perspective of the student, the dirty-message feedback has been adjusted to say things have been changed, but also informs the student that they are out of tries. Originally, it would have said something like "stuff changed; resubmit" and that seems confusing if you're out of resubmit tries.

* And finally, as a little fit-n-finish touch, the mouse cursor is changed to a "not-allowed" pointer when over the disabled submit button.

### DIFF
--- a/app/assets/javascripts/c_rater.coffee
+++ b/app/assets/javascripts/c_rater.coffee
@@ -10,7 +10,7 @@ class ArgumentationBlockController
   FEEDBACK_HEADER_SEL = '.ab-feedback-header'
   SUBMISSION_COUNT_SEL = '.ab-submission-count'
   SUBMIT_BTN_PROMPT = '.ab-submit-prompt'
-  MAX_ATTEMPTS = 3
+  MAX_ERROR_RETRIES = 3
 
   constructor: (argBlockElement) ->
     @$element = $(argBlockElement)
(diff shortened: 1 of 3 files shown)


knowuh:
@scytacki -- see 8fc419e

* Don't block forward navigation if resubmit limit is hit.
* Change language (text) for the message about being out of feedback attempts.
* Only increment student submission count if the feedback request was successful.

PR #388, opened 2019-03-13
Connectedbio theme

emcelroy:


scytacki:
@emcelroy is there a PT story for these updates?

emcelroy:
@scytacki: No, but there is now.

https://www.pivotaltracker.com/story/show/164674877

I was waffling on whether to actually complete this PR. Seems like the change could be applied to all the themes and that would be the better approach, but I'm not sure and looking into it and asking all the involved parties will likely take more time than I have right now. If it's easy enough to merge this in for now, please do.

PR #392, opened 2019-03-28
Fix: external-scripts block plugin editing.

knowuh:
🐞 FIX for this exception:

```
  An ActionView::Template::Error occurred in embeddable_plugins#edit:
  undefined method `is_hidden?' for #<Embeddable::ExternalScript:0x00000004573178>
```

Triggered when an author tries to edit a plugin on a page with an `Embeddable::ExternalScript`

These are the old style plugins such as the model feedback plugin.

They implemented `is_hidden` instead of `is_hidden?`.

New plugin edit pages need to find a list of all embeddables in the page that aren't hidden.  It sends `is_hidden?` to each embeddable.
This caused the above exception.

The fix was to add the method `is_hidden?` to `Embeddable::ExternalScript`

[#164970025]

https://www.pivotaltracker.com/story/show/164970025

### DIFF
--- a/app/models/embeddable/external_script.rb
+++ b/app/models/embeddable/external_script.rb
@@ -57,6 +57,10 @@ module Embeddable
       false
     end
 
+    def is_hidden?
+      is_hidden
+    end
+
     def export
       return self.as_json(only:[:name, :url, :configuration, :description])
     end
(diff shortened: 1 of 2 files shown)


scytacki:
This is failing un related tests, so it seems like failure might just be random.

PR #395, opened 2019-03-29
hack fix for debian repository change.

scytacki:
Correct fix is to switch to debian version 9, and update the docker-rails-base-ruby image

[#164992701]


scytacki:
The change was taken from here: https://unix.stackexchange.com/questions/508724/failed-to-fetch-jessie-backports-repository

PR #397, opened 2019-04-05
Move LARA Plugin API to a TypeScript sub-package

pjanik:
This is still a draft, but I think it'd be useful for a few devs to take a look at it and give some feedback sooner than later. I guess @sfentress and @scytacki might be interested (and maybe @knowuh and @dougmartin?).

#### What happened here?

1. `lara-plugin-api` dir has been created. It looks like a reasonably small TypeScript package that uses our Webpack + TS config to build an output.
2. `src/assets/javascript/lara-api.js` (that's an old Plugin API file) has been moved to `lara-plugin-api/src/lara-plugin-api.ts` and reimplemented using TypeScript and interfaces that we used to have in comments. Some errors have been already fixed since we get a real type checking.
3. Note that this API depends on a few other modules that are published by LARA using global `window` namespace. E.g. `Sidebar` or `Plugins`. I could easily handle that using webpack's `externals` option. Note that I could also move these modules to this new sub-repo and we can do that incrementally, one by one.
4. When we build `lara-plugin-api`, the JS output file is also being copied to `src/assets/javascript/lara-plugin-api.js` (see package.json build script). Then, LARA can include that in application.js and runtime.js, so things work as always. API is exposed to `window.LARA` like it used to be because of the webpack configuration (`library: "LARA"` and `libraryTarget: "UMD"`).

That's pretty much everything on the LARA side. What we get here is type checking, interfaces being used by the real code, not living only in comments, and a modern JS environment if we want to write new modules there.

Cons? Each time we update this package, we need to run `npm run build`, so LARA gets new build and commit it to the source code.

##### Solution 1

I've tested it with Glossary and it works fine. Let's say we publish this new package as "@concord-consortium/lara-plugin-api". Now, any plugin can just do:

```
import * as PluginAPI from  "@concord-consortium/lara-plugin-api";
```
And use PluginAPI with hints, type checking, etc. (I made sure that typings are generated).

However, it will only work if the plugin also defines  "@concord-consortium/lara-plugin-api" as an external dependency in webpack (as in the end it's exposed by LARA using global window namespace). Most of the plugis will aready have this section because of the React. In case of Glossary plugin, webpack's external section would look like:

```
    externals: {
      'react': 'React',
      'react-dom': 'ReactDOM',
      '@concord-consortium/lara-plugin-api': 'LARA'
    },
```

So, basically we tell webpack not to include '@concord-consortium/lara-plugin-api' library from node_modules as it'll be available under `window.LARA` in the runtime. In other words, we predent we use some library, but its implemenetation is not provided by node_modules in the end (becasue of webpack configuration).

This approach seems to work fine.

##### Solution 2

I haven't tested it, but I can see another option. '@concord-consortium/lara-plugin-api' package exports only interfaces. That why we wouldn't need to configure webpack's `externals` in plugin. But there are two other cons:

1. Configuration gets simpler, but the code will get a bit more strange and we'd need to provide clear instructions in readme anyway:

   ```
    import { ILaraPluginAPI } from  "@concord-consortium/lara-plugin-api";
    // (...)
    const PluginAPI: ILaraPluginAPI = (window as any).LARA as ILaraPluginAPI;
    ```

2. `ILaraPluginAPI` would need to be explicitely defined. Note that in the previous solutions we don't have to do that, as we simply use "library" and its typings. It happens automatically. Here, we'd need to declare interface first, and then provide implementation.

3. We would need to somehow create two different build products. One for LARA, actual implementation. And second one would be just exported interfaces.

I don't have a strong opinion which option is better. What do you think? It might seem a bit complicated, but I'd say it's all relatively easy if you know and understand externals option in webpack.


knowuh:
I haven't reviewed the code yet, but your approach and proposal seem good to me. 

I like **solution 1** above. To me it's quite intuitive, and matches what we do with externals for React anyway.

pjanik:
I've updated the format of comments to follow TypeDoc spec and also added configuration + scripts. I've tested a few different organizations of modules, output formats and themes. Finally, I've settled on one that you can see now (one central file with all the functions). It should also be possible to split this file into smaller modules, so it doesn't grow too much. I'll move sidebar-related API into separate file and commit. It's being already tested. Also, I've used Markdown rendering so we don't have to host docs somewhere else.

If you run `npm run build`, it'll also generate docs and copy them to ../docs/lara-plugin-api:
https://github.com/concord-consortium/lara/tree/ts-lara-api/docs/lara-plugin-api

These files are checked-in, so other developers can read them without building this subpackage.

scytacki:
This all looks great to me.  I agree with Noah that Solution 1 seems best. 

When ready this PR should update this section of the readme: https://github.com/concord-consortium/lara#external-scripting--new-lara-plugin-api and should include some instructions for developers about steps to locally test changes to the lara-plugin-api folder.

scytacki:
Also one of the key things I want to see in the API docs, is the IPluginContext that is passed to the plugin when it is initialized with during initPlugin. I'm guessing you just moved that into the lara-pugin-api subproject yet, or perhaps it is going to be hard to move that?

pjanik:
There's bunch of docs update and futher refactoring of the lara-plugin-api package. Various parts of the API have been moved to separate files/modules. I also converted old plugins.js and sidebar.js into TypeScrip modules, so it's easier to work with them and document their features. Plus we can avoid delegation from API to these modules and export necessary functions directly.

Last missing step is moving LARA Plugin API Jest tests to this new package. It's pretty much necessary and also, testing should get much easier now (no need to do custom resolving in Jest tests).

pjanik:
@scytacki, @knowuh, @sfentress, this is ready for review now.
Tests have been moved into this package and refactored to reflect new code structure.
Docs are updated too. This version should be still compatible with V2 API.

There's one tricky thing that we can encounter later: `initPlugin` obviously needs to be part of this codebase, but it shouldn't be used by Plugin developers. It's called by LARA itself.
So, I just added TypeDoc `@hidden` decorator, so this fuction is not listed in docs. 

I've published this package as "@concord-consortium/lara-plugin-api@2.0.0-pre.2" and used in Glossary Plugin. Things work as expected, plugin builds fine and has type checking.

scytacki:
Looks great to me!

PR #399, opened 2019-04-09
Remove run key from session [#93155846]

dougmartin:


app/views/sequences/show.html.haml
```diff
@@ -42,16 +42,18 @@
   .activities
     - @sequence.lightweight_activities_sequences.each do |las|
       - activity = las.lightweight_activity
+      - activity_path_params = @sequence_run ? [:run_key => @sequence_run.run_for_activity(activity).key] : []
```
scytacki: What is the case where there isn't a @sequence_run?

app/views/sequences/show.html.haml
```diff
@@ -42,16 +42,18 @@
...
+      - activity_path_link = pass_white_list_params(sequence_activity_path(@sequence, activity, *activity_path_params))
```
scytacki: Is this passing the run_key as a URL param instead of adding to the path of the URL?  We'd want it to be part of the URL path for consistency.

app/views/sequences/show.html.haml
dougmartin: Without that the teacher edition specs broke in this build:

https://travis-ci.org/concord-consortium/lara/builds/517974979

the first failed spec is here:

https://github.com/concord-consortium/lara/blob/master/spec/features/run_complete_sequence_spec.rb#L109

app/views/sequences/show.html.haml
dougmartin: I was worried that updating the route would break any bookmarked urls since they would not have the `/:run_key` suffix in the path.  If you think that isn't a problem I can update the code.

app/views/sequences/show.html.haml
scytacki: So that seems like it would be fixed if the sequence show page does a redirect like the activity show page.

app/views/sequences/show.html.haml
scytacki: The URL without the run_key in the path should still work, but it should redirect to a URL with the run_key in path.

dougmartin:
I'll take a look for a sequence menu.  I didn't see one when I setup my test sequence locally but I probably just missed it in the ui.  I'll add the run_key to the path too and do the redirect.

app/controllers/application_controller.rb
```diff
@@ -141,6 +141,17 @@ def user_id_mismatch
...
 
+  def not_authorized_run?(run)
+    begin
+      authorize! :access, @run
+    rescue
+      user_id_mismatch()
+      render 'runs/unauthorized_run', status: :unauthorized
+      return true
+    end
+    return false
```
scytacki: It would be better to throw an exception here instead of returning. This way callers of it don't need to use the form `return if not_authorized_run?`. However I'm not sure how that will interact with the render call. So it might be necessary to add an exception handler which does the actual rendering of 'runs/unauthorized_run'. 

This is done using a approach like https://github.com/concord-consortium/lara/blob/01046ef43b963c136c86b4a3bcd9eae8f972a769/app/controllers/application_controller.rb#L12-L13  So we'd have a custom exception and a rescue_from which catches that exception.

Perhaps it might even be best if the existing CanCan:AccessDenied handler just checks for a action of 'access' and a subject of 'Run' (or whatever strings or symbols should be used there).  And then does what is in not_authorized_run? is doing.  Then callers just need to do `authorize! :access, @run` .

app/controllers/interactive_pages_controller.rb
```diff
@@ -13,18 +13,16 @@ class InteractivePagesController < ApplicationController
...
-      authorize! :access, @run
-    rescue
-      user_id_mismatch()
-      render 'runs/unauthorized_run'
-      return
+      if @sequence
+        redirect_to sequence_page_with_run_path(@sequence, @activity.id, @page.id, @run_key, request.query_parameters) and return
+      else
+        redirect_to page_with_run_path(@activity.id, @page.id, @run_key, request.query_parameters) and return
+      end
```
scytacki: Since this logic of checking for the existence of `@sequence` is used in several places. It seems like this should be moved into a helper. So the helper is something like:  `page_with_run_path_helper(@sequence, @activity.id, page.id, @run_key, request.query_parameters)` and its code checks if @sequence is set or not.

app/controllers/sequences_controller.rb
```diff
@@ -34,9 +34,14 @@ def show
...
+          return
+        end
         if @sequence_run && @sequence_run.has_been_run
           unless params[:show_index]
-            redirect_to sequence_activity_path(@sequence, @sequence_run.most_recent_activity)
+            activity = @sequence_run.most_recent_activity
+            redirect_to sequence_activity_with_run_path(@sequence, activity, @sequence_run.run_for_activity(activity))
             return
           end
         end
```
scytacki: Is there a valid case where `@sequence_run` can be nil here? If not perhaps we should show an error.

app/controllers/sequences_controller.rb
```diff
@@ -179,8 +184,8 @@ def set_sequence
   end
 
   def find_or_create_sequence_run
-    if sequence_run_id = params['sequence_run']
-      @sequence_run = SequenceRun.find(sequence_run_id)
+    if sequence_run_key = params['sequence_run_key']
+      @sequence_run = SequenceRun.find_by_key(sequence_run_key)
       return @sequence_run if @sequence_run
     end
```
scytacki: Now that we have sequence run keys, the line below this should be removed. If there isn't a current_user then we should create an anonymous sequence_run.  So the code below needs to be changed to do that creation of the anonymous sequence run.

app/helpers/logo_helper.rb
```diff
@@ -20,9 +20,12 @@ def sequence_logo_tag
...
     return nil unless sequence
     logo = sequence.logo
-    my_params = request.query_parameters.merge({id: sequence.id, show_index: true})
-    my_params = my_params.merge({sequence_run: @sequence_run.id}) if @sequence_run
-    url = sequence_path(my_params)
+    my_params = request.query_parameters.merge({show_index: true})
+    if @sequence_run
+      url = sequence_with_sequence_run_key_path(sequence, @sequence_run.key, my_params)
+    else
+      url = sequence_path(sequence, my_params)
```
scytacki: If we add support for anonymous sequence runs. Then I don't think there is a valid case where `@sequence_run` will be nil.

app/controllers/sequences_controller.rb
```diff
@@ -34,9 +34,14 @@ def show
...
       format.html do
+        if @sequence_run && !params[:sequence_run_key]
+          redirect_to sequence_with_sequence_run_key_path(@sequence, @sequence_run.key, params)
+          return
+        end
         if @sequence_run && @sequence_run.has_been_run
           unless params[:show_index]
-            redirect_to sequence_activity_path(@sequence, @sequence_run.most_recent_activity)
+            activity = @sequence_run.most_recent_activity
+            redirect_to sequence_activity_with_run_path(@sequence, activity, @sequence_run.run_for_activity(activity))
```
scytacki: It looks like this will not pass the parameters. So if the `mode=teacher-edition` parameter was set then it would be lost.

app/controllers/lightweight_activities_controller.rb
```diff
@@ -80,20 +99,18 @@ def preview
   def single_page
     authorize! :read, @activity
     if !params[:run_key]
-      redirect_to activity_single_page_with_run_path(@activity, @run_key) and return
+      if @sequence
+        redirect_to sequence_activity_single_page_with_run_path(@sequence, @activity, @run_key) and return
+      else
+        redirect_to activity_single_page_with_run_path(@activity, @run_key) and return
+      end
```
scytacki: This also looks like it would lose the `mode=teacher-edition` url param. The old code looks like it would lose it as well.

app/controllers/lightweight_activities_controller.rb
```diff
@@ -47,7 +57,11 @@ def show # show index
     @run.increment_run_count!
 
     if @activity.layout == LightweightActivity::LAYOUT_SINGLE_PAGE
-      redirect_to activity_single_page_with_run_path(@activity, @run.key) and return
+      if @run.sequence
+        redirect_to sequence_activity_single_page_with_run_path(@run.sequence, @activity, @run.key) and return
+      else
+        redirect_to activity_single_page_with_run_path(@activity, @run.key) and return
+      end
```
scytacki: More lost `mode=teacher-edition`.  Currently teacher edition is not handled by single_page activities. But as long as this code is change it seems it should be supported.  However it might not as simple as just passing the current parameters so it is worth testing this out.

dougmartin:
@scytacki I forgot to ping you that this PR was updated yesterday.

app/controllers/embeddable/embeddable_answers_controller.rb
```diff
@@ -11,11 +11,6 @@ def set_answer
   end
 
   def authorize_run_access
-    begin
-      authorize!(:access, @answer.run)
-    rescue
-      user_id_mismatch()
-      render(nothing: true, status: :unauthorized)
-    end
+    raise_error_if_not_authorized_run(@answer.run)
```
scytacki: In the code replaced this is calling `render(nothing: true, status: :unauthorized)` instead of the template that is ultimately rendered by    `raise_error_if_not_authorized_run`.  I'd guess this is because it is in response to a post request or json request. So if we return html it might cause an un handled error in the javascript code. Perhaps this can be handled better by checking the format of the request and returning different things based on this format.

app/controllers/lightweight_activities_controller.rb
scytacki: In this case we were not passing the request.query_parameters before, but now we are.  I don't think this is is a problem though.

scytacki:
@dougmartin I made several more changes. I'm going to work on this a bit more, but perhaps you can look at it and see if notice anything looking broken.

app/controllers/lightweight_activities_controller.rb
```diff
@@ -103,8 +119,8 @@ def summary
     authorize! :read, @activity
     current_theme
     current_project
-    if !params[:response_key]
-      redirect_to summary_with_response_path(@activity, @session_key) and return
+    if !params[:run_key]
+      redirect_to summary_with_run_path(@activity, @run_key, request.query_parameters) and return
```
scytacki: I've confirmed that the Roadmap folks are not using this approach to generate run keys. I think this code was here to help when a run key was stored with the session or the user. Now that session based run keys are gone, this code makes less sense. And it might even be better to prevent the creation of an anonymous run_key when visiting the summary page.

app/controllers/sequences_controller.rb
```diff
@@ -179,14 +185,16 @@ def set_sequence
   end
 
   def find_or_create_sequence_run
-    if sequence_run_id = params['sequence_run']
-      @sequence_run = SequenceRun.find(sequence_run_id)
+    if sequence_run_key = params['sequence_run_key']
+      @sequence_run = SequenceRun.find_by_key(sequence_run_key)
```
scytacki: if an invalid sequence_run_key is used a 404 should be returned, the code should not continue and create a new sequence run

app/helpers/interactive_page_helper.rb
```diff
@@ -1,13 +1,16 @@
...
   def runnable_activity_page_path(activity, page)
-    path_base = fetch_path_base(activity, @run, page)
-    pass_white_list_params(path_base) if path_base
+    path_base = fetch_path_base(@sequence, activity, @run, page)
+    append_white_list_params(path_base) if path_base
   end
 
-  def fetch_path_base(activity, run, page)  # move me to be private.
+  def fetch_path_base(sequence, activity, run, page)  # move me to be private.
     run = run_for_activity(activity, run)
```
scytacki: What is this run_for_activity doing here? We already have a run, so why do we need to find another one?

app/helpers/interactive_page_helper.rb
```diff
@@ -1,13 +1,16 @@
 module InteractivePageHelper
   def runnable_activity_page_path(activity, page)
-    path_base = fetch_path_base(activity, @run, page)
-    pass_white_list_params(path_base) if path_base
+    path_base = fetch_path_base(@sequence, activity, @run, page)
```
scytacki: in this case we are using the instance variable `@sequence`, but the other arguments are passed through. The other arguments would also be available as instance variables, so we should be consistent.

app/helpers/page_helper.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def redirect_to_page_with_run_path(sequence, activity_id, page_id, run_key, query_parameters = {})
+    redirect_to append_white_list_params get_page_with_run_path(sequence, activity_id, page_id, run_key, query_parameters)
+  end
+
+  def get_page_with_run_path(sequence, activity_id, page_id, run_key, query_parameters = {})
+    if sequence
+      sequence_page_with_run_path(sequence, activity_id, page_id, run_key, query_parameters)
+    else
+      page_with_run_path(activity_id, page_id, run_key, query_parameters)
```
scytacki: Since this is a public helper method we should put the `append_white_list_params` on these calls instead in the `redirect_to_...` method.

app/models/ability.rb
```diff
@@ -66,4 +63,17 @@ def initialize(user)
...
+  private
+
+  def admin_owner_or_anon_run(user, run)
+    if user.admin?
+      true
+    elsif user.new_record?
+      # We're currently anonymous
+      run.user_id.nil?
+    else
+      run.user_id == user.id
```
scytacki: It is worth noting that this is a bit unusual:
Anonymous users can access anonymous runs, but logged in users cannot access anonymous runs.
This is probably a good approach so a logged in user doesn't start working on some anonymous run, but it should be documented here and perhaps in a test if not already.

app/views/interactive_pages/_show_completion.html.haml
```diff
@@ -69,12 +69,12 @@
     }
   end
 
-  report_path  = summary_with_response_path(@page.lightweight_activity, @session_key) if @session_key
+  report_path  = summary_with_run_path(@page.lightweight_activity, @run_key) if @run_key
```
scytacki: Is this going to include the sequence id if we are inside of sequence

app/views/interactive_pages/_show_completion.html.haml
```diff
@@ -69,12 +69,12 @@
...
   page_name    = @page.name || @activity.name + ":#{@page.position}"
 
   if @sequence_run
     # if there is a sequence_run then we need to explicitly indicate it
-    exit_link = sequence_path(@sequence, :show_index => true, :sequence_run => @sequence_run.id)
+    exit_link = sequence_with_sequence_run_key_path(@sequence, @sequence_run.key, :show_index => true)
   elsif @sequence
     # Anonymous users running a sequence will have a sequence but not a sequence run
```
scytacki: This is not the case any more, so probably this code can be remvoed.

app/views/interactive_pages/show.html.haml
```diff
@@ -30,7 +30,7 @@
     .button-center
       - if @page.last_visible? && @page.lightweight_activity.student_report_enabled
         .submit.report
-          %a{ :href => summary_with_response_path(@page.lightweight_activity, @session_key), :class => 'gen-report', :target => 'new' }
+          %a{ :href => summary_with_run_path(@page.lightweight_activity, @run_key), :class => 'gen-report', :target => 'new' }
```
scytacki: Will this have the sequence id in it if this is inside of a sequence?

app/views/lightweight_activities/show.html.haml
```diff
@@ -2,7 +2,7 @@
   = "#{@activity.name}"
 
 /
-  = @session_key
+  = @run_key
```
scytacki: Why is the run key added to the page here? I suspect this might have not be noticed because run_keys were not in the URLs of activity shows pages before.

app/views/shared/_activity_menu.html.haml
```diff
@@ -19,7 +19,7 @@
               About
       - if @activity
         %li.am-report
-          = link_to summary_with_response_path(@activity, @session_key),:title => "Report",:target => "_blank" do
+          = link_to summary_with_run_path(@activity, @run_key),:title => "Report",:target => "_blank" do
```
scytacki: will this include the sequence id if run in a sequence?

app/views/shared/_activity_nav.html.haml
```diff
@@ -18,7 +18,7 @@
                   .activity-list-head.open-close
                     = link_to "#{las.position.to_s}. #{activity.name}", runnable_activity_path(activity, :show_index => true)
                     %span{:class => "activity-icons"}
-                      = link_to "", summary_with_response_path(activity, @session_key), :class => "report-link", :title => "Report", :target => "_blank"
+                      = link_to "", summary_with_run_path(activity, @run_key), :class => "report-link", :title => "Report", :target => "_blank"
```
scytacki: ditto from previous comments

scytacki:
In addition to the comments above here is some remaining work:
- redirect activity show pages that don't have a run_key in the URL so the user sees a consistant URL regardless of how the run was created
- write tests to check the rendering of the not-authorized page, there are hints in the code that this rendering needs to happen after setting the theme, but no tests failed when this didn't happen correctly.
- write a test to verify that a student will correctly be able to run an activity that is added to a sequence even after the student has started the sequence. The code needs to make a new run for this new activity and add it to the associated sequence run. In theory this could even happen if the student is in the middle of running the an existing activity and then the student clicks on the sequence navigation menu.
- check that the sequence navigation menu is properly populated with URLs including sequence and run keys.

app/views/lightweight_activities/show.html.haml
scytacki: Oh well perhaps it would be useful for debugging it doesn't seem to hurt anything.

scytacki:
Just found a new issue, when requesting a page without a run_key rails reports that no route matches.

scytacki:
Another issue: anonymous users that access a page without a run key will always get the first anonymous run for that activity (and sequence if defined).

scytacki:
- [ ] when loading a sequence like /sequences/1 as a logged in user, the url of the redirect includes several extra params that shouldn't be there.

app/models/ability.rb
scytacki: this is now documented with tests

app/helpers/interactive_page_helper.rb
scytacki: This is too tricky to improve right now. It seems like the code that deals with pages, activities, and sequences could be improved to be more consistent and easier to understand, but I don't see a clear way to do it yet.

app/helpers/interactive_page_helper.rb
scytacki: This path might be for a different activity. So this method checks for a sequence_run and then finds the run for this sequence.  This is similar to the `runnable_activity_path` method, so perhaps it they could be combined better in the future.

scytacki:
- [ ] need to test the single_page handling with the shared run key examples, it should basically behave the same way.
- [ ] need to test the portal params handling with the single_page, I suspect these might be passed through, and they shouldn't be.

PR #400, opened 2019-04-12
Plugin API V3 changes

pjanik:
This is the first set of changes to LARA Plugin API, moving towards V3.

### Function changes:
 - `isTeacherEdition` removed (TE plugin can perform this URL check itself).

### Plugin Runtime Context changes: 
 - All the URLs are replaced with helper functions that return Promise or `null` when given data is not available.
 - Multiple `string` properties now return `null` instead of empty string (e.g. user email, various states, etc.). This will be beneficial for plugin authors. Type checking will warn them that given property can be null if they try to use it without checking if it's truthy first. Empty string won't ensure that.
- Renamed some properties (e.g. div -> container, as I don't think we'll always return div, etc.) and tried to clean up naming in general.
- Wrapped embeddable properties now live in `wrappedEmbeddable` sub-object. It's `null` or `IEmbeddableRuntimeContext`.
- I renamed `IRuntimeContext` to `IPluginRuntimeContext`. Mostly becuase now we have `IPluginContext`, `IEmbeddableContext` that are generated by LARA Haml template. And then we have `IPluginRuntimeContext` and `IEmbeddableRuntimeContext` that's generated by LARA Plugin API code and passed down to the plugin constructor. I hope this naming makes sense.

You can see changes to interfaces here:
https://github.com/concord-consortium/lara/tree/plugin-api-v3/docs/lara-plugin-api

I'm opening a draf PR because I'm wondering if we should keep V2 available for some time? Or can just migrate to V3 and update all the plugins?

Also, I want to add more tests of the helper functions that I copied from LARA Sharing plugin.


pjanik:
I think it's ready for review now.

I've improved helper functions that were based on sharing plugin implementation. Now they should handle errors better and I've added tests too. 

> I'm opening a draf PR because I'm wondering if we should keep V2 available for some time? Or can just migrate to V3 and update all the plugins?

To answer my own question - I ensured that both V2 and V3 is available at the moment. How? I included older `lara-plugin-api-v2.js` file from master branch. It publishes API under `window.LARA` namespace. The new one uses `window.LARA_V3` instead.

Plugin can choose which one they want to use. Either by just using global namespace, as they used to do till now, or by using different `@concord-consortium/lara-plugin-api` NPM package version. Note that then plugin needs to setup webpack externals differently. `@concord-consortium/lara-plugin-api` needs to be mapped either to `LARA` or `LARA_V3`. Docs mention only V3. V2 is available only not to break old plugins before we migrate them and so we don't have to sync updates of LARA and plugins.

scytacki:
Is there anyway to make methods show up in a different section on this page: https://github.com/concord-consortium/lara/blob/plugin-api-v3/docs/lara-plugin-api/interfaces/ipluginruntimecontext.md it isn't critical, it just confused me when I was looking for the new methods. I found them in the properties list. On the index page there is a seperate set of 'functions'.

scytacki:
What do you think about moving `saveLearnerPluginState` into `IPluginRuntimeContext`?  It seems like it would fit the pattern for it to be a method on IPluginRuntimeContext so then the pluginId wouldn't need to be passed into it.

scytacki:
We had suggested before that `getReportingUrl` would take an optional parameter which was the promise returned by `getInteractiveState`. This way if a plugin wants to access both the interactive state and the reporting url it can do so without triggering two network requests.

scytacki:
It would be good to add note to the `clickToPlayId` indicating that this is temporary and we plan to replace it in the future. In other words we reserve the right to remove support for it, without bumping up to V4.  

I spent a while looking around for documentation conventions or naming conventions for this kind of thing. The closest thing I could find are vendor prefixes in browser apis.  However recently browsers have moved away from vendor prefixes and instead started using runtime flags.  

So my current preference is to put this info in the documentation. But I'm curious if you have a different idea on how to handle this better.

docs/lara-plugin-api/interfaces/isidebarcontroller.md
```diff
@@ -23,7 +23,7 @@
 
 **● close**: *`function`*
 
-*Defined in [api/sidebar.ts:27](https://github.com/concord-consortium/lara/blob/3aa9451f/lara-plugin-api/src/api/sidebar.ts#L27)*
+*Defined in [api/sidebar.ts:27](https://github.com/concord-consortium/lara/blob/4825314e/lara-plugin-api/src/api/sidebar.ts#L27)*
```
scytacki: Can the documentation generator be configured to render these urls as:
`(../../../lara-plugin-api/src/api/sidbar.tx#L27)`?

I made a test branch to verify that this works:
https://github.com/concord-consortium/lara/blob/relative-md-test/docs/lara-plugin-api/interfaces/iclassinfo.md

Because the of the `../../..` requirement it would require some complex support from the documentation generator.

It'd be nice so documentation rebuilds don't result in changes for all of the generated files.

lara-plugin-api/src/api/plugins.ts
```diff
@@ -128,27 +55,21 @@ export const initPlugin = (label: string, runtimeContext: IRuntimeContext, plugi
  @param pluginId ID of the plugin trying to save data, initially passed to plugin constructor in the context.
  @param state A JSON string representing serialized plugin state.
```
scytacki: It would be useful to put a note about how the plugin scope plays a role here. If the plugin is at the activity level then this state should be shared across all pages. At the page and item level it is just for that page or item.

scytacki:
If I understand this correctly, while we are supporting both V2 and V3 we'll have to check out a different branch of LARA to rebuild the V2 api.  Do have that right?  I think that is fine, I just want to make sure I understand.

pjanik:
@scytacki, it's ready for review again. I think I've addressed all your comments, except ones related to TypeDoc, as it doesn't seem possible or easy. I've resolved merge conflicts too (auto-generated docs). Also, I've found a few more bugs in interfaces and I've fixed them both in V2 and V3.

pjanik:
> So my current preference is to put this info in the documentation. But I'm curious if you have a different idea on how to handle this better. 

Not really, but I randomly used @depreciated tag in TypeDoc comments and here's what RybyMine is showing in LARA Sharing Plugin:

Screenshot 2019-04-17 at 18 20 56

So, it seems to work well.

PR #402, opened 2019-04-25
this makes debugging easier

scytacki:


dougmartin:
Looks like you have a gem incompatibility:

```
byebug-11.0.1 requires ruby version >= 2.3.0, which is incompatible with the
current version, ruby 2.2.6p396
```

scytacki:
That issue was the mismatch between the travis ruby version and the docker ruby version. I fixed that on master and merge it in. Now this is only failing because of the usual flakey jasmine tests.

PR #404, opened 2019-05-03
LARA Plugin API => LARA typescript package, general refactoring, event / pub-sub support (logs, click-to-play)

pjanik:
The rationale is described in this PT story:
https://www.pivotaltracker.com/story/show/165743428

The main problem was that LARA Plugin API couldn't really live in isolation from the rest of the LARA JS code long-term. Various features share and will share some utilities, but some of them need to be accessible only from LARA internals and some should be published via Plugin API. This PR tries to solve that. In general, if you like this idea, then all the new LARA JavaScript code should actually go to `lara-typescript` package. So, we can have a modern JS env in LARA for everything, not only Plugin API.

1. lara-plugin-api dir has been refactored into generic lara-typescript.
2. there's `src/plugin-api` dir which is treated in a special way. TypeDocs are generated from this dir, there are separate README and package.json, as the content of this dir is published to NPM (in fact, only generated typings). It can be done via `npm run publish:plugin-api`.
3. I dropped `LARA_V3` and just used `LARA.PluginAPI` and `LARA.InternalAPI`. That way we avoid conflicts with V2 API that was using only LARA. ~~And plugins in the future won't have to change their configuration while switching from V3 to V4 for example. It'll be enough just to bump the version number in package.json.~~
4. README is updated and shows how to use this API in plugins. Now, I only publish typings. That way we can be 100% sure that plugin won't accidentally include PluginAPI code. There's just no actual implementation in `@concord-consortium/lara-plugin-api` package. But it still provides type checking and hits. I tried that and it works well. It's kinda similar to `@types/<old_libraries>` packages.
5. I implemented a simple event bus / pub-sub, but I decided to expose functions explicitly instead of relying on string literals. That provides automatic documentation, type checking and probably way better discoverability. String-based events would be quickly lost I guess, as we'll have to ensure that they're manually documented.
6. It was pretty easy to implement new click-to-play support, so I did it.

There are two things to do:
- Unit tests are missing. I'll work on them, but I deferred that in case you have some bigger feedback that changes code structure a lot.
- Al the plugins converted to V3 would require a tiny update. We need to bump  `@concord-consortium/lara-plugin-api` version and change webpack externals config from `LARA_V3` to `LARA.PluginAPI`. That's pretty minor change.


scytacki:
I haven't looked at this yet other than the description, but regarding the LARA_V3 -> LARA.PluginAPI change. When we release a non backward compatible LARA plugin V4. Won't we be in the same situation? How will LARA keep the V3 and V4 apis separate? For example let's say we change an existing method name to have a completely different signature.

pjanik:
> When we release a non backward compatible LARA plugin V4. Won't we be in the same situation?

You're right, it will be. Although, we can always add `_V4` suffix then. Or maybe we'll decide that for some reason we want to use a different namespace (like `PluginAPI`) - pretty much like what happened now. Hopefully, we won't have to make V4 changes too soon. Up to you, plugins won't really use this namespace anyway, as long as they use TypeScript and webpack.

scytacki:
I'd vote for `LARA.PluginAPI_V3` this makes it clear if we want to go to V4 in the future.  Since this name will only show up on one line of the config it seems good to keep the version explicit.

lara-typescript/src/plugin-api/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/concord-consortium/lara.git"
+  },
+  "author": "Concord Consortium",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/concord-consortium/lara/issues"
+  },
+  "homepage": "https://github.com/concord-consortium/lara#readme"
```
scytacki: It is probably better to point this at the plugin-api readme instead of the LARA readme. This confused Doug when he was trying to use the package. So in the 'old' code he changed it to: `https://github.com/concord-consortium/lara-plugin-api#readme` . It will be a different in this new code.

lara-typescript/src/plugin-api/types.ts
```diff
@@ -82,6 +82,14 @@ export interface IEmbeddableRuntimeContext {
    interactive status, including click to play state.
    ****************************************************************************/
   clickToPlayId: string | null;
```
scytacki: I think this can be removed from the V3 api. We haven't released a version with it, and now plugins can use onClickToPlayStarted method instead.

lara-typescript/webpack.config.js
```diff
@@ -1,15 +1,18 @@
...
-    entry: './src/lara-plugin-api.ts',
+    entry: './src/index.ts',
     mode: 'development',
     output: {
-      filename: 'lara-plugin-api.js',
-      library: 'LARA_V3',
+      filename: 'lara-typescript.js',
+      // Temporarily disabled, as it conflicts with V2 definition. Instead, src/index.ts file exports library
+      // manually by extending LARA namespace. When lara-plugin-api-V2 is removed, this can be uncommented.
+      // library: 'LARA',
```
scytacki: If we go with LARA.PluginAPI_V3 is this comment still relevant? If the line is un-commented will it mess up plugins that are expecting the api at LARA.PluginAPI_V3?  Sorry I haven't taken the time to figure out this output option in wepack.

pjanik: It is still relevant, unfortunately. The problem is that the old API is available under `LARA` namespace. And this new code still uses top-level `LARA` namespace. Webpack seems to create a new object instead of trying to check if it already exists and try to expand it. So, V2 namespace will be overwritten by this one.

In other words, we would need to change top-level namespace, not just the `PluginAPI` part. But that's a tiny thing, once we do V2 cleanup, this should be 30 seconds of work.

lara-typescript/src/plugin-api/types.ts
pjanik: Okay. I didn't know if I'd have time to actually update LARA Sharing Plugin and I didn't want to block the whole release process.

Actually, now it seems it's even worse. I think LARA Sharing Plugin doesn't use click-to-play at all. Something has been changed and I'm not sure if it's intentional or not. Probably not. The code performs monitoring, but its result is just never used. There are no tests. And the local setup of this whole thing is pretty painful. So, I can definitely remove this id and perhaps we should assume that Sharing Plugin should be fixed (as it's probably broken at the moment).

I can definitely take a look at it, although if Doug or Noah have some time, it might be more efficient for them. They understand this code better and probably they have a bunch of local tests available.

lara-typescript/src/plugin-api/package.json
pjanik: Sure, I'll fix that. 
Readme is correct here:
https://www.npmjs.com/package/@concord-consortium/lara-plugin-api 
NPM uses README.md file that's actually published to NPM, not this link.

pjanik:
@scytacki, all the issues should be fixed now. Also, I've added unit tests of the new and older features. Test coverage:
Screenshot 2019-05-06 at 11 46 45

pjanik:
@scytacki, all the issues should be fixed now. Also, I've added unit tests of the new and older features. Test coverage:
Screenshot 2019-05-06 at 12 01 16

lara-typescript/webpack.config.js
scytacki: OK, I think I understand now. So once this lara-typescript package contains all of the javascript that makes up the `LARA` object and there is no javascript in LARA outside of this package, then we can make the switch.  And this switch will not affect any plugins because the api will still be at LARA.PluginAPI_V3.

pjanik: Exactly.

lara-typescript/src/plugin-api/types.ts
scytacki: OK, that sounds like a good reason to leave this in for the time being since I'm not sure when we'll get to fixing the sharing plugin.

PR #406, opened 2019-05-06
We don't seem to be using therubyracer anymore

scytacki:
I couldn't find anything using therubyracer. And Doug mentioned it made some ruby updates harder, so I'm removing it.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,7 +1,6 @@
 source "https://rubygems.org"
 
 group :production do
-  gem "therubyracer", "~> 0.12"
   gem "unicorn"
 end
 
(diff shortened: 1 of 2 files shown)


scytacki:
I did some more searching, and it seems like some gems we use for asset compiling use execjs which tries to find a javascript engine automatically. So it is possible this PR breaks the assets pre-compile code. I'm testing now...

pjanik:
There's nothing that comes to my mind re this gem, unfortunately. I don't really remember / know why we would need this gem.

scytacki:
The asset compile completed locally without showing an error.  So I'm going to merge this in.

PR #411, opened 2019-05-13
Fix sizing of the interactives

pjanik:
[#165150685]

This PR fixes issues described here:
https://www.pivotaltracker.com/story/show/165150685

In particular, it fixes these two uses cases:
- default aspect ratio + state saving (header)
- use all available space + state saving

Interactive sizing was broken and there's a variable literally named `magicNumber`. It's gone now. I didn't add it, but I believe the calculations below are what the author wanted to achieve there. I've also added rescaling of question header to avoid this:

Screenshot 2019-05-13 at 20 59 35

Note that multiple interactives in the interactive box + some aspect ratio settings + responsive layout won't work well. This PR shouldn't make things worse, but it also doesn't improve that. Just something to keep in mind. I didn't try to fix it, as first, we would need to think how these cases should be handled.

### DIFF
--- a/app/assets/javascripts/interactive-sizing.js
+++ b/app/assets/javascripts/interactive-sizing.js
@@ -5,22 +5,37 @@ function interactiveSizing () {
     var $iframe = $(this);
     var aspectRatio = $iframe.data('aspect-ratio');
     var resizeMethod = $iframe.data('aspect-ratio-method');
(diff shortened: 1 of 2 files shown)


scytacki:
We'll probably want to back-port this, so we can apply it to production.

PR #412, opened 2019-05-13
Fixed onClickToPlayStarted() race condition [#165290560]

dougmartin:
onClickToPlayStarted() now immediately invokes the last non-null click to play event before the handler is added.

This guards against race conditions where the click to play element is clicked before the event handler is added in the plugin.


pjanik:
Going to merge it, but a few notes for future reference:

1. This PR is mostly about refactoring to `interactiveAvailable` name.
2. My comment above is wrong. It does not fix possible race condition actually and Doug just removed some code handling it. But we tried to analyze the code and it doesn't seem to be very likely. The problem is that `interactiveAvailable` property isn't updated. It's generated by HAML template on page rendering and we don't update it if the user starts the interactive in the meantime, before the plugin is loaded.
3. However, both click-to-play and plugins are initialized on `document.ready` event, so possible time between both things doesn't seem to be long enough for the user to actually click and start the interactive. So, I guess it's only a theoretical issue.

PR #413, opened 2019-05-13
A rake task to import activities into report system.

knowuh:
This uses task pushes data in the old export format into FireStore using a cloud function.

Update: This is no longer using any FireStore gem.  It just posts to our custom cloud function with HTTParty.  

To run this in a container just export these variables into the environment:

```
REPORT_SERVICE_TOKEN=super-secret-key
REPORT_SERVICE_URL="https://us-central1-report-service-dev.cloudfunctions.net/api/import_structure"
REPORT_SERVICE_SELF_URL="https://authoring.staging.concord.org"
```

[#165937226]
https://www.pivotaltracker.com/story/show/165937226

### DIFF
--- a/lib/tasks/lightweight_tasks.rake
+++ b/lib/tasks/lightweight_tasks.rake
@@ -10,4 +10,52 @@ namespace :lightweight do
       puts "This script only runs in the development environment."
     end
   end
+
+  desc "Publish All activity structures to FireStore report serivce"
+  task :publish_to_firestore => :environment do
+    cli = HighLine.new
+
+    # eg: 'super-secret-key'
+    token = ENV["REPORT_SERVICE_TOKEN"] || cli.ask("Report bearer token: ")
+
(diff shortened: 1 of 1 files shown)


knowuh:
I am not sure if this should be merged now or not. This task will have to be modified once we have the new publication format.

scytacki:
Have you thought about how to run this on production. We'd need to generate the `report-service-dev.json`file somehow. If it just has a few private parts, perhap we can use a template and the rake task can replace some parts with environment variable values. 

If the private contents are too long to come in through ENV variables, we'll need a new plan.

This file puts a monkey wrench in my idea of using a ECS Task to run this.

knowuh:
@scytacki using this method, it should be much easier to setup an ECS Task to run this.

dougmartin:
I'm going to use this branch off this branch to add the run export rake task.

PR #414, opened 2019-05-14
[SC/DL] Log messages now have w/ embeddabble's id, etc

DALove1025:
[SC/DL] Added code, in support of PT story 165182247 about adding logging for TeacherEdition. This change uses LARA to augment the log messages from a plug-in with 3 additional properties:

* plugin_id,
* embeddable_type, and
* embeddable_id.

In particular, see the `log` function around line 120 of `lara-typescript/src/lib/plugin-runtime-context.ts`. Perhaps this function could be improved with some refactoring.

Also, added a few unit-tests to flesh out the behavior.


lara-typescript/src/plugin-api/types.ts
```diff
@@ -40,6 +40,20 @@ export interface IPluginRuntimeContext {
...
   /** Wrapped embeddable runtime context if plugin is wrapping some embeddable. */
   wrappedEmbeddable: IEmbeddableRuntimeContext | null;
+  /**
+   * Logs event to the CC Log Server. Note that logging must be enabled for a given activity.
+   * Either by setting URL param logging=true or by enabling logging in Portal.
+   * ```
+   * PluginAPI.log("testEvent");
+   * PluginAPI.log({event: "testEvent", event_value: 123});
+   * PluginAPI.log({event: "testEvent", someExtraParam: 123});
+   * PluginAPI.log({event: "testEvent", parameters: { paramInParamsHash: 123 }});
```
pjanik: I guess this example should be updated now, e.g. `context.log(...)` instead of `PluginAPI`.

scytacki:
Yes those changes all make sense. Thanks Piotr.

lara-typescript/src/plugin-api/types.ts
DALove1025: Great comments. Thanks.

pjanik:
Thanks, it looks good now 👍

scytacki:
@pjanik I just realized that removing the PluginAPI.log method, we probably broke the model-feedback plugins.

pjanik:
@scytacki, I've updated model feedback to use new API:
https://github.com/concord-consortium/model-feedback/pull/7

PR #415, opened 2019-05-17
Provide CSS normalization helper, publish it in Plugin API NPM package

pjanik:
[#165859735]

I think new readme explains goal of this PR (see below).
It also setups handling of CSS in lara-typescript package.

Basically, LARA adds set of CSS rules that reset styles in `.normalized-container` elements.
These CSS rules are also separately exported and packaged in LARA Plugin API npm package, so plugins can embed these styles and have consistent rendering in their demo/authoring pages.

Note that this normalization might not be perfect and might require tweaking in the future. I also think that plugins could provide these reset helpers themselves. But I'm adding this, as I alraedy had that ready and I guess it might be nice to have one common place for generic "LARA CSS reset". :wink:

#### Plugin styling

Note that LARA Plugins will be executed in LARA context which includes a lot of different CSS styles and rules.
Sometimes it might be okay for plugin to inherit these styles and overwrite when necessary. In other cases, 
it might be useful for plugin to reset and normalize its containers, so it's less dependant on LARA styles.
Also, LARA styles obviously casue plugin content to be rendered differently in LARA and in some external 
demo or authoring pages provided by plugin.

To avoid these inconsistencies, plugins can use CSS reset/normalize helpers provided by LARA.
Reset CSS rules are simply applied to all the containers with `normalized-container` class name. 
Plugins can use this case in its top-level container, e.g.:

```javascript
render() {
  return (
    <div className="normalized-container">
      My plugin content
    </div>
  );
}
```

Also, these reset styles are part of the NPM package, so plugins can include them in demo and authoring pages 
to ensure that rendering is consistent:

```javascript
import "@concord-consortium/lara-plugin-api/normalize.css";
```


pjanik:
Also, I think we don't need to treat this as a perfect normalized styles. I guess we could add there some custom, opinionated rules that will be simply shared by all the plugins and that can be easily imported by them.

pjanik:
Thanks, I've fixed that.

PR #417, opened 2019-05-22
Create ReportService::RunSender, ReportService::ResourceSender and  reporting rake tasks.

knowuh:
This branch is not to be merged yet.

I put it here because in the run-sending work I refactored the task which sends the structure to the report service. 

It sounded like @pjanik was possibly going to make changes soon to the export format for sequences and activities, so I wanted to be sure he looked at these two files:

`app/services/report_service/resource_sender.rb`  and `lib/tasks/reporting.rake`  

I still have more work to do here with keying users to runs based on our discussion.


knowuh:
@scytacki  One thing to note, I changed the name of an ENV variable from something like `REPORT_SERVICE_SELF_HOST` to `REPORT_SERVICE_SELF_URL`.  You might need to double check that the Cloud Formation Template has the correct value.

pjanik:
@knowuh, here's PR with the new export format: https://github.com/concord-consortium/lara/pull/419

So, once it gets merged, you could replace `export_for_portal` with `export_for_report_service` in your ResourceSender code.

scytacki:
I'm going to merge this as is and @knowuh is going to fix the usage of `stub` in a future PR.

pjanik:
One comment once I started working with this code a bit. 
I'm wondering if we need classes, internal state, and inheritance here. It seems "sender" is just something that takes an input, preprocesses it, and sends somewhere. Perhaps could be implemented using a few simple, stateless functions in one module. Personally, I find them a bit easier to read, reason about and test (but maybe too much JS). Not that I'd change this code now, it's all fine, but just a general comment (very opinionated too :wink:) to consider/discuss in the future when we work on bigger things.

PR #418, opened 2019-05-23
Update plugin metadata [#166193888] [#165887809]

dougmartin:
This PR consists of two stories:

1. Added v3 plugin label override with internal label [#166193888]
2. Plugin metadata for authoring use [#165887809]

The plugin system now ignores the label provided by the plugin when it registers and instead uses an internally generated label set before each plugin is loaded.  This allows multiple copies of the same plugin to exist on a page.


scytacki:
looks good to me so far

scytacki:
This still looks good to me. I think it is waiting on a cypress test to verify the form works.

PR #424, opened 2019-05-29
165217423 batch import class info for runs

knowuh:
Part of the large bug / feature story about adding `class_info_url` and `class_hash`
to runs and sequence runs launched from the portal.

We need to update existing runs in the DB with this information.

[#165217423]
https://www.pivotaltracker.com/story/show/165217423

See also the [Portal PR #659](https://github.com/concord-consortium/rigse/pull/659) that exports the data that this Task imports from.

### DIFF
--- a/app/models/run.rb
+++ b/app/models/run.rb
@@ -4,7 +4,9 @@ class Run < ActiveRecord::Base
   class PortalUpdateIncomplete < StandardError; end
   class InvalidJobState < StandardError; end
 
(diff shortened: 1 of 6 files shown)


pjanik:
This seems to affect only old runs.
How would we add class_hash and class_info_url to new runs? Is that part of another PR?

knowuh:
Yes, sorry the other  PR is taking so long to put together.  Lots of interruptions.  I will put up what I have so we can at least talk about it.

knowuh:
@scytacki when you get a chance.  Or I could put @pjanik on this one.

PR #425, opened 2019-05-30
Added plugin authoring metadata and wrapper UI [#166193534] [#166194332]

dougmartin:
Adds authoring_metadata to approved scripts which currently has list of components scoped to activity, embeddable and embeddable-decoration.  The authoring_metadata field is a stringified JSON blob to allow for change during the remaining plugin authoring stories.  Once the json is finalized a schema validator should be added as part of the code that loads the authoring metadata from the external json manifest.

The activity forms that deal with plugins now select plugins appropriate to the scope and then saves the component label to the plugin so that the component name can be used and displayed instead of the generic plugin name.  This allows for plugins to expose multiple components at different scopes.

One final change is that once a component is added at either the activity or embeddable level there is no way to change the type of component in the edit form.

### DIFF
--- a/app/models/embeddable.rb
+++ b/app/models/embeddable.rb
@@ -7,7 +7,6 @@ module Embeddable
     Embeddable::Labbook        => "Labbook",
     QuestionTracker            => "Tracked Question",
     Embeddable::ExternalScript => "External Script",
-    Embeddable::EmbeddablePlugin => "Plugin"
   }
 
   InteractiveTypes = {
@@ -34,7 +33,7 @@ module Embeddable
   end
 
   def self.valid_type(class_string)
-    return Types.detect{|k,v| k.to_s == class_string }
+    return class_string == Embeddable::EmbeddablePlugin.to_s || Types.detect{|k,v| k.to_s == class_string }
   end
 
   def self.create_for_string(class_string)
--- a/app/models/embeddable.rb
+++ b/app/models/embeddable.rb
@@ -75,6 +75,10 @@ module Embeddable
     true
   end
 
+  def show_in_edit?
+    true
+  end
+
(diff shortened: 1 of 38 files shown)


app/models/embeddable.rb
```diff
@@ -34,7 +33,7 @@ def self.table_name_prefix
   end
 
   def self.valid_type(class_string)
-    return Types.detect{|k,v| k.to_s == class_string }
+    return class_string == Embeddable::EmbeddablePlugin.to_s || Types.detect{|k,v| k.to_s == class_string }
```
knowuh: I am not sure I understand why this was moved from line 10. If line 10 was  changed to:

`Embeddable::EmbeddablePlugin => Embeddable::EmbeddablePlugin.to_s` 

It would be  the same  right?

dougmartin: I wanted to remove the generic plugin option from the embeddable dropdown so I have to remove it from that list.

dougmartin:
I'm changing this to a WIP PR so I can add the wrapped UI story to it as the current code removes the ability to add wrapped plugins.

lara-typescript/src/plugin-api/json/teacher-tips-plugin-manifest.json
```diff
@@ -0,0 +1,29 @@
...
+        "label": "sideTip",
+        "name": "Side Tips",
+        "scope": "embeddable"
+      },
+      {
+        "label": "questionWrapper",
+        "name": "Question Wrapper",
+        "scope": "embeddable-decoration",
+        "decorates": [
+          "Embeddable::MultipleChoice"
```
scytacki: This should also include "Embeddable::OpenResponse", and "MWInteractive"

PR #427, opened 2019-06-03
Add resource_url to answers exported to Firestore, change user_id => user_email

pjanik:
[#166187624]

We need to query answers for a given activity or sequence in the Portal Report app. Current data format was missing a simple way to do it. We could theoretically query for every single question_key, but it doesn't seem too reasonable (we'd have dozens of Firestore observers and queries running in one report).  Also, `resource_url` seems better than `activity_id` or `sequence_id` as we're independent of the resource type and naming.

I've also changed "user_id" to "user_email" as that's what it actually is.

BTW, most of this data that we export for answers and runs won't be useful in reports. Not saying it's a problem, but I'm wondering if we should try to export everything that we can or things that we need.

### DIFF
--- a/app/services/report_service/run_sender.rb
+++ b/app/services/report_service/run_sender.rb
@@ -2,7 +2,7 @@ module ReportService
   class RunSender
     Version = "1"
     DefaultClassHash = "anonymous-run"
-    DefaultUserName = "anonymous"
+    DefaultUserEmail = "anonymous"
 
     def question_key(answer_hash)
       question_type = answer_hash[:type] || answer_hash['type']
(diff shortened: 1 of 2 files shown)


scytacki:
I talked with Piotr about this a bit this morning. I'm going to merge this in as is. But we'll likely want to add some LTI fields of platform_id, resource_link_id. The platform_id is like our domain field. And the resource_link_id is like the offering_id in our portal. It would be better to filter answers based on this offering_id since in the future the same class might have the same resource assigned multiple times. By adding the resource_link_id (offering id) We might be able to drop the resource_url but it kind of seems like a useful field to have on an answer even if we don't totally need it.

In general I'd say we should pair down what we send to the report service. We want the report service to be something that works with more than just LARA resources and answers. And this export from LARA is kind of a specification of what the report service needs, so it is useful to reduce it if we can.

pjanik:
> And this export from LARA is kind of a specification of what the report service needs, so it is useful to reduce it if we can.

I'd be cautious about that. LARA exports too much data and attributes. Not that it hurts, but if we think about specification and docs, it can be problematic. Other systems might not have the same concepts and features, so it can be difficult to provide all these attribute values.

Portal Report is trying to do that the other way. Its dummy data provides a subset of LARA export format - fields that we actually need to be able to provide a functional report. So, if we work on docs, it might be useful to look at these files:
https://github.com/concord-consortium/portal-report/tree/master/js/data

PR #428, opened 2019-06-03
Set Run info from portal launch parameters

knowuh:
When launched from the portal, the `class_info_url` `class_hash` will be set on Runs and Sequence runs.

This addresses a bug where `class_info_url` was not being set for sequence runs.
Also adds a `class_hash` property to the runs & sequence runs if provided in launch params.

Will not update existing run parameters if they already exist on runs.

This PR is based off of #424 and should be reviewed after it.

See also portal pull request: 
[#165217423]
https://www.pivotaltracker.com/story/show/165217423

### DIFF
--- a/app/models/with_class_info.rb
+++ b/app/models/with_class_info.rb
@@ -1,0 +1,28 @@
+module WithClassInfo
+
+  # If we have new class_info_url or class_hash in our query params
+  # set our class_hash and class_info URLs to that (if they are blank)
+  # Discussion: Not setting these values if they are already set seems safest.
+  # Returns changed attributes, or false if nothing was changed.
+  def update_class_info(raw_params)
+    params = raw_params.stringify_keys
+    new_class_info_url = params['class_info_url']
+    new_class_hash = params['class_hash']
+    update_attributes = {}
+    changed = false
+    if new_class_info_url && (self.class_info_url.blank?)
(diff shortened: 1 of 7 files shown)


app/models/with_class_info.rb
```diff
@@ -0,0 +1,28 @@
+module WithClassInfo
+
+  # If we have new class_info_url or class_hash in our query params
+  # set our class_hash and class_info URLs to that (if they are blank)
+  # Discussion: Not setting these values if they are already set seems safest.
```
scytacki: There is the case where a student moves from one class to another. We just added support for admins to do that in the Portal: https://github.com/concord-consortium/rigse/pull/653 
This code reuses the same Learner model so it has the same remote_endpoint, which means the same LARA Run or SequenceRun will be found.  

By resetting the class params on each launch, then they will be correct at the time a material is re-launched. However this approach is not going to help the report service which storing this info to reconstruct the report.

So it is probably better to have an explicit 'update' operation which can notify any external systems when something is changed and not need to relaunch.  There doesn't seem to be built in LTI support for this, but it could be a feature of our external Tool as long as we can come up with a way for a user to tell us what the two context_ids and resource_link_ids are.

There is a PT story for adding support to the report service for handling moving students from class to class: https://www.pivotaltracker.com/story/show/166217660 we'll have to deal with this issue when we work on that story.

PR #429, opened 2019-06-06
Define explicit `report_service_hash` methods for questions and answers

pjanik:
We won't heave to deal with inconsistencies present in portal_hash.
Also, I've removed `key` and `question_key` props. Now, I'm assuming that all the `id` should be unique among questions and answers. It could make sense for other types (activities, pages, sequences) if we think about providing a fully flexible activity structure in the future.


knowuh:
@pjanik I made two minor tweaks to spec tests.

PR #430, opened 2019-06-06
Don't send non-changed answers to the report service in `reporting.rake`

knowuh:
As per @pjanik 's suggestion, we only send answers who's `updated_at` is newer than `created_at`.

[#166524647]
https://www.pivotaltracker.com/story/show/166524647

### DIFF
--- a/app/services/report_service/run_sender.rb
+++ b/app/services/report_service/run_sender.rb
@@ -31,9 +31,9 @@ module ReportService
     end
 
     def serlialized_answers(run, host)
-      age_threashold_seconds = 0.25
+      age_threshold_seconds = 0.25
       modified_answers = run.answers.select do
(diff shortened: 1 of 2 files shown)


app/services/report_service/run_sender.rb
```diff
@@ -36,8 +36,11 @@ def serialized_answer(ans, run, host)
     end
 
     def serlialized_answers(run, host)
-      run.answers
-      .map do |ans|
+      age_threashold_seconds = 0.25
```
kswenson: "threashold" => threshold

knowuh:
@pjanik : The runs and answers actually have a 'dirty' flag that they maintain.  We can use that same approach when we incrementally update student  answers, but it wouldn't work for a batch operation.

PR #433, opened 2019-06-11
Export required and submitted props of the questions and answers

pjanik:
[#166414256]

1. Questions can be `required`.
2. Answers can be `submitted`.


scytacki:
Unfortunately I couldn't get 1.38.0 on production because I found a bug on staging. So I don't want to merge this, or update staging with it until we get 1.38.0 fixed.

PR #434, opened 2019-06-12
multiple plugins of the same type on the same page were broken

scytacki:
the plugin.id is different for each place the plugin is used.
instead we want an id that is constant for each javascript file that is loaded

[#166627102]


scytacki:
Just realized there is another potential problem. If someone registers the same plugin URL in two approved scripts the second one will not work properly because it will never be registered under the second approved_script id.

Probably we should use a hash of the url instead of the approved_script id. Or make sure there can't be two approved_scripts with the same url.  But that can be done later.

PR #436, opened 2019-06-14
Report service data/ids refactoring

pjanik:
Related to:
https://github.com/concord-consortium/rigse/pull/665

- Rename all the “class_hash” attributes to “context_id” (e.g. in Runs).
- Add “platform_id”, “platform_user_id” and “resource_link_id” attributes to Run.
- Update “import_clazz_info” rake task (perhaps it should be renamed too) to set “resource_link_id” for every run using “offering_id” attribute from the CSV file.
- Update “import_clazz_info” rake task to set “platform_id” and “platform_user_id” fields in every run based on “user_id” from the CSV file and hostname (manually provided when rake task is started).
- Update LARA code to save “domain_id”, “domain_uid”, and “resource_link_id” provided during the launch of the activity, as “platform_id”, “platform_user_id”, “resource_link_id”  in the newly created Run object.
- Update LARA report-service export code to include “platform_id”, “platform_user_id” and “resource_link_id” attributes in every answer.
- Update LARA report-service export code to include “tool_id”, “tool_user_id” in every answer.


app/services/report_service.rb
```diff
@@ -20,7 +20,7 @@ def self.make_key(*values)
   end
 
   def self.make_source_key(url)
-    make_key(url.gsub(UrlRegex, '\1'))
+    url.gsub(UrlRegex, '\1')
```
scytacki: Does the cloud function replace the dots with underscores when it makes the top level source document name?  Or is this going to change the top level source document names?

Either approach seems fine, I'm just curious.

lib/tasks/reporting.rake
```diff
@@ -110,34 +109,33 @@ namespace :reporting do
         puts "    Updated #{updated_srun_count} SequenceRuns"
         puts "    Updated #{updated_run_count} Runs"
       end
-      clazz_id, class_hash, learner_id, learner_key = import_line.strip.split(",")
+      clazz_id, class_hash, user_id, learner_key, resource_link_id = import_line.strip.split(",")
```
scytacki: This was refactored by Noah, so we should merge in master and resolve merge conflicts. The new code uses the learner_id when the learner_key is missing.

app/services/report_service.rb
scytacki: based on other code it seems likely the keys now have dots.

lib/tasks/reporting.rake
scytacki: I think I just fixed this. I'll wait for it to tests to run now.

lib/class_info_import_helper.rb
```diff
@@ -18,33 +18,38 @@ def self.class_info_url_path(class_id)
...
+      context_id: class_hash,
+      platform_id: platform_id,
+      platform_user_id: user_id,
+      resource_link_id: resource_link_id
+    }
+    SequenceRun
+      .where(remote_endpoint: remote_endpoint)
+      .each do |srun|
+        srun.update_attributes(attrs)
+        updated_srun_count = updated_srun_count + 1
```
scytacki:
```suggestion
```
Missed this reference to srun which won't work anymore.

lib/class_info_import_helper.rb
```diff
@@ -18,33 +18,37 @@ def self.class_info_url_path(class_id)
...
-          class_info_url: info_url,
-          class_hash: class_hash
-        })
+    Run
+      .where(remote_endpoint: remote_endpoint)
+      .each do |run|
+        run.update_attributes(attrs)
         updated_run_count = updated_run_count + 1
       end
+    line_count = line_count + 1
```
scytacki:
```suggestion
```

app/services/report_service.rb
pjanik: Sorry for the late answer, I didn't check this email, but I saw on Slack that this PR was merged.
Yes, I changed underscores to dots, so we can assume that the source is a host. That makes the code in the client-side app a bit simpler (no magic transformations required to get a source name from URL).

pjanik:
Thanks for the fixes!

PR #437, opened 2019-06-17
only send Portal Runs to the report service

scytacki:
this is is much fewer runs at the time of the commit this was 3,000 runs instead 800,000 runs.

We can currently only use the report with portal runs, so this isn't breaking anything.
When we support report service reports on teacher previews and anonymous runs, then we'll need to change this.


scytacki:
Yeah I thought we would also want all runs to be sent too, but we haven't put any work into using this report instead of LARA own report.  That LARA report is the only way to view a report of these non portal runs.

PR #439, opened 2019-06-17
In `#import_clazz_info` of `reporting.rake` fall back to `learner_id`

knowuh:
When `learner_key` isn't present, use `learner_id` to figure out `remote_endpoint`

When importing `class_hash` values from from portal CSV, if the learner secret key is missing, fall back to the `learner_id`.

[#166747707]

https://www.pivotaltracker.com/story/show/166747707

### DIFF
--- a/lib/class_info_import_helper.rb
+++ b/lib/class_info_import_helper.rb
@@ -1,0 +1,50 @@
+
+module ClassInfoImportHelper
+
+  def self.remote_endpoint_path_for(import_line)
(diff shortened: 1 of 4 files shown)


knowuh:
Most of this PR is tests ... the actual implementation (pre-refactor) was in https://github.com/concord-consortium/lara/pull/439/commits/20ec8c5344e0bf1ad1c535a424d15414c1420bfa

PR #441, opened 2019-06-18
Plugin authoring gui

scytacki:
I'm opening this Draft PR so it is easier to track this work.


lara-typescript/src/lib/plugin-context.ts
```diff
@@ -30,12 +41,17 @@ export interface IPluginContext {
...
+export interface IPluginAuthoringContextOptions extends IPluginCommonOptions {
+  /** Type of the plugin */
+  type: "authoring";
+  /** URL used to save plugin authoring state. */
+  authorDataSaveUrl: string;
 }
 
 /** Data generated by LARA, passed to .initPlugin function, and then used to generate final IEmbeddableRuntimeContext */
-export interface IEmbeddableContext {
+export interface IEmbeddableContextOptions {
```
scytacki: Why is `Options` added to the name here?

lara-typescript/src/lib/plugins.spec.ts
```diff
@@ -18,11 +20,12 @@ describe("Plugins", () => {
...
     it ("should let plugin to register itself only once", () => {
-      const pluginConstructor = jest.fn();
+      const runtimeClass = jest.fn();
+      const authoringClass = jest.fn();
       setNextPluginLabel("test");
-      expect(registerPlugin("*DEPRECATED*", pluginConstructor)).toEqual(true);
+      expect(registerPlugin({runtimeClass, authoringClass})).toEqual(true);
       setNextPluginLabel("test");
-      expect(registerPlugin("*DEPRECATED*", pluginConstructor)).toEqual(false);
+      expect(registerPlugin({runtimeClass, authoringClass})).toEqual(false);
```
scytacki: Since authoringClass is optional now. It'd be good to have a test for that.

app/controllers/plugins_controller.rb
```diff
@@ -39,14 +32,9 @@ def update
...
       @plugin.update_attributes(params['plugin'])
-      @plugin.reload
     end
 
-    respond_to do |format|
-      format.js do
-        # will render update.js.erb
-      end
-    end
+    redirect_to(request.env['HTTP_REFERER'].sub(/\?.+/, ''))
```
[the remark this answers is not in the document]
sfentress: Doug and I hacked that solution when trying to understand why the activity-level plugin form didn't reload the activity edit page (like the page-level plugin forms do) and instead tried to load a non-existent plugin page.

The redirect line itself mimics the same line at https://github.com/concord-consortium/lara/blob/master/app/controllers/embeddable/embeddables_controller.rb#L20, but is implemented slightly differently there.

We weren't sure if this was the correct solution (though it appears to work fine).

lara-typescript/src/lib/plugin-context.ts
sfentress: The original name `IEmbeddableContext` was confusing because it is different from the `IPluginRuntimeContext` that is part of the API and is passed to a plugin. Instead this is a set of options that is passed in order to generate the `IPluginRuntimeContext`.

lara-typescript/src/lib/plugins.spec.ts
sfentress: :+1: I will add.

sfentress: Done.

lara-typescript/src/lib/plugin-context.ts
scytacki: To me it is weird that we now have IPluginRuntimeContext and IPluginContextOptions, and IEmbeddableRuntimeContext and IEmbeddableContextOptions.

If we are going with Options then it seems better to have IPluginContext and IPluginContextOptions  or to have IPluginRuntimeContext and IPluginRuntimeContextOptions.  Removing Runtime and adding Options seems inconsistent.

I'm looking to see if there is a new IPluginAuthoringContext...

scytacki: Looking at the code it looks like IEmbeddableContext was also defined in the embeddable-runtime-context.ts file. It looks like that file was not changed by this PR.

Doh, I see there is a IPluginAuthoringContextOptions and IPluginRuntimeContextOptions.  Sorry. So I guess my only problem was really IEmbeddable* . I'm continuing to grok this to see if there is really anything that can be improved...

lara-typescript/src/lib/plugin-context.ts
```diff
@@ -14,12 +15,22 @@ export interface IPluginContext {
   pluginId: number;
   /** The authored configuration for this instance (if available). */
   authoredState: string | null;
+  /** Reserved HTMLElement for the plugin output. */
+  container: HTMLElement;
+  /** Wrapped embeddable runtime context if plugin is wrapping some embeddable. */
+  wrappedEmbeddable: IEmbeddableContextOptions | null;
```
scytacki: So this is added to the IPluginAuthoringContextOptions interface, but IEmbeddableContextOptions (previously IEmbeddableContext) has some fields that are Runtime specific: interactiveStateUrl and interactiveAvailable.  As far as I can tell the wrappedEmbeddable is not passed through to the IPluginAuthoringContext so it is probably best to just remove this from IPluginCommonOptions for now. 

We'll need to add something like it back in to support a real preview, but it will need a different type signature.

app/controllers/plugins_controller.rb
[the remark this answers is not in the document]
scytacki: Oh well, I don't have time to figure out a better way now.

lara-typescript/src/lib/plugin-context.ts
sfentress: We could remove it, but I think we'd add it right back in when supporting preview.

I think the `IEmbeddableContextOptions` type is ok. `interactiveStateUrl` is allowed to be `null`, and if `interactiveAvailable` refers to `clickToPlay` (I think it does), then we probably want to know that when rendering the preview, so that the teacher-edition model looks the same as it would look to the student.

But I don't feel strongly about this, so it could also be removed.

scytacki:
Looking at the docs, it'd be useful to explain more of what the IPluginAuthoringConstructor and IPluginRuntimeConstructor are. Or perhaps they should be removed from the docs if a plugin developer doesn't need to care about them.  https://github.com/concord-consortium/lara/tree/plugin-authoring-gui/docs/lara-plugin-api#ipluginauthoringconstructor

Is the container here: https://github.com/concord-consortium/lara/blob/plugin-authoring-gui/docs/lara-plugin-api/interfaces/ipluginauthoringcontext.md#container where the authoring UI should be placed?  If so it'd be better to make that more explicit in the docs.

lara-typescript/src/lib/plugin-context.ts
scytacki: I'm pretty sure the IEmbeddable* object that is needed for the preview will be different. We talked about having a route in LARA to render the html of the embeddable, so the plugin could use that. So then the wrappedEmbeddable object would contain a URL to this embeddable route. That URL wouldn't be something we'd want to expose to plugin runtime. Supporting click to play will be complicated with this preview approach since it requires not only html but also javascript which seems tricky to get right, so I'd guess we will just avoid it on first pass.

So I still think we should remove it for now, so we we don't feel we need to use this same interface when working on the preview.

lara-typescript/src/lib/plugin-context.ts
scytacki: Probably my rambling above is hiding the one thing that I think needs to change here:

IEmbeddableContext seems to also be defined in embeddable-runtime-context.ts and that file doesn't seem to be changed by this PR.  I don't know why it is defined twice, but if it is renamed it should be renamed there or the duplication should be fixed.

dougmartin:
@scytacki I've removed IEmbeddableContext

scytacki:
So my remaining suggested changes are:

- move wrappedEmbeddable out of IPluginCommonOptions
- document the the constructors better (or remove from docs if unnecessary)
- document what the 'container' is for better

scytacki:
Oh here is another one.  The documentation on this page needs to be updated:
https://github.com/concord-consortium/lara/tree/plugin-authoring-gui/docs/lara-plugin-api

It is still referring to the old 'registerPlugin' method signature.

lara-typescript/src/lib/plugin-context.ts
```diff
@@ -138,7 +138,7 @@ const log = (context: IPluginContextOptions, logData: string | ILogData): void =
 };
 
 const fetchPluginEventLogData = (context: IPluginContextOptions) => {
```
scytacki: It is too late to change this now. But it is code like this, which is why using Options doesn't seem right. This object is not just a set of options to configure the context. It is actually the internal object used by the plugin api implementation.

dougmartin: That shouldn't be using options but rather the context.  Its easy to fix - I'll do that now.

scytacki:
This page is still referring to the old registerPlugin signature:
https://github.com/concord-consortium/lara/tree/plugin-authoring-gui/docs/lara-plugin-api

This container documentation still doesn't really make it clear this is where the plugin should be putting its authoring UI:
https://github.com/concord-consortium/lara/blob/plugin-authoring-gui/docs/lara-plugin-api/interfaces/ipluginauthoringcontext.md#container

The constructor documentation is just repeating what the code is saying.  I think these constructors are the main constructor referred to here, right?
https://github.com/concord-consortium/lara/tree/plugin-authoring-gui/docs/lara-plugin-api#plugin-implementation

But otherwise this PR is fine. I'll add a seperate story for improving the documentation.

PR #444, opened 2019-06-21
Add API Key to LARA Users

knowuh:
[NP/DL]

Addresses the PT story [#166795180] Add API Key to Users

https://www.pivotaltracker.com/n/projects/736901/stories/166795180

Follows the pattern in Portal to implement an API key for LARA users. The UI is implement as an additional check-box in the users-admin screen. When clicked, a UUID style API key is generated and stored in the user's record. If so enabled, the API key is displayed adjacent to the check-box where it can be easily copied.

Un-clicking the checkbox removes the API key from the user (and the database).

REQUIRES MIGRATION, as a column was added to the users table in the schema.

Spec tests have been added.

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -34,6 +34,10 @@ class Ability
       cannot :export, LightweightActivity
       cannot :export, Sequence
     end
+    if user.can_export?
+      can :export, LightweightActivity
+      can :export, Sequence
+    end
     # Everyone (author and regular user) can read public, hidden and archived sequences or activities.
     ['public', 'hidden', 'archive'].each do |allowed_status|
       can :read, Sequence, :publication_status => allowed_status
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -13,7 +13,8 @@ class User < ActiveRecord::Base
   has_many :imports
 
   # Setup accessible (or protected) attributes for your model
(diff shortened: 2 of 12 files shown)


scytacki:
Did you check the existing code that is using Bearer tokens to make sure this devise code doesn't break it?   

That existing code uses a module called `PeerAccess`. If that code is broken by this new code then I think the solution is to disable devise on the controller actions that are using the `PeerAccess#authorize_peer!` method. I don't know how to disable devise but I'm pretty sure there is a way.

The other possible issue with PeerAccess is that it is disabling `verify_authenticity_token` for each action in the whole controller if the token matches. This is using PeerAccess's definition of "matches" not the API key's code definition of matches.  So that will conflict with these API key based token matches if we try to use one for a POST request in one of the PeerAccess controllers.   

`authorize_peer!` is only used in two actions and both of them are only used by the portal. So it seems safe to disable the `verify_authenticity_token` just for those two actions explicitly in the two controllers.

This verify_authenticity_token code does bring up an issue for these API keys. They are going to need to do the same thing if we want to use them for actions that are being POST'd to.  I think it is fine to punt dealing with that issue for now. And instead put a comment in the code that these API keys only work for GET requests. And perhaps a description in the User editing form where the API can be enabled saying that the key can only be used for GET requests.

PR #446, opened 2019-06-25
Updated plugin api documentation [#166851983]

dougmartin:
- converted registerPlugin to use new single object parameter signature
- added authoring information to plugin api readme, including which plugins don't support authoring
- added manifest section to plugin api readme
- removed unneeded constructor type aliases to reduce confusion in the generated docs


pjanik:
Looks good 👍 

> * removed unneeded constructor type aliases to reduce confusion in the generated docs

Not sure about this one. But as long as constructors are getting type-checked in the plugins, I'm fine. That was pretty useful when I was updating plugins to V3 - I saw my calls to `registerPlugin` being highlighted. But if I read the code correctly, this will still work, right?

dougmartin:
@pjanik Yes the typecheck for the constructors still works.  The two type aliases had to be exported because they were part of an exported interface but they really are internal so it made sense to just inline the type alias and get rid of the need to document them.

scytacki:
@dougmartin not sure if you saw my last comment above. ⬆️

dougmartin:
@scytacki I updated the docs pre your request.

PR #448, opened 2019-06-27
Fix plugin wrapping of interactives when no info/assessment block is defined  [#166724753]

dougmartin:
Update:
Instead of removing check for page.show_info_assessment I instead created a new view that outputs all wrapper plugins before the if check.  The view that lists embeddables under the if is now provided an array without the wrapper plugins.

Originally:
Removes check for page.show_info_assessment before listing embeddables.  Since plugins are treated as hidden embeddables the plugin show code was being skipped if no info or assessment blocks were added to the page.

### DIFF
--- a/app/helpers/interactive_page_helper.rb
+++ b/app/helpers/interactive_page_helper.rb
@@ -26,6 +26,14 @@ module InteractivePageHelper
     return link_to name, runnable_activity_page_path(activity,page), opts
   end
 
+  def is_wrapping_plugin?(e)
+    e.respond_to?(:wrapping_plugin?) && e.wrapping_plugin?
+  end
+
+  def main_section_wrapping_plugins(page, run)
+    page.main_visible_embeddables.select { |e| is_wrapping_plugin?(e) }
+  end
+
   def main_section_visible_embeddables(page, run)
     finder = Embeddable::AnswerFinder.new(run)
     # Limit visible embeddables to ones that do not belong to any section.
--- a/app/views/interactive_pages/_show.html.haml
+++ b/app/views/interactive_pages/_show.html.haml
@@ -12,6 +12,7 @@
       = render partial: 'interactive_pages/interactive', locals: {page: page, layout: layout}
 
   .questions-mod.ui-block-2{ :class => page.embeddable_display_mode == 'carousel' ? 'jcarousel' : '' }
+    -# Wrapping plugins don't necessarily wrap items in the assessment block, so they need to be rendered separately.
     = render partial: 'interactive_pages/list_wrapping_plugins', locals: {wrapping_plugins: main_section_wrapping_plugins(page, @run)}
     - if page.show_info_assessment
       = render partial: 'interactive_pages/list_embeddables', locals: {embeddables: main_section_visible_embeddables(page, @run).select {|e| !is_wrapping_plugin?(e) }}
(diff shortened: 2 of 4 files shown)


dougmartin:
@pjanik I tagged you for review as this removes an if check you added last year as part of this story:

https://www.pivotaltracker.com/n/projects/736901/stories/155515297

pjanik:
Hm, it's a bit difficult for me to verify that. If I added that one year ago trying to fix responsive layout, I don't think that it became unnecessary now, just like that. I guess you'd need to check responsive layout rendering. It might work, as my fix could have been about ensuring that `.questions-mod.ui-block-2` always exists rather than this `if` specifically.

dougmartin:
@pjanik I changed the fix to keep the if check and add an explicit output of the wrapper plugins.  Could you re-review?

app/helpers/interactive_page_helper.rb
```diff
@@ -26,11 +26,17 @@ def page_link(activity,page, opts={})
...
+    page.main_visible_embeddables.select { |e| e.is_hidden_plugin? }
+  end
+
   def main_section_visible_embeddables(page, run)
     finder = Embeddable::AnswerFinder.new(run)
     # Limit visible embeddables to ones that do not belong to any section.
     # Don't return answer objects for interactives.
-    page.main_visible_embeddables.map { |e| Embeddable::is_interactive?(e) ? e : finder.find_answer(e) }
+    page.main_visible_embeddables
+      .select { |e| !e.is_hidden_plugin? }
```
pjanik: it feels a bit strange `main_visible_embeddables` returns hidden plugins.
Could we just move this `.select {...}` to `main_visible_embeddables`?
So, `main_visible_embeddables` would return whatever it returns now + `.select { |e| !e.is_hidden_plugin? }`.

app/models/embeddable.rb
```diff
@@ -79,6 +79,10 @@ def show_in_edit?
     true
   end
 
+  def is_hidden_plugin?
+    false
+  end
```
pjanik: `is_hidden_plugin?` is a bit confusing to me. As we're adding one more term on top of other plugin properties. I'd tend to be explicit and just check for `wrapping_plugin?` as then it's easier to understand this whole problem.
Using `hidden plugin` around, it makes me think that plugins can be somehow hidden.

But I'd leave it up to you whether to change that or not.

dougmartin:
@pjanik I updated the code based on your comments.

app/views/interactive_pages/_show.html.haml
```diff
@@ -12,8 +12,9 @@
       = render partial: 'interactive_pages/interactive', locals: {page: page, layout: layout}
 
   .questions-mod.ui-block-2{ :class => page.embeddable_display_mode == 'carousel' ? 'jcarousel' : '' }
+    = render partial: 'interactive_pages/list_wrapping_plugins', locals: {wrapping_plugins: main_section_wrapping_plugins(page, @run)}
```
pjanik:
```suggestion
    -# Wrapping plugins don't necessarily wrap items in the assessment block, so they need to be rendered separately.
    = render partial: 'interactive_pages/list_wrapping_plugins', locals: {wrapping_plugins: main_section_wrapping_plugins(page, @run)}
```

dougmartin:
@pjanik I added the comment and I also updated the list embeddables view to use the is_wrapped_plugin? helper.

pjanik:
👍

PR #450, opened 2019-07-16
Yet another fix of interactive sizing

pjanik:
This PR addresses a few separate issues (but all related to dynamic interactive sizing):

1. Size calculations were triggered too early, while some resources were still loading (e.g. images or stylesheets provided by plugins). It was causing issues with wrapping plugins described in this story: https://www.pivotaltracker.com/story/show/167044217. To fix that, it was enough to change `document.ready` event to `window.load`.

2. Sometimes interactive header width wasn't matching interactive size. That was caused by the fact that at the beginning of `setSize` execution, we're only resetting the size of the iframe and we were forgetting about the header. So, sometimes header was using previously calculated width (e.g. when interactive provides its own aspect ratio and calculations need to be triggered again). This has been fixed by enduring that we always change both the size of the interactive and its header.

3. To fix the problem described here: https://www.pivotaltracker.com/story/show/167317323, I changed DOM elements that have `width` updated. Instead of setting `width` separately for iframe and for its header, I just update the width of the whole `embeddable-container`. That will let plugins observe this value easily (see lara-sharing-plugin PR: https://github.com/concord-consortium/lara-sharing-plugin/pull/17) and adjust their own dimensions if necessary.

### DIFF
--- a/app/assets/javascripts/interactive-sizing.js
+++ b/app/assets/javascripts/interactive-sizing.js
@@ -1,6 +1,22 @@
 /*global $ */
 
 function interactiveSizing () {
+  function setInteractiveWidth($iframe, width) {
+    $iframe.css('width', width);
+    var $header = $iframe.parents(".embeddable-container").find(".question-hdr")
+    if ($header.length > 0) {
+      $header.css('width', width);
+    }
+  }
+
+  function limitInteractiveHeight($iframe, maxHeight) {
+    if ($iframe.height() > maxHeight) {
+      var scale = (maxHeight / $iframe.height()) * 100 + '%';
+      setInteractiveWidth($iframe, scale);
+      $iframe.css('height', maxHeight);
+    }
+  }
+
   function setSize () {
     var $iframe = $(this);
     var aspectRatio = $iframe.data('aspect-ratio');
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/interactive-sizing.js
```diff
@@ -1,42 +1,65 @@
...
-        // Rescale question header too (it's present when interactive saves state).
-        var $header = $iframe.parents(".embeddable-container").find(".question-hdr")
-        if ($header.length > 0) {
-          $header.css('width', scale);
-        }
-      }
+      $iframe.css('height', $iframe.width() / aspectRatio);
+      // PJ [7/16/2019]: Why is the height limited only in the DEFAULT mode, but not in MANUAL?
+      // I don't find this intuitive, but it has been implemented because of some issues described
+      // in this story: https://www.pivotaltracker.com/n/projects/736901/stories/161458506
```
scytacki: That story describes it well. For interactives that don't scale well, we need a way to force them to have a height larger than the maxHeight.  

Additionally, @dougmartin is about to add support for interactives that dynamically request more height.  So in this case we'll have interactives that need more than the maxHeight and require a page scroll to see the whole interactive. 

There is probably a better way to handle all this but we haven't figured it out yet. The variables that need to be considered are:
- not too confusing for authors
- works well when author changes page layout
- supports the variety of sizing options required by interactives and authors

PR #451, opened 2019-07-16
Publish Activity changes to reporting service

knowuh:
When the structure of an Activity or Sequence changes, enqueue an update the FireStore reporting service if one is configured.

This implements the same basic pattern as is used to update the Portal.  Some minor differences:

* If there is no reporting service configured, we don't send anything (duh).
* If the Activity updated time is newer than the enqueue time, then we don't send it. We wait for the newly queued job to send it.
* We save a hash of the last serialized json hash.
* We don't send anything if the serialzed json hasn't changed from last update.  

[#165204332]

https://www.pivotaltracker.com/story/show/165204332

### DIFF
--- a/spec/models/send_to_report_service_job_spec.rb
+++ b/spec/models/send_to_report_service_job_spec.rb
@@ -56,18 +56,7 @@ describe "SendToReportServiceJob" do
       end
     end
 
-    describe "when the published item has been modified afeter enqueing" do
-      let(:queuetime) { 1.hour.ago }
-      it "the job should not throw an exception" do
-        expect { job.perform }.not_to raise_error
-      end
-      it "the job should not make a requst to the report server" do
-        expect(sender).not_to receive(:send)
-        job.perform
-      end
-    end
-
-    describe "when the published item has been modified afeter enqueing" do
+    describe "when the published item has been modified afrer enqueuing" do
       let(:queuetime) { 1.hour.ago }
       it "the job should not throw an exception" do
         expect { job.perform }.not_to raise_error
@@ -83,7 +72,7 @@ describe "SendToReportServiceJob" do
       it "the job should not throw an exception" do
         expect { job.perform }.not_to raise_error
       end
-      it "the job should not make a requst to the report server" do
+      it "the job should not make a request to the report server" do
         expect(sender).not_to receive(:send)
         job.perform
       end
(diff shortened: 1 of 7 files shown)


knowuh:
@scytacki -- as you pointed this branch never made it to a into pull request. I thought it had already made it to master. 

@dougmartin --  I would appreciate a review from you too, to see if I have missed any edge cases.

spec/models/send_to_report_service_job_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+      it "should not raise an excpetion" do
+        expect {  job.perform }.not_to raise_error
+      end
+      it "should not send a request to the report server" do
+        expect(sender).not_to receive(:send)
+        job.perform
+      end
+    end
+
+    describe "when the published item has been modified afeter enqueing" do
```
scytacki:
```suggestion
    describe "when the published item has been modified after enqueuing" do
```

spec/models/send_to_report_service_job_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+      it "the job should not throw an exception" do
+        expect { job.perform }.not_to raise_error
+      end
+      it "the job should not make a requst to the report server" do
+        expect(sender).not_to receive(:send)
+        job.perform
+      end
+    end
+
+    describe "when the published item has been modified afeter enqueing" do
```
scytacki: This block: https://github.com/concord-consortium/lara/blob/9213d47c2d3ba64f1cbf69433ac7c8ebe3386b56/spec/models/send_to_report_service_job_spec.rb#L70-L79 

looks like a duplicate of the block above

spec/models/send_to_report_service_job_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+        job.perform
+      end
+    end
+
+    describe "when the published item has the same payload_hash" do
+      let(:last_report_service_hash) { payload_hash }
+      it "the job should not throw an exception" do
+        expect { job.perform }.not_to raise_error
+      end
+      it "the job should not make a requst to the report server" do
```
scytacki:
```suggestion
      it "the job should not make a request to the report server" do
```

scytacki:
👍

PR #454, opened 2019-07-24
167209213 remove `external_report_url` from Activities / Sequences

knowuh:
We will be managing `external_reports` exclusively from the portal side.

This PR removes all `external_report_url` references from Activities and Sequences.

See: https://www.pivotaltracker.com/story/show/167209213

### DIFF
--- a/app/views/sequences/_form.html.haml
+++ b/app/views/sequences/_form.html.haml
@@ -30,11 +30,6 @@
         .activity_thumbnail#thumbnail_preview
           %img
     .field
-      = f.label :external_report_url, t("EXTERNAL_REPORT_URL")
-      = f.text_field :external_report_url
-      %br
-      .hint=t("EXTERNAL_REPORT_WARNING")
-    .field
       = f.label :theme_id, "Theme"
       = f.select :theme_id, options_from_collection_for_select(Theme.all, 'id', 'name', @sequence.theme_id), { :include_blank => "None (inherit from project, or site default)" }
     .field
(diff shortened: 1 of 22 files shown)


knowuh:
Local specs were all green. 

@kevinsantos-cc  -- This PR should only impact the following items:

* Serial representation of activities and sequences (export / import / duplicate)  I updated the JSON files that I think are used in Cypress tests.

* The new & edit forms for activities and sequences.  I smoke-tested those forms locally with no issues.

This is a big PR, but it was mostly mechanical.

app/views/sequences/_form.html.haml
```diff
@@ -29,11 +29,6 @@
           [preview]
         .activity_thumbnail#thumbnail_preview
           %img
-    .field
-      = f.label :external_report_url, t("EXTERNAL_REPORT_URL")
-      = f.text_field :external_report_url
-      %br
-      .hint=t("EXTERNAL_REPORT_WARNING")
```
dougmartin: Maybe too OCD of me but shouldn't `EXTERNAL_REPORT_WARNING` be removed from en.yml?

knowuh: Nice catch.

knowuh: Added 7ce35f8 to do this.  Thanks.

knowuh:
@dougmartin -- Good question about the holding off on the migration.  I see your point, but the portal should have a copy of this information already, and it will no longer be accepting `external_report` changes from LARA.

Those `external_report_urls` have been captured in the portal they were published too.  

Portal has a migration which copies those URLs from `external_report` to a join table.  The data shouldn't be lost.

@scytacki -- Did we want to do an addition export of these URLS before migrating the data?

scytacki:
I don't think we need to do an export to capture the report urls from LARA. As you said these are already saved in the portal. The URL in LARA counterparts has just added confusion, so it is better to remove sooner than later. In other words I think it is best to just to remove them without waiting.

Removing them now will also help us more quickly find any possible code in LARA that is assuming this url is available.

PR #457, opened 2019-08-06
fix embeddable_plugin copying

scytacki:
In the previous code, without an id, then the setup of the embeddable.plugin.plugin_scope was failing. This only happened during duplicate. I didn't track down exactly why, but I'd guess
in other cases the embeddable was being saved first so it actually had an id.
Perhaps this happens on the `embeddable.plugin = Plugin.create({})` call.

I initially switch to after_create to make sure the embeddable had an id. However
as the comment said in that commit it was better to remove the duplicate plugin_id here. 

So the final change is to remove the plugin_id from EmbeddablePlugin, and switch to an after_initialize callback.  This also includes a migration that serves two purposes: 
1. clean up any EmbeddablePlugins that that pointed at plugins with a nil plugin_scope_id
2. remove the plugin_id from EmbeddablePlugin

### DIFF
--- a/app/models/embeddable/embeddable_plugin.rb
+++ b/app/models/embeddable/embeddable_plugin.rb
@@ -31,11 +31,16 @@ module Embeddable
     delegate :component_label=, to: :plugin
     delegate :component, to: :plugin, allow_nil: true
 
-    before_create do |embeddable|
+    # TODO: this approach isn't good, it is a circular reference in the database
+    # It'd be better if the embeddable.plugin method computed the plugin by finding
+    # plugin that referenced this embeddable with its plugin_scope
+    # this might work with a simple has_one :plugin, as: plugin_scope
+    after_create do |embeddable|
       unless(embeddable.plugin)
         embeddable.plugin = Plugin.create({})
       end
       embeddable.plugin.plugin_scope = embeddable
+      embeddable.plugin.save!
     end
 
     def self.name_as_param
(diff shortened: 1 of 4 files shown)


scytacki:
@dougmartin can you particularly look at the SQL in the migration? I'm not that fluent, and not confident there isn't some mistake with the inner join and possibly the where clause in the `down`.  I did test it out locally so at least the syntax is correct.

app/models/embeddable/embeddable_plugin.rb
```diff
@@ -10,7 +10,7 @@ def self.table_name
     attr_accessible :plugin, :approved_script_id, :description, :author_data,
     :is_full_width, :is_hidden, :component_label
 
-    belongs_to :plugin, autosave: true
+    has_one :plugin, as: :plugin_scope
```
dougmartin: I will leave it to @knowuh to review this part - I'm still not confident in my knowledge of rails model relationship magic.

db/migrate/20190806191648_remove_duplicate_plugin_id.rb
```diff
@@ -0,0 +1,36 @@
+class RemoveDuplicatePluginId < ActiveRecord::Migration
+
+  def up
+    # need to go through all EmbeddablePlugins and make sure each of their
+    # plugins correctly sets the plugin_scope field on the plugin
+    # we cannot easily use the polymorphic support of rails because we are also using
+    # temporary model classes, so plugin_scope_type added by the polymorphic code would
+    # end up being:  RemoveDuplicatePluginId::Embeddable::EmbeddablePlugin
+    execute <<-SQL
```
dougmartin: These queries look good - the inner join with the plugin on the "left" guarantees no nulls in the embeddable_plugins.

PR #459, opened 2019-08-09
165887899 wrapped authoring gui

dougmartin:


app/views/plugins/_author.haml
```diff
@@ -1,6 +1,21 @@
 - if plugin.version == 3
   - v3_plugin_label = "plugin#{plugin.approved_script_id}"
   - output_id = "plugin_authoring_output"
+  - wrapped_id = "wrapped_embeddable"
+  - show_wrapped_embeddable = @embeddable && (plugin.component.scope == "embeddable-decoration") && plugin.component.guiPreview
+  - wrapped_embeddable = show_wrapped_embeddable && @embeddable.embeddable
+  - wrapped_embeddable_lara_json = wrapped_embeddable ? (LaraSerializationHelper.new).export(wrapped_embeddable).to_json : 'null'
+  - if wrapped_embeddable && wrapped_embeddable.respond_to?(:answers)
+    - wrapped_embeddable = wrapped_embeddable.answers[0]
```
scytacki: This approach won't work.  Here are two problems:
1. there might not be any answers for this embeddable especially if an author has just created it and hasn't previewed it yet.
2. this will load up every single answer from the database that is associated with this embeddable and then just take the first one

What you can do instead is create Run and use the answer finder to get an answer for the embeddable:
```
    finder = Embeddable::AnswerFinder.new(Run.new())
    Embeddable::is_interactive?(wrapped_embeddable) ? wrapped_embeddable : finder.find_answer(wrapped_embeddable)
```

This is how the interactive_page does it. And the print_blank route uses this Run.new() approach. So all of the code should handle Runs that haven't been saved to the database yet.

dougmartin:
@scytacki I've made all the suggested changes, each as a separate commit.

app/views/plugins/_author.haml
scytacki: I tested that approach out locally and it seems to work for me.

scytacki:
I tested this out locally and it looks like the plugin code which is trying to add the checkboxes to the correct answer are not working correctly.  Perhaps there is something wrong with my local setup:

image

scytacki:
New changes look good. 👍 
Only known thing left is the issue with the check marks.

scytacki:
@dougmartin in standup I think you said that the checkboxes were broken in the runtime. At least in a production activity they seem to be working: 
image

This is from this page: https://authoring.concord.org/sequences/501/activities/9699/pages/124937/?mode=teacher-edition

scytacki:
They also work on the standalone authoring page: http://teacher-edition-tips-plugin.concord.org/authoring.html

And they seem to be working on authoring production where the TE authoring is embedded on the LARA authoring page, but it doesn't have a real preview.

PR #463, opened 2019-08-30
Fix collaboration runs, set LTI params

pjanik:
Also, cleanup logic of submit_dirty_answers_job a bit (removed redundant ifs) + added logic and job aborting if POST to Firestore fails.

### DIFF
--- a/app/controllers/sequences_controller.rb
+++ b/app/controllers/sequences_controller.rb
@@ -219,9 +219,9 @@ class SequencesController < ApplicationController
       if portal.valid?
         @sequence_run.disable_collaboration
       end
-      # If we have new class_info_url or class_hash update it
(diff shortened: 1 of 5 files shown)


pjanik:
Thanks, that was a good catch. Previous tests based on doubles weren't testing much actually, so I reworked them quite a bit.

> The fact that line 9 is not covered is confusing because the spec "should raise error to re-run job". should hit line 9. But perhaps the send method is never called because an error is raised on line 8 for some reason. I think the spec could check that the error raised is actually "bang", to rule out this possibility.

Yes, an exception was thrown in the constructor of ReportService::RunSender.

New tests check for "bang", and they also check what happens if report service returns non-success response.

PR #464, opened 2019-09-04
Added protocol to session domain lookup in jwt controller

dougmartin:


scytacki:
Actually what might be more useful in the long term is to store the Authentication#id in the session. That way we could look up the provider from the Authentication. And we'd also have info on the specific authentication that is being used which includes the token for making portal requests.

dougmartin:
Closed in favor of new PR: https://github.com/concord-consortium/lara/pull/465

PR #469, opened 2019-09-13
Waters theme

emcelroy:
PT story: https://www.pivotaltracker.com/story/show/168480078


app/assets/stylesheets/theme-waters.scss
```diff
@@ -0,0 +1,265 @@
...
+          opacity: 1;
+        }
+        i.fa { display: none; }
+        span.text { display: inline; }
+      }
+    }
+  }
+} // end #nav-activity-menu
+
+.precipitating-change-footer {
```
scytacki: Is this correct that there is a precipitating-change footer element?
Is that because the activity has a project setting of precipitating-change?

emcelroy: Thanks for catching that. Those should've been updated from 'precipitating-change' to 'waters-'. I've made that change.

PR #474, opened 2019-10-03
Add report tool id config

scytacki:
This is the LARA side of this PR: 
https://github.com/concord-consortium/portal-report/pull/130

The tool id is configured in LARA using the optional REPORT_SERVICE_TOOL_ID environment variable.

If I have time I'd like to remove the need for the REPORT_SERVICE_SELF_URL. It really is only needed during rake task runs when it isn't clear what the host url is. In other case we can figure out the host url from the request and save it in the delayed jobs which is how the portal publishing works. But that work is separate from this work, so it'll be in a separate PR.

### DIFF
--- a/app/services/report_service.rb
+++ b/app/services/report_service.rb
@@ -47,6 +47,18 @@ module ReportService
       ENV['REPORT_SERVICE_TOKEN']
     end
 
+    def tool_id
+      if ENV['REPORT_SERVICE_TOOL_ID'].present?
+        ENV['REPORT_SERVICE_TOOL_ID']
+      else
+        self_url
+      end
+    end
+
+    def source_key
+      ReportService::make_source_key(tool_id)
+    end
+
     def api_endpoint
       "#{report_service_url}/#{api_method}"
     end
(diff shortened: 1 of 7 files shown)


pjanik:
It looks good to me. It requires some manual steps, but fairly simple. 
I don't think if that should be part of this PR, but I'm wondering if we should try to generate this TOOL_ID automatically? E.g. append username? It seems to be available udner `$USER` env variable on OS X. That should be unique among developers. Then, perhaps we could do a similar thing in Portal Report (probably playing a bit with webpack config).

scytacki:
Yeah I was thinking the same thing about doing it automatically. If you see an easy way to do that, please do. I guess it would be set in the docker-compose.yml (assuming it can evaluate local environment variables). 

The other thing I've been thinking about is having LARA send the tool_id to the portal. It is also ready sending a 'source' which is like the tool_id. If the portal was storing the tool_id, then it could be sent down to the portal-report.  Then this way a developer wouldn't need to customize the external report url. A side benefit of this is that the same external report could be used by multiple lara's or other tools. And it kind of makes sense that the portal would know the tool_id of the external_activity.  But before I go to far with that I want to look at LTI again to see if this fits with their model.

PR #476, opened 2019-10-15
sequences should be republished to report service

scytacki:
this also adds support to debug the app when running in docker
this also replaces some unused shared example code in the embeddable_controller_helper

https://www.pivotaltracker.com/story/show/169066612

### DIFF
--- a/app/models/publishable.rb
+++ b/app/models/publishable.rb
@@ -137,10 +137,20 @@ module Publishable
       has_many :portal_publications, :as => :publishable, :order => :updated_at
 
       # all changes will be queued for auto publishing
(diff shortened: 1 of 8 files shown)


scytacki:
Yeah it isn't that obvious. 
Previously only `queue_auto_publish_to_portal` was calling itself on the sequences of the activity. I could have fixed it by adding the same block of code to  `queue_publish_to_report_service`. 

What I did instead was to add a new method `queue_publish` which calls `queue_auto_publish_to_portal` and `queue_publish_to_report_service`. And then after calling those methods it calls itself (`queue_publish`) on the activity's sequences. This way the sequences will be republished to both the portal and the report service.

PR #484, opened 2020-01-13
Updated image zoom to replicate glossary zoom [#170155168]

dougmartin:


dougmartin:
@scytacki use the "hide whitespace changes" options to see the changes in the `jquery.colorbox.js` file (otherwise it will show all the trimmed lines in the diff)

PR #485, opened 2020-01-14
Set portal info in a newly created activity run when a sequence is launched

pjanik:
@scytacki, I had a few iterations on this, even though the fix is tiny.

Actually, the old code:
```
def update_platform_info(url_params)
    updates = super(url_params)
    if (updates)
      self.runs.each { |r| r.update_platform_info(url_params) }
    end
end
```
 would work perfectly fine if it was without this `if` statement. And my final fix is pretty much based on that. The problem was that when sequence run already had platform info, `updates` was equal to false, and we're skipping activity run updates. The other issue with this is that `update_platform_info` function defined in mixin isn't really prepared to return information whether there were any updates or not (it worked by accident, as the last statement is one of the updates).

Your suggestion to update `SequenceRun#run_for_activity` would work fine too and it was my first approach. But I'd also need to update `SequenceRun#lookup_or_create` to save platform info (before activity runs are created). Or still keep using `SequenceRun#update_platform_info`. Generally, we'd start adding this code to various lookup_or_create functions. And I'd be totally lost when platform info is set or when it can be set.

I thought that it's better to call `update_platform_info` in controllers explicitly as we used to do before. At least it's visible, more explicit and it's a bit easier to follow. But it's easy to change if you think the other way is better.

### DIFF
--- a/db/migrate/20200114203042_add_missing_platform_info.rb
+++ b/db/migrate/20200114203042_add_missing_platform_info.rb
@@ -1,0 +1,13 @@
+class AddMissingPlatformInfo < ActiveRecord::Migration
+  def up
+    runs = Run.joins(:sequence_run).where('runs.platform_id IS NULL AND sequence_runs.platform_id IS NOT NULL').pluck(:id)
+    puts "Found #{runs.count} runs with missing platform info"
+    runs.each do |run_id|
+      run = Run.find(run_id)
+      run.update_platform_info(run.sequence_run.attributes)
+    end
+  end
+
+  def down
+  end
+end
--- a/db/migrate/20200114203042_add_missing_platform_info.rb
+++ b/db/migrate/20200114203042_add_missing_platform_info.rb
@@ -13,7 +13,7 @@ class AddMissingPlatformInfo < ActiveRecord::Migration
       sender = ReportService::RunSender.new(run, { send_all_answers: true })
       result = sender.send
       if !result || !result["success"]
-        puts "Failed to send Sequence Run: #{run.id}"
+        puts "Failed to send Run: #{run.id}"
       end
     end
   end
--- a/db/migrate/20200114203042_add_missing_platform_info.rb
+++ b/db/migrate/20200114203042_add_missing_platform_info.rb
@@ -5,6 +5,16 @@ class AddMissingPlatformInfo < ActiveRecord::Migration
     runs.each do |run_id|
       run = Run.find(run_id)
       run.update_platform_info(run.sequence_run.attributes)
+      # Sending run to firestore
+      # check that the local environment has
+      # ENV["REPORT_SERVICE_SELF_URL"]
+      # ENV["REPORT_SERVICE_URL"]
+      # ENV["REPORT_SERVICE_TOKEN"]
+      sender = ReportService::RunSender.new(run, { send_all_answers: true })
+      result = sender.send
+      if !result || !result["success"]
+        puts "Failed to send Sequence Run: #{run.id}"
+      end
     end
   end
 
--- a/app/controllers/sequences_controller.rb
+++ b/app/controllers/sequences_controller.rb
@@ -204,6 +204,8 @@ class SequencesController < ApplicationController
   def find_or_create_sequence_run
     if sequence_run_key = params['sequence_run_key']
       @sequence_run = SequenceRun.where(key: sequence_run_key).first!
+      @sequence_run.make_or_update_runs
+      @sequence_run.update_platform_info(params)
       return @sequence_run
     end
 
@@ -220,6 +222,8 @@ class SequencesController < ApplicationController
         @sequence_run.disable_collaboration
       end
     end
+    # Update platform info in case sequence run has been just created or new activities have been added.
+    @sequence_run.update_platform_info(params)
   end
 
 end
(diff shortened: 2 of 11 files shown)


scytacki:
Unfortunately I probably won't have a lot of time to look at this. But without looking closely to see if it is valid, my concern is that the url_params will not be setup the right way when students return to the sequence. They will likely be redirected to an activity page and that redirect process removes the parameters and I think will the update call is skipped until after the redirect at least it used to be. I'll look the code more to check this, but wanted to put it down while I had time.

scytacki:
Looking at the tests it seems like it is not covering the case I was worried about above. Where the student has navigated into a page of the activity, and then an activity is added to the sequence, and then the student launches it again.  Perhaps I'm mis remembering and the update call will happen before the redirect when the parameters are still in the URL.

pjanik:
> Where the student has navigated into a page of the activity, and then an activity is added to the sequence, and then the student launches it again

I'm wondering what difference navigating to a page makes?

I think that when we launch from Portal, we always send the student to a basic sequence URL (like `sequence/123` + url params) as that's the URL that Portal knows. So, that means that we'll always trigger `show` action of the sequences controller. Which is actually calling `find_or_create_sequence_run` first, where we update activity runs with platform info. And then yes, we'll redirect, but that happens after the update. So, I think there should not be much difference or I'm missing something.

But another case came to my mind. A student is running a sequence, a new activity is added, and the teacher asks students just to refresh a page (without going back to the Portal). In this case, we will never be on a page with URL params with portal info. To fix this case, we could probably update various Run lookup/create methods (there's quite a few of them). Check if there's sequence available, and if so, reuse its platform info. Do we want to handle this case in this PR (or handle at all)? Probably it'd require testing and documenting various locations where student can refresh the page (sequence index, activity index, activity page, activity summary etc.) and checking which lookup_or_create methods are triggered.

scytacki:
Yes you are right. So launches from the portal are taken care of. 
Based on what you wrote I can think of 2 edge cases that won't work though:
- student reloads the window when the activity has been added after the initial portal launch
- student navigates to around the activity when the activity has been added after the initial portal launch by the student.

I think it is best try to address those edges cases in this PR.  I think if you go the other route and only setup the runs as their are needed.  There are tests for all of those places you are worried about I think. I'll try to write more later.

scytacki:
I don't have the code in front of me so I'm going from memory. But what I was thinking is that update platform info would stay, but it wouldn't change the activity runs. So that means sequence run doesn't need to override the method. Then there is another method that finds the activity runs I guess that is lookup or create. So if that then sets the platform info it should take care of the edge cases.

I see your point about making sure we find all the cases that access activity runs. But I we need to do that as long as we are fully supporting the addition of activities after the sequence is started. I think we've covered all of those cases already, bit maybe I forgetting something. To be more safe we could override the sequencerun#runs method so we create and update the runs in that case.

Also if we are careful to not initially create the runs until they are needed and at that point set the platform info. This means if we missed something we will also see a broken ui since the run we expect to be there will be null.

pjanik:
Okay. I think my first approach was like that, pretty much following your comments. But in the initial PR message I described why I changed it:
>Your suggestion to update SequenceRun#run_for_activity would work fine too and it was my first approach. But I'd also need to update SequenceRun#lookup_or_create to save platform info (before activity runs are created). Or still keep using SequenceRun#update_platform_info. Generally, we'd start adding this code to various lookup_or_create functions. And I'd be totally lost when platform info is set or when it can be set.

I just thought that explicit calls to `update_platform_info` in the main action of main controllers are a bit more clear than this logic baked into various lookup/create methods. But that's quite opinionated. Also, it seems that to support page reloading, we have to do that anyway. It seems there's no choice anyway.

db/migrate/20200114203042_add_missing_platform_info.rb
```diff
@@ -0,0 +1,13 @@
+class AddMissingPlatformInfo < ActiveRecord::Migration
+  def up
+    runs = Run.joins(:sequence_run).where('runs.platform_id IS NULL AND sequence_runs.platform_id IS NOT NULL').pluck(:id)
+    puts "Found #{runs.count} runs with missing platform info"
+    runs.each do |run_id|
+      run = Run.find(run_id)
+      run.update_platform_info(run.sequence_run.attributes)
+    end
```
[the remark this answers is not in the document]
pjanik: Thanks, done.

db/migrate/20200114203042_add_missing_platform_info.rb
```diff
@@ -5,6 +5,16 @@ def up
...
       run.update_platform_info(run.sequence_run.attributes)
+      # Sending run to firestore
+      # check that the local environment has
+      # ENV["REPORT_SERVICE_SELF_URL"]
+      # ENV["REPORT_SERVICE_URL"]
+      # ENV["REPORT_SERVICE_TOKEN"]
+      sender = ReportService::RunSender.new(run, { send_all_answers: true })
+      result = sender.send
+      if !result || !result["success"]
+        puts "Failed to send Sequence Run: #{run.id}"
```
scytacki: Since this is now sending a activity run instead of a sequence run, this log message should be updated.
I'm assuming the RunSender will work fine with either sequence or activities runs, right?

pjanik: Yes, your code is also sending activity runs, not sequence runs. I changed this log message by accident. Fixed, thanks.

pjanik:
Okay, I think now the fix is closer to what you expect. I've also reused RemotePortal class (you suggested that on Slack).

scytacki:
Looks great to me.

PR #499, opened 2020-03-16
Waters theme footer nav

emcelroy:
Related PT story: https://www.pivotaltracker.com/story/show/171978879

### DIFF
--- a/app/assets/stylesheets/activity-menu.scss
+++ b/app/assets/stylesheets/activity-menu.scss
@@ -16,6 +16,11 @@
 }
 
 .activity-nav-mod {
+  &.footer-nav {
+    .ui-block-2 {
+      width: 769px;
+    }
+  }
   .sticky-wrapper {
     margin-bottom: 0px;
     position: fixed;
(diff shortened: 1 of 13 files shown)


emcelroy:
@scytacki: The WATERS project requested that their activities have pagination links at the bottom of each activity page as well as at the top. Do you generally agree with this approach? (I still need to work out how to exclude the non-pagination stuff at the bottom of each page, but didn't want to do that till I confirmed this is the way to go.)

app/assets/stylesheets/activity-menu.scss
```diff
@@ -16,6 +16,11 @@
 }
 
 .activity-nav-mod {
+  &.footer-nav {
+    .ui-block-2 {
+      width: 769px;
+    }
+  }
```
scytacki: Ethan is going to look at whether this is necessary.
Since some themes modify the site-width class, it is possible that this hard coded width would not look right in those cases. But it seems to look fine when testing.

PR #505, opened 2020-04-29
[WIP] Add interactive library admin

dougmartin:
This a WIP branch to allow for review during development:

Currently done:

1. Added React admin CRUD placeholder to `lara-typescript` folder as a new generated javascript
2. Added admin protected route in home controller to page that uses exported window namespace from step 1 to render a placeholder message on the page.


scytacki:
This looks OK to me. 

My biggest concern is that it is introducing a new pattern.
The other admin pages uses rails templates: themes, projects, plugins, users, Score Mappings
But if you think this will be a simple pattern that will be easy to maintain, it is fine with me. If you are looking at one to compare it to, the projects are a simple example. With an index, new, edit and delete.

Once this is done it'd be useful to estimate how long it would take to convert one of the other admins pages to the new pattern. Since anytime we have multiple patterns for doing the same thing it introduces tech debt.

Ideally we'd be using a framework to provide the api to access the LibraryInteractives. Like GraphQL, but adding that into an older Rails project like this would be a pain. So sticking with a custom api controller seems the most prudent.

dougmartin:
Ok, I guess I'm confused as I thought in the meeting we decided to do a React based admin.  I'll put this WIP branch aside and redo it in Rails.

dougmartin:
Closing as I rewrote it using standard Rails forms.

PR #509, opened 2020-05-05
WIP: Added managed interactives [#172610374]

dougmartin:


dougmartin:
@scytacki I wasn't able to use a base interactive model class for both `MWInteractive` and `ManageInteractive` as the `constantize` method used when rendering pages returned the base class name which breaks the core page render.  I tried moving the base class code into a module but that caused other issues.  This is the cause of the duplicate code CodeClimate issues.

Now that I have a good test suite passing I'll look again tomorrow to see if I can refactor out the common code.

You'll also notice a few `TODO REVIEW` comments for questions I have.

app/models/managed_interactive.rb
```diff
@@ -0,0 +1,269 @@
...
+      :inherit_click_to_play,
+      :custom_click_to_play,
+      :inherit_full_window,
+      :custom_full_window,
+      :inherit_click_to_play_prompt,
+      :custom_click_to_play_prompt,
+      :inherit_image_url,
+      :custom_image_url
+    ])
+    if library_interactive
```
scytacki: It should always have a library interactive (at least based on code above) so this check could be removed.

dougmartin:
@scytacki I've added tests for import/export of managed interactives and library interactives using the export hash we discussed.  I was also able to use a common module after I removed one method from the module that called a method in the ManagedInteractive subclass and used an attribute with the same name in the MWInteractive.

app/models/managed_interactive.rb
```diff
@@ -0,0 +1,269 @@
...
+        # REVIEW QUESTION: should we call #save here - if we don't and the user imports an activity
+        # that uses the same library interactive more than once it won't find the newly created
+        # library interactive
+      end
+    end
+
+    managed_interactive
+  end
+
+  # REVIEW TODO: is update_labbook_options needed?
```
scytacki: Lets take this out but put a comment in the MwInteractive code indicating this is not supported by the new ManagedInteractive. So we need to migrate this feature to a better solution.

I'll add a story about this in the hopes we can get to it before putting this down.

app/models/managed_interactive.rb
```diff
@@ -0,0 +1,269 @@
...
+  def click_to_play_prompt
+    inherit_click_to_play_prompt ? library_interactive.click_to_play_prompt : custom_click_to_play_prompt
+  end
+
+  def image_url
+    inherit_image_url ? library_interactive.image_url : custom_image_url
+  end
+
+  def enable_learner_state
+    library_interactive ? library_interactive.enable_learner_state : false
```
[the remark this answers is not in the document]
dougmartin: I've updated the model and spec to not assume the library interactive is set.  The user could add a managed interactive to the page but not select a library interactive.

app/models/managed_interactive.rb
dougmartin: Done.

app/models/managed_interactive.rb
```diff
@@ -0,0 +1,171 @@
...
+        # that uses the same library interactive more than once it won't find the newly created
+        # library interactive
+      end
+    end
+
+    managed_interactive
+  end
+
+  # this can't be moved into BaseInteractive as enable_learner_state is defined as a method in
+  # this class and that causes reportable? not to be defined correctly for outside callers
```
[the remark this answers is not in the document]
scytacki: Nevermind, I see this is a module not a class. So ignore this.

app/models/managed_interactive.rb
```diff
@@ -0,0 +1,171 @@
...
+    if imported_library_interactive
+      library_interactive = LibraryInteractive.find_by_export_hash(imported_library_interactive["hash"])
+      if library_interactive
+        managed_interactive.library_interactive = library_interactive
+      else
+        managed_interactive.library_interactive = LibraryInteractive.import(imported_library_interactive["data"])
+
+        # REVIEW QUESTION: should we call #save here - if we don't and the user imports an activity
+        # that uses the same library interactive more than once it won't find the newly created
+        # library interactive
```
scytacki, line 153: Go ahead with the save.  Just to know about it it would be good to look at the second pass code that is handling related linked interactives. And so we don't forget about this save when allow authors to import, add a comment to the top level import.

app/controllers/managed_interactives_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+  before_filter :set_interactive, :except => [:new, :create]
+
+  private
+  def set_interactive
+    @interactive = ManagedInteractive.find(params[:id])
+    set_page
+  end
+
+  # REVIEW: this was copied from the MWInteractivesController but it doesn't look to be used anywere
+  # def create_interactive
```
scytacki, line 13: You can take it out.

dougmartin:
@scytacki I've addressed all the review comments.  I also double checked the managed interactives against the library interactive and mw interactive and found two unproxied attributes that I added.  I also added a test to cover the save of the library and note about the save in the import model.

app/models/managed_interactive.rb
```diff
@@ -33,59 +33,52 @@ class ManagedInteractive < ActiveRecord::Base
...
   def url
+    return nil unless library_interactive
     # BONUS: parse library_interactive.base_url query parameters and merge them with url_fragment query parameters
     "#{library_interactive.base_url}#{url_fragment}"
   end
 
   # getters for inherited attributes
   def aspect_ratio_method
-    inherit_aspect_ratio_method ? library_interactive.aspect_ratio_method : custom_aspect_ratio_method
+    inherit_aspect_ratio_method && library_interactive ? library_interactive.aspect_ratio_method : custom_aspect_ratio_method
```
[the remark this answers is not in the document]
scytacki: We talked about this and decided to leave it as is for now. We might want to tweak it later after we can play with the full experience more and see what happens when authors create interactives but don't give them a library interactive.

PR #510, opened 2020-05-06
Accordion ui

emcelroy:
Related PT story: https://www.pivotaltracker.com/story/show/172593895


emcelroy:
@scytacki I think the 2 issues to fix Code Climate is reporting are things that've existed for a while and aren't something we need to fix now, right?

I'm stuck on what to do about the remaining total coverage issue. It's not clear to me what needs to be done. I'll return to it a little later. If you happen to have any suggestions for that, please let me know.

app/assets/javascripts/common.js
```diff
@@ -55,4 +61,9 @@ $(document).ready(function () {
             });
         }
     });
+
+    // If an Info/Assess block was just added, open that item in the accordion
+    if (window.location.search.match('edit_')) {
+      $('.accordion_embeddables').accordion('option', 'active', -1);
+    }
```
emcelroy, line 68: @scytacki I'm especially wondering what you think about this check that I added. It's how I set it up to open the most recently added info/assessment item in the accordion. It checks for the 'edit_[type]' parameter that was already being set whenever an item is added. I'm wondering though if I should add a new parameter that could be used to specify that a particular item in the accordion should be open on page load. Would that be better?

emcelroy:
I was able to fix the total coverage issue I mentioned earlier. 

The tests I added seem of limited value, but they do provide coverage for the code that Code Climate identified as needing it.

app/controllers/interactive_pages_controller.rb
```diff
@@ -179,12 +179,16 @@ def edit_embeddable_redirect(embeddable)
         param = { :edit_embed_lb => embeddable.id }
       when Embeddable::Xhtml
         param = { :edit_embed_xhtml => embeddable.id }
+      when Embeddable::ExternalScript
```
scytacki, line 182: Is this added so the item will be expanded in the accordion automatically?

Based on the comment on line 193 this should have caused these parameters should have caused the edit dialog to open automatically after the embeddable was added. But that doesn't happen.

@dougmartin you should know about this since you will be working on the popups. 

I think the code that is supposed to handle this is here:
https://github.com/concord-consortium/lara/blob/5e9343a9c0a9163f5cfacc897b448cfcbbba22ed/app/views/shared/_embedded_editor_links.html.haml#L64
And it looks like it is broken because it is expecting the id to be `"#{type}-#{embeddable.id}"` but the id sent above doesn't include the `#{type}-`

It looks like it was broken 5 years ago and we never caught it:
https://github.com/concord-consortium/lara/commit/4658516aa94fee2bb9dc4636c331c66f0965c2cc

That code was checking for the embeddable id here: https://github.com/concord-consortium/lara/commit/4658516aa94fee2bb9dc4636c331c66f0965c2cc#diff-7c27f2754a64715165c8e351a0e65349L8

But it was changed to look for the dash-id here:
https://github.com/concord-consortium/lara/commit/4658516aa94fee2bb9dc4636c331c66f0965c2cc#diff-e5acde65bdae3c0eeb5f79ed8be98d54R25

I'm going to ask about this on slack to see what authors prefer. If they prefer the current behavior of not automatically opening the dialog we should remove this code.

app/assets/javascripts/common.js
scytacki: I think this depends on what we do about automatically opening the edit dialog, that I commented on below.

If you do, do this check, then it seems you should be checking the actual parameter not just the full url. And you should be checking the id value of the parameter, like the code for automatically opening the dialog is trying to do.

Also another case to consider is what happens after someone edits an item. The page will be reloaded in this case. It seems natural to keep the item still expanded after doing this. However the item might not be the last one on the page.

app/views/interactive_pages/edit.html.haml
```diff
@@ -87,17 +87,28 @@
...
                 - tracked = e.respond_to?(:tracked_question) && e.tracked_question
                 - tracked_class = tracked ? 'tracked' : ''
                 .authorable{:id => "embeddable_#{e.id}.#{e.class.to_s}", :class => tracked_class}
-                  = render :partial => "#{e.class.name.underscore.pluralize}/author", :locals => { :embeddable => e, :page => @page, :allow_hide => true }
+                  %h3
+                    - if e.respond_to? :name
+                      - if !e.name.nil? && !e.name.empty?
+                        #{e.name}
+                      - else
+                        #{e.class.human_description.titleize}
```
scytacki, line 106: This is where you'd use `model_name.human` I'm not sure if you'll need titleize in this case or not.

app/controllers/interactive_pages_controller.rb
emcelroy: @scytacki Before I added those lines, an edit_[type] param was appended to the URL only when you added certain assessment item types. I added those lines so that an edit param was appended whenever you add _any_ assessment item type. That way I could then check for the presence of an edit param and open the item that was just added when the page reloads.

app/views/interactive_pages/edit.html.haml
emcelroy: It doesn't matter that much, but that code `window.location.search.match('edit_')` isn't looking for a match in the full URL. It's looking for a match in `window.location.search` which is only the query string part of the URL. Since the edit parameter name varies depending on which assessment type you add, I thought that was the simplest way to just check that an item of any type had been added. I do understand it's not ideal, but the edit param names vary according to which type of assessment item you add and I was wary of adding something to check all possible edit param name possibilities.

Anyway, I think the code you pointed to in _embedded_editor_links.html.haml in a previous comment is a better example to follow though. I'll do something like that instead.

Also FWIW: When you edit an existing item, there isn't a param added to the URL. It's only when you add a new item. Maybe the param should be added when you edit something though?

app/models/embeddable/external_script.rb
```diff
@@ -20,7 +20,7 @@ def self.display_partial
     end
 
     def self.human_description
-      "External Script"
+      "External script"
```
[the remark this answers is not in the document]
emcelroy: Thanks. I'll work on that.

app/controllers/interactive_pages_controller.rb
scytacki: Yeah that is what I figured after digging into this more.  I added a separate tory about fixing the popup behavior: https://www.pivotaltracker.com/story/show/172711110

I asked Doug to work on it,  so we can get it onto staging as we revise the plan for popup authoring.

app/assets/javascripts/common.js
scytacki: @emcelroy somehow your comment below ended up in the wrong thread. I'm copying it here
> It doesn't matter that much, but that code window.location.search.match('edit_') isn't looking for a match in the full URL. It's looking for a match in window.location.search which is only the query string part of the URL. Since the edit parameter name varies depending on which assessment type you add, I thought that was the simplest way to just check that an item of any type had been added. I do understand it's not ideal, but the edit param names vary according to which type of assessment item you add and I was wary of adding something to check all possible edit param name possibilities.

> Anyway, I think the code you pointed to in _embedded_editor_links.html.haml in a previous comment is a better example to follow. I'll do something like that instead.

> Also FWIW: When you edit an existing item, an edit param does not get added to the URL when the page reloads. It's only when you add a new item. Maybe the param should be added when you edit something, too?

scytacki: Sorry I missed the 'search' part of that code above. 

The point of edit param (which Doug is fixing) is that it is supposed to trigger the edit popup to show automatically on page load.  So this would be bad if we do it after closing the editing dialog since that would just open it again.

I think at this point it better not to worry about re-opening the last edited embeddable. It is more important to get this in a place where people can demo it.

PR #511, opened 2020-05-06
Added React-based managed interactives authoring [#172464978]

dougmartin:
Adds structure for the React-based managed interactives authoring.  It adds a new `PageItemAuthoring` namespace to the `window.LARA` namespace to contain all the React-based authoring for page items.  Currently this will only be the managed interactives but this gives a placeholder for migrating other page items.


dougmartin:
Here are two managed interactives and a image interactive.  You can see the edit link is removed for the managed interactives.

image

dougmartin:
Here is the picker moved into the popup dialog using the `react-select` component:

image

dougmartin:
@scytacki I had planned to put the name and url fragment in the options tab as part of the next story (https://www.pivotaltracker.com/story/show/172612620) which is why I left them out of this one.

scytacki:
OK, just note that the story you linked was about existing interactives by that I meant "MWInteractives" . So if you are going to use react for this, you are going to have to re-implement pretty the whole form which seemed a little daunting before, but probably it isn't too bad.

It is fine if we loose the name and url fragment for the new managed interactives at least for a little while. I'm just trying to do this incrementally so that we can release master at anytime.

scytacki:
I guess I won't put this on staging yet since it doesn't really add any new features that other people should look at yet.

PR #512, opened 2020-05-07
Fixed show of edit dialogs automatically after new items are added [#172711110]

dougmartin:


dougmartin:
@scytacki I've updated the PR to handle all the page embeddables.  I manually tested them on the page and added tests for the url param in the interactive pages controller.

PR #521, opened 2020-05-13
Added popup Authoring for Non Library Interactives [#172767822]

dougmartin:
Along with adding popup authoring for MWInteractives this also adds the interactive preview to both MWInteractives and ManagedInteractives.

### DIFF
--- a/lara-typescript/src/page-item-authoring/common/components/interactive-authoring-preview.tsx
+++ b/lara-typescript/src/page-item-authoring/common/components/interactive-authoring-preview.tsx
@@ -1,0 +1,63 @@
+import * as React from "react";
+import { useRef, useState } from "react";
+
+import { InteractiveIframe } from "./interactive-iframe";
+
+export interface IPreviewInteractive {
+  url: string;
+  aspect_ratio_method: string;
+  authored_state: string | object;
+};
+
+interface Props {
+  interactive: IPreviewInteractive;
+}
+
+export const InteractiveAuthoringPreview: React.FC<Props> = ({interactive}) => {
+  const iframe = useRef<HTMLIFrameElement|null>(null);
+  const [authoredState, setAuthoredState] = useState<object|null>(
+    typeof interactive.authored_state === "string"
+      ? JSON.parse(interactive.authored_state || "{}")
+      : interactive.authored_state
+  );
+
+  const handleHeightChange = (newHeight: number | string) => {
+    if (iframe.current) {
+      iframe.current.style.height = `${newHeight}px`;
+    }
+  };
+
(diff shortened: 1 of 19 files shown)


dougmartin:
NOTE: this PR's base is messed up due to a rebase after push.  Once the base commit it merged I'll rebase to master and force push this which will result in a much smaller changeset.

lara-typescript/src/page-item-authoring/common/components/interactive-authoring-preview.tsx
```diff
@@ -0,0 +1,63 @@
...
+      ? JSON.parse(interactive.authored_state || "{}")
+      : interactive.authored_state
+  );
+
+  const handleHeightChange = (newHeight: number | string) => {
+    if (iframe.current) {
+      iframe.current.style.height = `${newHeight}px`;
+    }
+  };
+
```
scytacki, line 29: I'd hope that duplicate code like this can be refactored when working on moving the interactive runtime api into typescript.

scytacki:
@dougmartin So I found a few  issues when looking at it deployed here: https://lara-interactives-test.concordqa.org/

- the iframe interactive edit form is coming up blank 
- the iframe interactive edit dialog doesn't come up when it a new one is added to the page
- the previewed iframes don't handle the height setting api, so they are all a fixed height
- the managed interactive doesn't have a linked_interactive field in the advanced options form

Perhaps there is something broken with this deployment. It should be using this image: https://hub.docker.com/repository/docker/concordconsortium/lara-railslts/builds/106b7fed-e508-4a35-87d3-df7e06009c38

The last two issues could be punted.

dougmartin:
> * the iframe interactive edit form is coming up blank
> * the iframe interactive edit dialog doesn't come up when it a new one is added to the page

I've fixed these two issues - the url was null on new interactives and I had added a message when the url was empty which broke on null.

> * the previewed iframes don't handle the height setting api, so they are all a fixed height

I just checked and it is setting the height.  I tested with Piotr's new interactive and it was sending the wrong height (20 pixels).

> * the managed interactive doesn't have a linked_interactive field in the advanced options form

I'll do this next.

dougmartin:
I've added the linked_interactive_id to the managed interactive edit form.

scytacki:
This works a lot better. I updated a demo activity here:
https://lara-interactives-test.concordqa.org/activities/19995
If you preview the page you can see the open response interactive is setting its height to get taller.
Perhaps there is something about the authoring context that is causing the interactive to not send the right height.

I think I'll merge this as is so the height thing can be worked out in a new PR.

PR #522, opened 2020-05-15
Convert interactive api to TypeScript [#172798312]

dougmartin:


scytacki:
You are probably working on this, but it'd be great if the code for the client (actual interactive) was separate from the code used for implementing the api in LARA.

I think this is slightly different from the plugin code. In that case the plugins are running in the LARA javascript environment, so there is no separation. Putting the interfaces (typing info)  with the implementation makes sense.  In this case there is the postMessage barrier in between.

The tricky part I suppose is that both the LARA implementation and the Interactive client will want to share the same types. My gut would be to put the types in the interactive client code and then have the LARA implementation depend on that package to access the types.  This way the interactive client code is completely self contained.

Another thing to consider here, which doesn't have to be tackled right away, is using TypeDoc for this client like the plugin api does.

dougmartin:
I haven't started the client yet but I was planning on using the `types.ts` file already in the PR as the common file between the LARA api and the client.  The client would not be compiled into the LARA api but would be compiled by webpack as part of the build process and then published to npm (like the plugin api).  The details may change as I get into the implementation today but that is the general idea.

lara-typescript/src/interactive-api/iframe-saver.ts
```diff
@@ -0,0 +1,390 @@
...
+      if (this.successCallback) {
+        return this.successCallback();
+      } else {
+        return IFrameSaver.defaultSuccess();
+      }
+    };
+
+    // Do not send the same state to server over and over again.
+    // "nochange" is a special type of response.
+    // "touch" is an another special type of response which will triger timestamp update only.
```
kswenson, line 235:
```suggestion
    // "touch" is an another special type of response which will trigger timestamp update only.
```

lara-typescript/src/interactive-api-client/README.md
```diff
@@ -0,0 +1,37 @@
+# LARA Interactive API
+
+### This documentation is meant to be used by LARA interactive developers.
+
+#### Setup and webpack configuration
+
+The API is available under `window.LARA.InteractiveAPI` object / namespace once the plugin is initialized by LARA.
```
scytacki, line 7: This sounds like it should be updated. In the client I'd assume this object is available immediately. In other words it shouldn't have to wait until LARA "initializes the interactive".

lara-typescript/src/interactive-api-client/README.md
```diff
@@ -0,0 +1,37 @@
...
+However, if the interactive is implemented using TypeScript, the best way to get type checking and hints in your IDE is to
+install [LARA Plugin API NPM package](https://www.npmjs.com/package/@concord-consortium/lara-interactive-api):
+
+```
+npm i --save-dev @concord-consortium/lara-interactive-api
+```
+
+Then, you need to configure [webpack externals](https://webpack.js.org/configuration/externals/), so webpack does not
+bundle the API code but looks for global `window.LARA.InteractiveAPI` object instead (and do the same for React if the
+plugin uses it).
```
scytacki, line 18: This also looks like it should be updated since we do want the API code bundled in this case.

lara-typescript/src/interactive-api-client/index.ts
```diff
@@ -0,0 +1,344 @@
...
+            } else {
+              this.log("debug", "onHello listener called, no client callback found");
+            }
+
+            if (this.options.supportedFeatures) {
+              this.setSupportedFeatures(this.options.supportedFeatures);
+            }
+          });
+
+          this.addListener("getLearnerUrl", () => {
```
scytacki, line 178: This is deprecated and strongly discouraged. I think it should be removed from the client. If an interactive is upgraded to use the client, then it should also be upgraded to no use deprecated messages like this.

lara-typescript/src/interactive-api-client/index.ts
```diff
@@ -0,0 +1,344 @@
...
+            (initMessage: LaraInteractiveApi.IInitInteractive<AuthoredState, InteractiveState, GlobalInteractiveState>) => {
+              if (this.options.onInitInteractive) {
+                this.log("debug", "initInteractive listener called, calling client callback");
+                this.options.onInitInteractive(initMessage);
+              } else {
+                this.log("debug", "initInteractive listener called, no client callback found");
+              }
+          });
+
+          this.addListener("loadInteractive", (interactiveState: InteractiveState | null) => {
```
scytacki, line 209: This is deprecated. I think it should be removed from the client. If an interactive is upgraded to use the client, then it should also be upgraded to no use deprecated messages like this.

lara-typescript/src/interactive-api-client/index.ts
```diff
@@ -0,0 +1,344 @@
...
+    public setNavigation(options: INavigationOptions) {
+      return this.post("navigation", options);
+    }
+
+    public setAuthoredState(authoredState: AuthoredState) {
+      return this.post("authoredState", authoredState);
+    }
+
+    public getAuthInfo(): Promise<IAuthInfo> {
+      return new Promise<IAuthInfo>((resolve, reject) => {
```
scytacki, line 272: This promise approach is great for these RPC type messages. However I think if we are going to formalize it like this. We need to add an id to `getAuthInfo` message which then is returned in the `authInfo` message. Otherwise it seems possible that this a client which asks for getAuthInfo twice could result in an unresolved promise.

lara-typescript/src/interactive-api-client/index.ts
```diff
@@ -0,0 +1,344 @@
...
+    public setHeight(height: number | string) {
+      return this.post("height", height);
+    }
+
+    public setSupportedFeatures(supportedFeatures: ISupportedFeatures) {
+      return this.post("supportedFeatures", supportedFeatures);
+    }
+
+    public setNavigation(options: INavigationOptions) {
+      return this.post("navigation", options);
```
scytacki, line 264: @pjanik this is the api that supports the ability for an interactive to block page navigation.

lara-typescript/src/interactive-api-client/index.ts
```diff
@@ -0,0 +1,344 @@
...
+          resolve(authInfo);
+          this.removeListener("authInfo");
+        };
+        this.addListener("authInfo", listener);
+        this.post("getAuthInfo");
+      });
+    }
+
+    public getFirebaseJWT(options: IGetFirebaseJwtOptions): Promise<string> {
+      return new Promise<string>((resolve, reject) => {
```
scytacki, line 287: This pattern matches the one above, so just like that one I think this should have an id added to the messages.

scytacki:
Things to test: 
- global state
- interactive save notify

docs/interactive-api-client/README.md
```diff
@@ -0,0 +1,316 @@
+
```
scytacki, line 1: Can we put a comment at the top of this file which indicates it is generated and shouldn't be modified directly. I found this list of ways of adding comments to Markdown: https://stackoverflow.com/questions/4823468/comments-in-markdown
If the html style comments work that seems the best to me.

lara-typescript/src/interactive-api/mock-iframe-phone.ts
```diff
@@ -0,0 +1,238 @@
...
+
+  public _add(msg: any) {
+    this._messages.push(msg);
+  }
+}
+
+let autoConnect = true;
+export const setAutoConnect = (value: boolean) => autoConnect = value;
+
+export class MockIframePhoneManager {
```
scytacki, line 38: Seems like this ought to be moved to the iframe-phone module, so it could be shared by other iframe-phone users.

lara-typescript/src/interactive-api/global-iframe-saver.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+          type: "loadInteractiveGlobal",
+          content: globalState
+        }
+      });
+    });
+
+    it("should send state to LARA server", () => {
+      expect(ajaxParams.url).toBe(saveUrl);
+      expect(ajaxParams.type).toBe("POST");
+      // No idea why raw data needs to be wrapped in array - caused by jQuery or MockAjax?
```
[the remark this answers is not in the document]
dougmartin: I didn't write that comment - that was copied over from the automatic coffeescript conversion.  I'll clean up the comments.

lara-typescript/src/interactive-api-client/use-api-hook.ts
```diff
@@ -0,0 +1,118 @@
...
+  const [ authoredState, setAuthoredState ] = useState<AuthoredState|null>(null);
+  const [ interactiveState, setInteractiveState ] = useState<InteractiveState|null>(null);
+  const [ globalInteractiveState, setGlobalInteractiveState ] = useState<GlobalInteractiveState|null>(null);
+
+  useEffect(() => {
+    client.current = new Client<AuthoredState, InteractiveState, GlobalInteractiveState>({
+      startDisconnected: false,
+      supportedFeatures: hookOptions.supportedFeatures,
+
+      onInitInteractive: (initMessage) => {
```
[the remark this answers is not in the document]
dougmartin: It sounds like you don't want the client hidden in the hook.  In your example the `api` object is really just the client and could be written as:

```
const client = new LaraInteractiveApi.Client<IInteractiveState>({supportedFeatures: {interativeState: true}});
const [interactiveState, setInteractiveState] = useInteractiveState(client);
```

but as you said then I believe interactiveState and setInteractiveState still wouldn't be typed and the user would have to write their own initInteractive handler which greatly reduces the utility of the hook.

We should talk about this over zoom.

PR #524, opened 2020-05-18
Removed separate page-item-authoring js/css files [#172887832]

dougmartin:
This simplifies the lara-typescript build.

Also added a spec test to cover the exports from the PageItemAuthoring object and needed to add identify-obj-proxy to jest config so it could handle the imported react-tabs css file.


dougmartin:
@scytacki I think you should wait on this until after the api changes are merged as this touches the same files.

dougmartin:
@scytacki this has been rebased over the merged interactive api changes and is ready for review.

PR #528, opened 2020-05-21
Fix checkboxes not saving and DRYup customizable options

dougmartin:


scytacki:
codeclimate doesn't show it here, but this PR fixes 28 of the issues it found before.

dougmartin:
>  passing just the string name of the field instead of using formField()

In case it wasn't clear the reason I'm using formField is that it allows us to type check the field name against the JSON interface.  By using:

`const formField = RailsFormField<IManagedInteractive>("managed_interactive");`

then `formField("some_invalid_field_in_IManagedInteractive).name` will not compile.

the type checking is enforced by using the `keyof` operator on the type variable passed to the helper:

```
// returns a type-safe form id/name generator
export const RailsFormField = <T>(name: string) => {
  return (field: keyof T) => ({
    id: `${name}_${field}`,
    name: `${name}[${field}]`,
  });
};
```

PR #530, opened 2020-05-25
Interactive API: support "hint" message

pjanik:
[#172877675]

It might not be super beautiful as it depends on CSS classes, but at least seems to be simple, self-contained, and testable. IframeSaver gets some HTML elements in the constructor (disable button), but it would only spread this logic into more files. I guess that this code might be refactored soon anyway? (when we update change iframe-saver or replace code handling displaying questions).


pjanik:
Thanks, forgot about client. It's updated now.

scytacki:
After discussion with @pjanik we are going to merge this as is. And then update the message format in another PR. This will help with demoing.

PR #531, opened 2020-05-26
update all locked dependencies

scytacki:
I was having build issues with fsevents which was locked at 1.2.7
To fix it, I removed my node_modules and package-lock.json and then ran npm install
The result is several updated dependencies.
If the tests pass, this seems like a good thing do.


scytacki:
This seems to have broken the tests:
https://travis-ci.org/github/concord-consortium/lara/builds/691175117
It looks like this is due to more strict type checking.

PR #532, opened 2020-05-26
Upgraded TypeScript to 3.9.3 [#172919264]

dougmartin:
Needed to fix a couple of null/undefined issues.

Also added version controlled settings.json to point to the workspace typescript version.


dougmartin:
@scytacki this depends on the page-item-authoring pending PR (as I wanted to ensure that also compiled in thew new TS version).

PR #533, opened 2020-05-26
Fixed ordering of type variables in readme example

dougmartin:


dougmartin:
@scytacki this should probably wait until all other PRs are in so the readme can be regenerated.

PR #536, opened 2020-05-28
Rename Managed Interactive to Library Interactive in the UI

scytacki:
I also took the time to clean up the duplicate use of localized model names and couple of custom name mappings
It doesn't look like this will break anything, but lets see when the full tests run

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ To use SSO with the portal you need to make sure your `PORTAL_HOST` is set corre
 
 The rake task above assumes your local lara instance is running at http://app.lara.docker
 If it is running somewhere else, then you can log into the portal as an admin and edit
-the Auth Client created by the rake task. 
+the Auth Client created by the rake task.
 
 ## Users and administration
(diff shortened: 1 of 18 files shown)


README.md
```diff
@@ -130,7 +130,8 @@ To support new Embeddables:
 * Add a view directory at `app/views/embeddable/<embeddable_name>`
 * Provide a `_lightweight.html.haml` partial within that view directory (for showing the Embeddable within an InteractivePage)
 * Provide a `_author.html.haml` partial as well (for editing the Embeddable)
-* Add the Embeddable's name as a string to the the `Embeddable::Types` constant in `app/models/embeddable.rb`.
+* Add the Embeddable to the the `Embeddable::Types` constant in `app/models/embeddable.rb`.
```
kswenson, line 133:
```suggestion
* Add the Embeddable to the `Embeddable::Types` constant in `app/models/embeddable.rb`.
```

app/models/embeddable.rb
```diff
@@ -1,40 +1,41 @@
...
-    Embeddable::ExternalScript => "External Script",
-  }
-
-  InteractiveTypes = {
-    MwInteractive      => "Iframe Interactive",
-    ManagedInteractive => "Managed Interactive",
-    ImageInteractive   => "Image",
-    VideoInteractive   => "Video"
-  }
+  # The Order here determins the order in the authoring selection menu
```
kswenson, line 2:
```suggestion
  # The Order here determines the order in the authoring selection menu
```

scytacki:
I'm going to push this through because we might be demoing it tomorrow.
I'm also going to back port it to LARA 1.51.x so it would be part of that release.

PR #538, opened 2020-05-29
Refactor Interactive API client

pjanik:
[#173082951]

This is an initial PR so I can get feedback on the approach. Main changes:

- I reduced `Client` to very basic features. 
- It manages some interactive state (interactive, authoring, global). Later, I'm explaining why/
- The user-facing code is defined in `api.ts` and `hooks.ts` files. They are just set of functions, that can be imported separately. It also means that these files can be easily divided in the future if they get too big. Everything is independent of each other.
- I haven't converted tests yet.
- I tried to finish all that today, so probably there might be many things to polish.

I've also refactored question-interactives repo to use this new client:
https://github.com/concord-consortium/question-interactives/pull/15/files
So, you can see how this can be used in practice. For example, take a look at `useRequiredQuestion` or `useHint` there.

This new approach looks probably more complicated (and it is due to state management). But because of that interactives development should be easier:

1. Hooks can be imported multiple times within one app, by any component, and some states will be automatically synchronized. For example, when two components import and use `useInteractiveState` and one of them updates it, the other will be notified too. That's why I added `ManagedState` class and used EventEmitter. This is opinionated approach. This synchronization could be also implemented in the interactive itself, e.g. using React context, Redux, or any other state manager. But it'd make interactive development more complex and more error-prone (sooner or later somebody wouldn't synchronize interactive states).

2. Since the client manages some state, it can automatically respond to some messages from LARA, e.g. "getInteractiveState". Interactive doesn't have to care about it anymore. This is helpful, as it lets us "decentralize" talking to LARA (we don't have to have one central point in the interactive that initializes interactive-LARA communication). 

3. Since the client is no longer a class, but set of independent functions, interactive can just import what it needs like: `import { setHint } from "@concord-consortium/lara-interactive-api"`. Since we don't have to manage Client state anymore, these functions can be used directly in React component. We don't need hooks for that. That's why number of hooks (or hook-based features) is reduced and we don't have anything like `useSetHint` here. Before these hook methods were just calling equivalent client functions (after checking client is available).


lara-typescript/src/interactive-api-client/hooks.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  useEffect(() => {
+    // useEffect callback can't be async.
+    (async () => {
+      setInitMessage(await client.getInitInteractiveMessage());
+    })();
+  }, []);
+
+  return initMessage;
+};
```
kswenson, line 86: One other hook I can see being useful is a `useSupportedFeatures()` hook which encapsulates the necessary `useEffect()` logic. I added a similar comment/question in the client PR.

lara-typescript/src/interactive-api-client/hooks.ts
```diff
@@ -0,0 +1,86 @@
+import { useEffect, useState } from "react";
+import { IInitInteractive } from "./types";
+import * as client from "./api";
+
+export const useInteractiveState = <InteractiveState>() => {
+  const [ interactiveState, setInteractiveState ] = useState<InteractiveState | null>(client.getInteractiveState<InteractiveState>());
```
scytacki, line 6: I'm not sure if this is needed, but it might be. 

Since the client is actually maintaining the state there isn't really a need to store it in react's state.
So instead the line above could be:
`const [ , updateState ] = useState(0);`
So then the return from this hook would be `{ interactiveState: client.managedState.interactiveState, ...}`
The default value of `0` is so it can be incremented each time a change comes in.
That approach was taken from here: https://stackoverflow.com/a/55862077

Since React doesn't seem to do a deep compare of the state, as I noted below, calling setInteractiveState might not trigger the expected re-render.

But at the same time storing the actual state variable would might make it easier to use the React developer tools to see what is going on. But I think then there'd have to be a least a shallow copy of the interactiveState made on each update.

lara-typescript/src/interactive-api-client/hooks.ts
```diff
@@ -0,0 +1,86 @@
...
+      client.removeInteractiveStateListener<InteractiveState>(handleStateUpdate);
+    };
+  }, []);
+
+  const handleSetInteractiveState = (state: InteractiveState) => {
+    setInteractiveState(state);
+    client.setInteractiveState<InteractiveState>(state);
+  };
+
+  return { interactiveState, setInteractiveState: handleSetInteractiveState };
```
scytacki, line 25: Maybe it would be better to follow React's useState convention and return an array here instead of an object. That way the calling code can decide on its own variable and function names. But I guess unlike useState it doesn't make sense to use this hook multiple times in the same component.

lara-typescript/tslint.json
```diff
@@ -19,6 +19,7 @@
     "ordered-imports": false,
     "trailing-comma": false,
     "variable-name": [true, "ban-keywords", "check-format", "allow-pascal-case", "allow-leading-underscore"],
-    "no-namespace": false
+    "no-namespace": false,
+    "max-line-length": false
```
scytacki, line 23: I do like having a max line length. It helps to make it easier to compare files side by side. 

But I think I might be the only one that prefers it. :) 
I think @dougmartin didn't really want a max line length.  I can't remember what @kswenson thought about it.

lara-typescript/src/interactive-api-client/hooks.ts
```diff
@@ -0,0 +1,86 @@
...
+import * as client from "./api";
+
+export const useInteractiveState = <InteractiveState>() => {
+  const [ interactiveState, setInteractiveState ] = useState<InteractiveState | null>(client.getInteractiveState<InteractiveState>());
+
+  useEffect(() => {
+    // Setup client event listeners. They will ensure that another instance of this hook (or anything else
+    // using client directly) makes changes to interactive state, this hook will receive these changes.
+    const handleStateUpdate = (newState: InteractiveState) => {
+      setInteractiveState(newState);
```
scytacki, line 12: If I understand this right, the newState passed by the the client could be the same object just with different properties.  So then React will use `Object.is` and not think the object has changed. So then it won't trigger a re-render.

If the updateState approach was used above then this would be turned into:
`updateState(prev => prev + 1)`

lara-typescript/src/interactive-api-client/hooks.ts
pjanik: Hm, so the counter approach would force re-render each time someone calls `setInteractiveState`. I can see two possible issues then. If we call `client.setInteractiveState` and return `client.managedState.interactiveState`, we implicitly assume that `client.setInteractiveState` **immediately** updates `client.managedState.interactiveState`. Of course, that's how things work in this particular case, but generally, it might not always be true (if there's some asyc operation, etc.). But this is more of a theoretical issue.

 > But I think then there'd have to be a least a shallow copy of the interactiveState made on each update.

Yes, I'm assuming that user of this hook would always do `Object.assign({}, interactiveState, {newProperty: 123})` or use something similar (jQuery extend, ImmutableJS, etc.). I'd say it's developer responsibility. Right now we're just consistent with `useState` behavior - if `useState` triggers re-render when the same object is passed, the developer is lucky. ;) If not, this is how React works (I believe it works that way and I always use Object.assign). 

Having said that, I don't think that forced re-renders would break things in practice or cause some performance issues in our case. But I think we'd add a pattern (forced updates) which is not fully consistent with react world/rules/hooks.

lara-typescript/src/interactive-api-client/hooks.ts
pjanik: `useSupportedFeatures` actually is just a function that interactive calls and never waits for any results. So, I don't think we need a hook for it.

Actually, this brings another topic - do we really need it? So far LARA seems to accept multiple calls and in any order. But it might be a bit confusing for a developer - when this should be called, can it be called multiple times, etc etc. We mostly use `interactiveState`, `authoredState` and `aspectRatio`. Does LARA really need to know whether interactive supports interactive and authored state? Can't it just pass the sate if it's present? And always listen for incoming states. Aspect ratio could be moved to a separate function like `setHeight`. If there're some other features like global or linked state that require this upfront information for some reason, maybe we could split it into separate functions, so it's easier for hooks to encapsulate everything?

lara-typescript/src/interactive-api-client/hooks.ts
pjanik: Right, I think array is mostly used to let developers use the same hook multiple times. It's not the case here. I think object works a bit better with types and it's easier to extend in the future. But I can change it if you think it's cleaner to follow `useState` convetion.

lara-typescript/tslint.json
pjanik: I'm 50-50. I usually don't mind max-line-length (although I prefer a bit more relaxed values than 120 ;)). But here we have so many generic types that many lines get pretty long and it's not that easy to divide them nicely. Let me know, I can re-enable it.

lara-typescript/src/interactive-api-client/hooks.ts
pjanik: > So then React will use Object.is and not think the object has changed. So then it won't trigger a re-render.

Right, although I think this is developer's fault and personally I wouldn't try to fix it. The same way React doesn't try to fix it in `useState`. That's why I was using `Object.assign` in the interactives I was writing (even before this hook was added). But if others think that super-bulletproof approach is better, let me know.

lara-typescript/src/interactive-api-client/types.ts
```diff
@@ -173,6 +173,11 @@ export type ClientMessage = DeprecatedIFrameSaverClientMessage |
                             IFrameSaverClientMessage |
                             GlobalIFrameSaverClientMessage;
 
+export type ClientEvents = "interactiveStateUpdated" |
```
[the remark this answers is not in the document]
pjanik: Fixed, thanks.

lara-typescript/src/interactive-api-client/api.ts
```diff
@@ -0,0 +1,131 @@
...
+    const request: IGetFirebaseJwtRequest = {
+      requestId,
+      ...options
+    };
+    client.addListener("firebaseJWT", listener, requestId);
+    client.post("getFirebaseJWT", request);
+  });
+};
+
+// tslint:disable-next-line:max-line-length
```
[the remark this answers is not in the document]
pjanik: The rule is enabled again.

lara-typescript/src/interactive-api-client/hooks.ts
scytacki: Yes this makes sense.

Perhaps then, the setInteractiveState method should also accept a function (like setState) so the developer can do:
`setInteractiveState(prev => {...prev, newProperty: 123})`
Which would also be a more safe pattern if the setInteractiveState method is called inside of a callback. Since the interactiveState that is closed around would be frozen at the time of the last render.

docs/lara-plugin-api/README.md
```diff
@@ -258,12 +252,6 @@ ___
 
 *Defined in [popup.ts:86](../../lara-typescript/src/plugin-api/popup.ts#L86)*
 
-Ask LARA to add a new popup window.
-
-Note that many options closely resemble jQuery UI dialog options which is used under the hood. You can refer to jQuery UI API docs in many cases: [https://api.jqueryui.com/dialog](https://api.jqueryui.com/dialog) Only `content` is required. Other options have reasonable default values (subject to change, so if you expect particular behaviour, provide necessary options explicitly).
-
-React warning: if you use React to render content, remember to call `ReactDOM.unmountComponentAtNode(content)` in `onRemove` handler.
-
```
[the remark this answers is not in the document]
pjanik: Right, I didn't notice it. I tried to run docs generation again and it's still the same. Maybe it's related to the recent dependencies update (e.g. TypeScript 3.9)? @dougmartin, does doc generation work for you? Doesn't it remove typedocs comments like here?

When I updated Typedoc and markdown plugin, it actually helped. But changed a bit structure of generated docs. So, not sure I should commit it here.

Also, a related question - when I was working on plugin-api, there's specific setup of files and directories, so the generated docs are showing **only** members exported in NPM package. In case of LARA Interactive API, it seems that docs are generated for everything, even the internal helpers. Do we want to work on that, so these files are not included in the docs? If so, it should be separate PR I think. But wondering if it makes sense, as probably most devs won't read that anyway (but use typing in IDE).

pjanik:
Thanks! I've fixed smaller issues and replied to some comments where solution is not clear. ;)

lara-typescript/src/interactive-api-client/hooks.ts
scytacki: The general intention was so the interactive could say what it supports and then the authoring and runtime system could adjust without needing to maintain external settings for the interactive. But in practice we haven't be able to implement it this way.  LARA now has settings to say if the interactive supports interactiveState, and whether it supports authoring.  The interactive state setting is used to determine whether it should block the page navigation to make sure the state of the interactive is saved. The authored state setting is used to determine if the authoring system should show the interactive in authored mode or it should just show some generic properties to customize the interactive.

But again these are settings in LARA instead of using the values from the supportedFeatures. This isn't great because it means when a developer changes an interactive we have to find all references to it and update their LARA settings. That is where the supportedFeatures came from. 

A new api we are adding is the ability for the interactive to support custom messages. This so plugins or interactives can send messages to the interactive. And that will be used by the teacher edition plugin to tell the multiple choice interactive to show its teacher-edition:correct overlay or teacher-edition:distractor overlay.  In order for the teacher edition authoring to function as it does now, it will need to know if the interactive supports these correct and distractor overlays. So the way we put this into the interactive types is through the supportedFeatures.  The interactive would say which of these custom messages it handles. The plugin authoring would have access to this and so then the teacher edition authoring could know if it is should show its correct and distractor tabs. 

However, all of this is really static meta-data about the interactive. It isn't something that will change while the interactive is running.  So if each interactive provided meta data next to its html page, then we could improve this approach. This is what we tried to do with the LARA plugins. 

However I don't think this approach with the plugins worked out so well. In particular it still requires an admin to go in and refresh the meta data. It has the benefit that the author can't set the meta data to be something totally wrong, and we don't need a UI for all of this metadata.  But it does mean the meta data can become stale because doesn't automatically get updated.

Anyhow, sorry I don't have a concrete answer, but hopefully having more people aware of the details will help us work out a solution.

lara-typescript/src/interactive-api-client/hooks.ts
scytacki: I think it is fine using the object approach, I just wanted to raise the point.

lara-typescript/src/interactive-api-client/hooks.ts
scytacki: I think it is fine to keep it as is, but perhaps a comment in the client's 'setInteractiveState' definition about it would be useful. Since the handleStateUpdate call might be triggered by someone using the client directly to send a new state to LARA. If they aren't in the React part of the code, they might not realize that they need to copy the state in order to cause any React listeners to update.

scytacki:
@pjanik what about the issue with the jsdoc comments disappearing from the plugin api documentation: https://github.com/concord-consortium/lara/pull/538#discussion_r433020968

Perhaps that is unrelated to this PR, the change to the generated files is in the PR.

scytacki:
> @pjanik what about the issue with the jsdoc comments disappearing from the plugin api documentation: [#538 (comment)](https://github.com/concord-consortium/lara/pull/538#discussion_r433020968)
> 
> Perhaps that is unrelated to this PR, the change to the generated files is in the PR.

Oops I see you responded, sorry. I'll delete these two comments.

docs/lara-plugin-api/README.md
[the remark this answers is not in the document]
scytacki: I think improving the documentation for the interactive api should be its own story and PR. We can deal with the mixing of internal types with external ones.  I do think it is important to do since I think it would be good to have an automatically generated reference which can be linked to.

Moving the fix for the lara plugin api docs to a separate PR seems also fine to me.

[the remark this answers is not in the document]
scytacki: I added a story for this documentation fix: https://www.pivotaltracker.com/story/show/173116252

scytacki:
I approved this. 

The only unresolved comment I had was about accepting a form like:
`setInteractiveState(prev => {...prev, newField: "hello"}`

Also I think @pjanik said there are some tests that need updating. I'm assuming that should happen before merging this in. Codeclimate is not complaining, but I'd guess that is because it isn't including coverage of the lara-typescript folder.

lara-typescript/src/interactive-api-client/hooks.ts
pjanik: Good point, thanks. I've added comments. I've also added "freezing" of this state in the last comment + Readonly TypeScript modifier. So, it should be impossible to change this state.

pjanik:
@scytacki, latest commits add updateState function handling and freezing of state objects.
Yes, tests are missing, I need to convert existing tests and possibly add some more.

lara-typescript/src/interactive-api-client/hooks.ts
pjanik: Thanks, actually I didn't know `setState` accepts it. I've added it in the last commit. 
btw, I would have to think more about it, but my first impression is that it's just for convenience. Not sure if I can come up with cases where it'd be safer than a regular update. I think that hook value is always equal to the value that is in functional component, as hook's state update means re-render.

lara-typescript/src/interactive-api-client/api.ts
```diff
@@ -22,6 +22,23 @@ export const getInteractiveState = <InteractiveState>(): InteractiveState | null
...
 
+/**
+ * Note that state will become frozen and should never be mutated.
+ * Each time you update state, make sure that a new object is passed.
+ *
+ * Good:
+ * ```
+ * setInteractiveState(Object.assign({}, previousState, {someProperty: "a new value"}));
+ * // or
+ * setInteractiveState({...previousState, {someProperty: "a new value"}});
```
scytacki, line 33:
```suggestion
 * setInteractiveState({...previousState, someProperty: "a new value"});
```

scytacki:
The comments and updateFunction support look great, just found the issue with the spread syntax.

lara-typescript/tslint.json
kswenson: I don't feel particularly strongly about it. If it's enabled, I wouldn't want it to be any smaller than 120, but having some (reasonably expansive) limit probably isn't a bad idea.

lara-typescript/src/interactive-api-client/hooks.ts
kswenson: The `setState` function returned by `useState` is asynchronous in the same way that the class-based `setState` is, so if the state is updated between the time that the hook runs and the state is actually set, then the update would use the stale value rather than the most current value. I've seen examples with a button that increments a counter where if you press the button rapidly some clicks are "lost".

pjanik: I see. I'm wondering how to make it a problem in practice though.

Well, we can do:
```
setInteractiveState(Object.assign({}, intState, {foo: 1}));
setInteractiveState(Object.assign({}, intState, {bar: 2}));
```
And this will obviously fail. But the functional update I've added won't help. I guess React's implementation is smarter about this.

Also, I'm wondering what would be a more real-life scenario. Even if the interactive state is updated in some callback, I guess the previous state will be the same in the last render call and in the hook body (which is executed during... render). So, again, React's implementation must be smarter. 

The only case I can think of, where things can fail without obvious developer's fault, is when React is batching state updates. Then a few state updates can be executed at the same time. Each will have the same previous state, without new properties added by these calls. So, my fix won't help in this case either.

I'll think about some better implementation. I think I could pass `client.managedState.interactiveState` as a previous state to the update function. As `client.managedState.interactiveState` always gets updated synchronously, so it should be always more up to date than internal `setState` copy kept by React.

Thanks, I haven't thought about that before. :wink:

pjanik: Since we went deep here - if someone is interested, here's a test that actually shows the difference between non-functional and functional update:

https://github.com/concord-consortium/lara/blob/46e4baff5c2350ba7f6881713da6ba2de1416910/lara-typescript/src/interactive-api-client/hooks.spec.ts#L73-L93

And if I revert this commit/changes: https://github.com/concord-consortium/lara/pull/538/commits/9a376ed9a2011370eee060b4b2043278761c0472, this test of the functional updates fails.

PR #545, opened 2020-06-03
Interactive API: hook tests

pjanik:
This PR:
- updates React and ReactDOM (it was necessary for hook testing library to work)
- adds TSLint "rules of hooks"
- adds missing hook tests

Screenshot 2020-06-03 at 14 11 27

### DIFF
--- a/lara-typescript/package.json
+++ b/lara-typescript/package.json
@@ -39,7 +39,9 @@
     "url": "https://github.com/concord-consortium/lara/issues"
   },
   "jest": {
-    "setupFilesAfterEnv": ["<rootDir>src/setupTests.ts"],
+    "setupFilesAfterEnv": [
+      "<rootDir>src/setupTests.ts"
+    ],
     "transform": {
       ".(ts|tsx)": "ts-jest"
     },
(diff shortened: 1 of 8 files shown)


lara-typescript/package.json
```diff
@@ -86,8 +89,9 @@
     "node-sass": "^4.12.0",
     "npm-run-all": "^4.1.5",
     "raw-loader": "^1.0.0",
-    "react": "^16.8.6",
-    "react-dom": "^16.8.6",
+    "react": "^16.13.1",
+    "react-dom": "^16.13.1",
```
scytacki, line 93: There is also the react version provided by LARA which is older.
If we bump the version here, then it will be easy for developers to use react features that aren't compatible with that older version in LARA.

Doug thought it wouldn't be easy to update the older version in LARA because all of the ITSI authoring page is written with this older version of React (and probably coffeescript). 

I'm not sure how to handle this.

lara-typescript/src/interactive-api-client/hooks.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+      result.current.setInteractiveState(Object.assign({}, result.current.interactiveState, {c: 3 }));
+    });
+
+    expect(result.current.interactiveState).toEqual({a: 1, c: 3});
+    expect(getClient().managedState.interactiveState).toEqual({a: 1, c: 3});
+
+    act(() => {
+      // If state is updated incrementally, functional update should be used. No property will be lost here.
+      result.current.setInteractiveState((prevState: any) => Object.assign({}, prevState, {d: 4 }));
+      result.current.setInteractiveState((prevState: any) => Object.assign({}, prevState, {e: 5 }));
```
scytacki, line 89: I notice you aren't using the spread syntax. Is there reason not to?

lara-typescript/package.json
scytacki: Perhaps this version bump of React won't break the itsi pages. Based on the version numbering that should be true. So perhaps that is the best approach. Bump it for all of LARA and then we can do some smoke testing of the ITSI authoring.

cc @dougmartin

pjanik:
@scytacki, I've updated LARA React to 16.13.1 and used spread operator in hook tests.

PR #546, opened 2020-06-03
LARA Typescript reorganized + improved docs

pjanik:
This PR got quite big. Note that auto-generated documentation is not included in this PR, so the diff is still readable. I'll open another PR with auto-generated Typescript docs, so they can be reviewed separately.

I've updated Typedoc to 0.17.0-3 (some pre-release) that adds "library" mode. It's great, as it was exactly what I was missing before. Now, we can point TypeDoc at a single file and it'll generate documentation **only** for all the **exported members**. This is what we want for documentation of LARA Plugin API and LARA Interactive API. We don't want to include internal helpers in our docs.

Before it wasn't possible. I had to divide LARA Plugin API into multiple directories and generate documentation just for one folder. This resulted in a bit awkward directory structure and dependencies. This structure was driven by doc-generation requirements. 

Since we no longer have to do it, I thought it makes sense to go ahead and clean up things. That's why you can notice that many files have been moved and some naming has been changed.

The general rule is that every directory in lara-typescript folder has its own `index.ts`. This file exports members that are later available under `LARA.<lib-name>` namespace. It's all defined in top-level `index.ts`.

1. I deleted `LARA.InternalAPI` as Doug added new features under other namespaces (PageAuthoring and InteractiveAPI) and I think it makes sense to keep separate features in separate directories and export them separately. So, `InternalAPI` is now replaced by `LARA.Events` and `LARA.Plugins` namespaces. Both are used by LARA haml files and/or coffeescript files.

2. I deleted `lib` directory. This was implementation of `LARA.Events` and `LARA.Plugins`. Now they live in their own directories. 

3. I renamed `interactive-api` dir to `interactive-api-parent`, so it's clear why it's separate from `interactive-api-client`. 

4. I updated all the README files. Sections that are not relevant are deleted (e.g. about docs generation and specific file structure). Interactive API is updated.

5. Common Typedoc options now live in `typedoc.json`. Line numbers are disabled (note that it means that source links are totally gone, it doesn't seem possible to link at least to a file). I've also simplified a bit build process and I'm generating docs directly in the final directory, without separate copy step (fewer package.json scripts).

Generally, it's a bit scary, as it's quite a lot of changes, but I think it's worth cleaning up these things before everything gets even bigger.


scytacki:
@pjanik looks like this broke some plugin specs.

lara-typescript/src/interactive-api-client/README.md
```diff
@@ -14,127 +13,55 @@ First install the package:
...
+}
 ...
-
 <button onClick={handleButtonClick}>{interactiveState.numButtonPresses}</button>
 ```
 
-Calling the `setAuthoredState`, `setInteractiveState` and `setGlobalInteractiveState` methods returned by the hook will both update the state held inside the hook and also send the message to Lara.  This allows you to use the update the interactive state and use it in your render result.  You also do not have to supply an `onGetInteractiveState` callback as the hook handles that for you using the current local hook state.
\ No newline at end of file
+Calling the `setInteractiveState` method returned by the hook will both update the state held inside the hook and 
+also send the message to Lara. This allows you to use the update the interactive state and use it in your render result.
```
scytacki, line 63: This doesn't read right. Perhaps:
"This means that calling setInteractiveState will trigger a new React render of your components that use the `useInteractiveState` hook"

pjanik: Thanks, fixed in the last commit.

pjanik:
> This shouldn't break the existing v2 or v3 plugins, right?

It shouldn't. But I think some smoke testing would be definitely useful.

PR #547, opened 2020-06-03
Updated auto-generated docs

pjanik:
This should be part of https://github.com/concord-consortium/lara/pull/546 but I split it, so it's easier to review both PRs.


pjanik:
> There is one thing that seems broken though. It isn't very obvious how to get from the Readme to the generated docs and then back again. Perhaps this is bug in typedoc?

Not sure if it's a bug or feature, but yes, I've noticed this change too.

I pushed a commit which adds a big "API documentation" link at the beginning.

scytacki:
It seems like the markdown formatting of that new link isn't right:
https://github.com/concord-consortium/lara/blob/af4dfdf4de3927c183e1cbf9f6594ab893aedb71/docs/interactive-api-client/README.md

scytacki:
I'm going to approve this now, and go ahead and merge it once the formatting is fixed.

PR #550, opened 2020-06-04
Update hint message format

pjanik:
@scytacki, this is something you mentioned some time go in one of the PR reviews.
We'd need to update the client packge in question-interactives repo, but no other changes will be necessary there.

BTW, maybe it's just my impression but I was a bit lost in types - I didn't know if I should add `IHint`, `IHintOptions`, `IHintRequest`, etc. :wink: Tiny thing, but maybe some patterns in types naming could be useful long term. Or they're there already, but I might have not noticed.

### DIFF
--- a/app/assets/javascripts/lara-typescript.js
+++ b/app/assets/javascripts/lara-typescript.js
@@ -25899,16 +25899,16 @@ var IFrameSaver = /** @class */ (function () {
             _this.$iframe.data("height", height);
             _this.$iframe.trigger("sizeUpdate");
         });
-        this.addListener("hint", function (hint) {
+        this.addListener("hint", function (hintRequest) {
             var $container = _this.$iframe.closest(".embeddable-container");
             var $helpIcon = $container.find(".help-icon");
-            if (hint) {
+            if (hintRequest.text) {
                 $helpIcon.removeClass("hidden");
             }
             else {
                 $container.find(".help-icon").addClass("hidden");
             }
-            $container.find(".help-content .text").text(hint || "");
+            $container.find(".help-content .text").text(hintRequest.text || "");
         });
         this.addListener("supportedFeatures", function (info) {
             if (info.features && info.features.aspectRatio) {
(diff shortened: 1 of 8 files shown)


scytacki:
@dougmartin can you take a look at this? 

My interpretation is:
`IFoo` - interface for a simple request that doesn't require a response. Examples: ISupportedFeatures, IAuthoringMetadata, ....
`IBarOptions` - for requests that have an associated response, this is the set of options an interactive using the client passes to the exported function. So this is the 'public' interface. Examples are IGetFirebaseJwtOptions, 
`IBarRequest` - this is what is actually sent over the 'wire'. If the request is one that has a response then this is usually the merge of I*Options plus IBaseRequestResponse. This is a private interface used inside of the client. However it is useful to expose in the docs in order to specify the over the wire message format.

Below are some types I found that break this pattern.

Breaks pattern, but doesn't look like a bug:
- INavigationOptions: this doesn't have a response but it uses the I*Options naming.

Breaks pattern, and looks like a bug. There are corresponding I*Options types, but they used by the client functions:
- IGetInteractiveListRequest: this requires a response, but this is the interface used in the public
- IGetLibraryInteractiveListRequest: this requires a response, but is the public interface
- IGetInteractiveSnapshotRequest: this requires a response, but is the public interface

Perhaps it would be easier if we simplified the pattern from the perspective of the client user. So either it is always IFoo or always IFooOptions. That would be regardless of whether the function has a response. Any opinions?

cc @kswenson

dougmartin:
I had a long response but I edited down to we should think about using this for two-way communication:

1. `I<ObjectName>RequestOptions` which merges with `IBaseRequestResponse` to form `I<ObjectName>Request`.  The `I<ObjectName>RequestOptions` is used as the type for the parameter to the callable client or server function.
2. `I<ObjectName>ResponseOptions` which merges with `IBaseRequestResponse` to form `I<ObjectName>Response`.  `I<ObjectName>ResponseOptions` is used as the parameter on the responder message client or server function.

and just `I<Object>` for 1-way communication.

So we should rename `INavigationOptions` to `INavigation`  and keep the fix the stub functions that use `I<ObjectName>Request` parameter type to use `I<ObjectName>RequestOptions`.

dougmartin:
If you want to merge this as is I can go in either tonight or tomorrow morning and fix the types.

kswenson:
I think a consistent naming convention is a good idea for all the obvious reasons. I like the consistency of Doug's suggestion, but one quibble is that "Options" doesn't seem like the right noun for information coming back from LARA to the client. I'm not sure what the right noun should be, but perhaps something like `I<ObjectName>ResponseData` or `I<ObjectName>ResponseInfo` although something less generic would be preferable if we can come up with it.

PR #554, opened 2020-06-08
Interactive API: automatically parse strings in LARA init message

pjanik:
LARA sometimes passes authored and interactive states as JS objects, sometimes as JSON strings. Also, the portal-report is not consistent. Handle that in the Client.

### DIFF
--- a/lara-typescript/src/interactive-api-client/client.ts
+++ b/lara-typescript/src/interactive-api-client/client.ts
@@ -14,6 +14,10 @@ interface IListenerMap {
   [key: string]: IRequestCallback[];
 }
 
+const parseJSONIfString = (data: any) => {
+  return typeof data === "string" ?  JSON.parse(data) : data;
+};
+
 const phoneInitialized = () => iframePhone.getIFrameEndpoint().getListenerNames().length > 0;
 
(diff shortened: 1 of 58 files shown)


lara-typescript/src/interactive-api-client/client.ts
```diff
@@ -14,6 +14,10 @@ interface IListenerMap {
   [key: string]: IRequestCallback[];
 }
 
+const parseJSONIfString = (data: any) => {
+  return typeof data === "string" ?  JSON.parse(data) : data;
```
kswenson, line 18: Do we need to worry about potential parse errors here?

pjanik: Not sure. I think it's pretty unlikely to happen. But if it does, we can't initialize interactive anyway. So, the only difference is whether JSON parse error is thrown or not. And whether we display some message to user. I think this error should NOT be handled gracefully, as this is not something that is expected. So, Rollbar should report it. We could add a message for users so it's easier to report issues manually too. But I would need to be an (annoying) alert I guess, as we can't do much more from iframe level without having access to interactive itself.

PR #558, opened 2020-06-10
Disable logging to file

pjanik:


scytacki:
After approving I saw the tests were failing due to all of the log messages.
My change above ought to fix this...

PR #563, opened 2020-06-12
Hide hidden questions

pjanik:
[#173038370]

It should fix unwanted padding at the top of assessment block, caused by hidden questions (in fact plugins): https://authoring.staging.concord.org/activities/20634/pages/307494/

### DIFF
--- a/app/assets/stylesheets/partials/_runtime-interactives.scss
+++ b/app/assets/stylesheets/partials/_runtime-interactives.scss
@@ -66,6 +66,9 @@
     font-style: italic;
     font-size: 0.9em;
   }
+  &.hidden {
+    display: none;
+  }
 }
 
 .challenge {


scytacki:
@pjanik can you explain why this broke? It was working before. If the code that broke it was written by someone else it would be fine to hand this question off to them.

pjanik:
Are we sure there was no unwanted padding before? 
This problem is on production too: https://authoring.concord.org/activities/10872/pages/136417/preview

There are two things:
- `.question` has bottom-margin 20px
- `.question .hidden` was not actually hidden 

I checked bottom-margin CSS line, but it was added 6 years ago by Noah. So it's not a recent change. Why hidden question is not hidden anymore - I don't really know. I bet some `.hidden { display: none; }` got removed or moved. This is a bit hard to track, as it could have been defined anywhere. Ethan was making quite a lot of changes to authoring CSS, so maybe it's related (as some changes sneaked to runtime too).

scytacki:
I'm just worried that this problem is more complex than this, so this simple change might just be masking the problem.

Also this style was added to an interactive style sheet but really it is being applied to an all embeddables right?

I know that this has been a problem before. I know that there is complex code which does something special when a plugin is wrapping an interactive. The wrapping plugin is added to the list of embeddables in the database, but I think it is not actually rendered during the runtime via this logic. 

Here are some stories about this: 
https://www.pivotaltracker.com/story/show/169578603
https://www.pivotaltracker.com/story/show/166713447

As long as you review particularly this last story and don't think this is some how related to wrapping a managed interactive, then I think this approach is OK. Hopefully there is a better place to put the style though.

pjanik:
Right, I didn't try to investigate it super deep. My assumption was that if the element is rendered with CSS classnames `question hidden` it should be probably hidden. ;) It was not, so it felt safe to do it. 

`hidden` class has been added by Doug a year ago:
https://github.com/concord-consortium/lara/commit/2ce654b77f33a94e10c485498c2ac2bf9cd6fadf#diff-ec84d4c75be3461f39142c5c2bc72fc0R85
but I guess it could have been added even earlier in some other place.

I was looking at this logic briefly, but basically wrapping plugins are all rendered separately at the beginning. And hidden (at least that seemed to be the intention):
https://github.com/concord-consortium/lara/blob/master/app/views/interactive_pages/_show.html.haml#L18-L19

But I didn't have time to look more at this. If you think it's worth it, I can do it next week. 

Although, this fix seems right to me anyway (as hidden things should be probably hidden, even if we're not solving all the issues).

> Also this style was added to an interactive style sheet but really it is being applied to an all embeddables right?

Yes, it's applied to all questions/embeddables.

> Hopefully there is a better place to put the style though.

This is the file where I added it:
https://github.com/concord-consortium/lara/blob/master/app/assets/stylesheets/partials/_runtime-interactives.scss

Mostly because the `.question` class is described there, including this bottom margin I mentioned before. So, the place/context seems pretty okay to me. But I have no idea why this file is named _runtime-**interactives**. It describes content-mod, intro-txt and many unrelated elements. Maybe we should change its name.

scytacki:
OK that all sounds good.

PR #564, opened 2020-06-12
Interactive and InteractiveRunState: support metadata in #portal_hash

pjanik:
[#172778790]
[#173015432]

I tested it manually and Portal details report seems to be working fine. `portal_hash` is now based on report_service_hash, but it applies necessary transformations.

It's not related, but I disabled STDOUT logging in test env, as it was impossible to see anything when tests are running due to db logs :wink:

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -41,9 +41,4 @@ LightweightStandalone::Application.configure do
 
   # Print deprecation notices to the stderr
   config.active_support.deprecation = :stderr
-
-  if ENV["RAILS_STDOUT_LOGGING"].present?
-    # Disable logging to file. It might have performance impact while using Docker for Mac (slow filesystem sync).
-    config.logger = Logger.new(STDOUT)
-  end
 end
--- a/app/models/base_interactive.rb
+++ b/app/models/base_interactive.rb
@@ -19,7 +19,7 @@ module BaseInteractive
 
     result[:native_width] = result[:width]
     result[:native_height] = result[:height]
-    result[:is_required] = result[:required]
+    result[:is_required] = !!result[:required]
     # If name is not available, but prompt is, this should make Portal details report a bit more readable.
     unless result[:name].present?
       result[:name] = result[:prompt]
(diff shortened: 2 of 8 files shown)


pjanik:
I've just added quite a lot of specs.

First, I extracted BaseInteractive specs to a separate helper, so we can cover both MwInteractive and ManagedInteractive. Then, I noticed that actually #report_service_hash was totally not covered by any specs... I've added them + added tests of these new transformations (interactive -> open response etc.).

I hope these tests could be used as a description of the required format. I was manually inspecting files and checking what LARA is currently sending to Report Service and Portal (so, I was checking e.g. open_response.rb, open_response_answer.rb, etc.). When something was problematic or unclear, I was also checking Portal code (e.g. that's why I skipped answer_texts that isn't used by Portal anyway).

config/environments/test.rb
```diff
@@ -41,9 +41,4 @@
 
   # Print deprecation notices to the stderr
   config.active_support.deprecation = :stderr
-
-  if ENV["RAILS_STDOUT_LOGGING"].present?
-    # Disable logging to file. It might have performance impact while using Docker for Mac (slow filesystem sync).
-    config.logger = Logger.new(STDOUT)
-  end
```
scytacki, line 48: I suppose this means when you are running the tests locally you might end with large log files that are being synced and also saved in the docker container. But I guess if that isn't causing you problems yet, then we can deal with it later.

To deal with the syncing part I'd modify this line: https://github.com/concord-consortium/docker-volume-sync/blob/master/entrypoint.sh#L46 so we can pass additional `-ignore` options via a environment variable to unison.  So then we can configure the sync container to ignore the log folder.

pjanik:
@scytacki, good catch about ManagedInteractives, thanks. I've fixed that and added specs that check these cases too. InteractiveRunState got a bit lengthy/repetitive, but at least the cases are clearly visible and always tested the same way.

config/environments/test.rb
pjanik: > But I guess if that isn't causing you problems yet, then we can deal with it later.

Right, I think the problem starts with huge log files (like my old ones that had a few GB). I didn't have better idea of how to quickly fix it. But with STDOUT logging tests were not usable, as results were totally lost in SQL statements.

scytacki:
Thanks. The changes look good.

PR #565, opened 2020-06-15
Collapsible columns fix

emcelroy:
Adds resize-observer-polyfill and uses that for determining when to trigger interactiveSizing() on column collapse.

PT Story: https://www.pivotaltracker.com/story/show/173180827


scytacki:
@emcelroy I think I fixed this to work better. 
When I tested your initial changes locally,they still didn't handle resizing a pinned interactive correctly. 

Fixing that required a few more changes, and I fixed a few interactive sizing issues along the way.

scytacki:
This PR adds ResizeObserver to the PageItemAuthoring namespace. It really shouldn't be there, since it isn't used in authoring. I'll try to open a separate PR with a new namespace for stuff that could be used either by authoring or runtime.

scytacki:
@pjanik I added you back as a reviewer because you've worked with the resize observer recently, and you worked on this sizing code of both the responsive layout and pinned interactives.

pjanik:
It's a bit hard to follow all the logic without an example and debugging. But the code looks good to me. I thought about the namespace, but yea, it can be updated easily later.

So, now interactive sizing will initialize this whole logic all the time for all iframes with [data-aspect-ratio], right? Even if it's authoring or non-responsive layout? I guess that's the goal here, right?

I'm thinking about these lines that are removed now:
```
- if page.layout == 'l-responsive'
  :javascript
    // Initialize responsive layout when page is loaded.
    $(function() { responsiveLayout(); });
```

And window.resize handler added directly to interactive-sizing.

scytacki:
Yes. I should have included this message from the last commit:
- update the all interactives' size when the window is resized even when not in responsive layout
- monitor the size of the interactive's container so when the questions are collapsed the interactive is resized
- remove the special responsive layout handling because it is now done everywhere
- clean up a trigger on data-height that is not necessary and wouldn't have found anything anyway

scytacki:
There is at least one remaining issue with the bottom navigation. 
https://www.pivotaltracker.com/story/show/173397525

But I think this is better, and this might get replaced with the single page lara runtime (activity player is the new name for that

PR #566, opened 2020-06-15
Implement `showModal()` and `closeModal()` APIs

kswenson:
- currently only implemented for alerts (lightbox/dialog support to follow)
- uses existing jQuery method for showing the alerts (for now)
- adds scripts and documentation to facilitate testing interactive API with local clients

@sfentress I had thought that you would be using these APIs and so it would make sense for you to review this PR. It sounds like you're going a different direction that won't require these APIs, in which case you can consider your review optional.


pjanik:
It also looks good to me. I'd agree that it'd be nice to start leaving iframe-saver slowly.
Since we have `IframePhoneManager` and `IframePhoneManager.getPhone` that should take care of automatic reusing of iframe-phone, I guess it might be not that bad to create a new file(s) that handle various Interactive API messages.

kswenson:
I had the same thought about `iframe-saver.ts` but as that seemed to be the established pattern I was just trying to follow it. If we can come up with a better organization I'm happy to follow it. Should each API "section" be in its own file, e.g. the modal code could go in `iframe-modal.ts`?

kswenson:
Looking at the code again, what if the modal code were implemented as a "plugin" which exported an object with a `getListeners()` function? Then `IFrameSaver` could call `addListener()` for each such exported listener from all such plugins. If that seems reasonable I could make that change in the upcoming lightbox PR.

scytacki:
Yes, I think putting each api section in its own file would be good.  I'm not looking at the code, but just from the descriptions above it sounds like the `getListeners` approach would be abstracting the IframePhomeManager. I'm not sure that new layer of abstraction is needed. An api section might want to add and remove listeners dynamically.  However, I assume there has to be one central place that initialized each of the 'api sections'.  But again I'm not looking at the code. :)

kswenson:
For now what I have in mind would be internal to the `IFrameSaver` and so shouldn't need to involve the `IFramePhoneManager` abstraction. I'll play around with it in the lightbox PR and you can see what you think.

PR #568, opened 2020-06-17
improve authoring interactive sizing

scytacki:
it now handles aspect ratio better
Note this is still not the same handling as at runtime,
it doesn't handle the 'MAX' setting, limit the height, or change depending on the context

### DIFF
--- a/lara-typescript/src/page-item-authoring/common/components/interactive-authoring.tsx
+++ b/lara-typescript/src/page-item-authoring/common/components/interactive-authoring.tsx
@@ -29,6 +29,10 @@ export const InteractiveAuthoring: React.FC<Props> = ({interactive, onAuthoredSt
     onAuthoredStateChange(newAuthoredState);
   };
 
+  const handleSupportedFeatures = (info: any) => {
+    setAuthoringSupported(!!info.features.authoredState);
+  };
+
   const handleReset = () => {
     setAuthoredState(null);
     setResetCount(resetCount + 1);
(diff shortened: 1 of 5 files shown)


scytacki:
Sorry I didn't realize this was half baked. I've got a to update the large FIXME comment.

scytacki:
@dougmartin ok this is ready to look at now

lara-typescript/src/page-item-authoring/common/components/interactive-iframe.tsx
```diff
@@ -25,13 +21,38 @@ interface IFramePhoneParentEndpoint {
...
+  // So probably the best option is to abstract that sizing code so it can be
+  // used by both jQuery and React
+  const [height, setHeight] = useState<number|string|null>(null);
+  const [aspectRatio, setAspectRatio] = useState<number>(authoredAspectRatio);
+
+  const handleHeightChange = (newHeight: number | string) => {
+    setHeight(newHeight);
+  };
+
+  const handleSupportedFeatures = (info: any) => {
```
dougmartin, line 50: This doesn't handle the call to `setAuthoringSupported(!!info.features.authoredState)` that was deleted in this PR in this file:

lara-typescript/src/page-item-authoring/common/components/interactive-authoring.tsx

lara-typescript/src/page-item-authoring/common/components/interactive-authoring.tsx
```diff
@@ -29,30 +29,11 @@ export const InteractiveAuthoring: React.FC<Props> = ({interactive, onAuthoredSt
...
   };
 
-  const handleHeightChange = (newHeight: number | string) => {
-    if (iframe.current) {
-      iframe.current.style.height = `${newHeight}px`;
-    }
-  };
-
-  const handleSupportedFeatures = (info: any) => {
-    setAuthoringSupported(!!info.features.authoredState);
```
dougmartin, line 39: This call to `setAuthoringSupported` isn't handled in the new refactored code.

lara-typescript/src/page-item-authoring/common/components/interactive-iframe.tsx
scytacki: Good catch. next commit takes care of this.  However it looks like it was not being used, but we might use it in the future.

PR #569, opened 2020-06-17
Add header block

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/173020289


app/models/interactive_page.rb
```diff
@@ -307,7 +313,25 @@ def self.import(page_json_object, helper=nil)
...
+      helper = LaraSerializationHelper.new if helper.nil?
+      intro_embeddable_hash = {
+                         "content": intro_text,
+                         "is_full_width": true,
+                         "is_hidden": false,
+                         "name": "",
+                         "type": "Embeddable::Xhtml"
+                        }
+      intro_embeddable = helper.import(intro_embeddable_hash)
+      import_page.show_header = true
```
scytacki, line 333: I'm not sure this ever happens, but it seems possible the old show_introduction could be false. And when that content is then imported show_header will be true.

I don't think this is worth fixing though. As far as I know we don't need to fully support importing legacy content.

app/views/interactive_pages/_edit_section.html.haml
```diff
@@ -0,0 +1,48 @@
+:ruby
+  if section_name == 'HEADER_BLOCK'
+    section_hidden_field = hidden_field_tag :section, InteractivePage::HEADER_BLOCK
+    selector_menu = embeddable_selector
+    section_embeddables_list = @page.header_block_visible_embeddables
+  elsif section_name == 'INTERACTIVE_BOX'
+    section_hidden_field = hidden_field_tag :section, InteractivePage::INTERACTIVE_BOX
+    selector_menu = embeddable_interactives_selector
+    section_embeddables_list = @page.interactive_box_visible_embeddables
```
scytacki, line 9: This and line 5 seem to only be getting the visible embeddables. I'd think during authoring it'd need to get all of the embeddables.

app/views/interactive_pages/_edit_section.html.haml
```diff
@@ -0,0 +1,48 @@
+:ruby
+  if section_name == 'HEADER_BLOCK'
```
scytacki, line 2: Probably `'HEADER_BLOCK'` should be replaced with `InteractivePage::HEADER_BLOCK`

app/views/interactive_pages/_edit_section.html.haml
```diff
@@ -0,0 +1,48 @@
...
+    section_embeddables_list = @page.header_block_visible_embeddables
+  elsif section_name == 'INTERACTIVE_BOX'
+    section_hidden_field = hidden_field_tag :section, InteractivePage::INTERACTIVE_BOX
+    selector_menu = embeddable_interactives_selector
+    section_embeddables_list = @page.interactive_box_visible_embeddables
+  else
+    section_hidden_field = nil
+    selector_menu = embeddable_selector
+    section_embeddables_list = @page.section_embeddables(nil)
+  end
```
scytacki, line 14: If `section_name` is nil, in this last case then you can simplify this code. You could put:
`section_embeddables_list =   @page.section_embeddables(section_name)`
outside of the `if,elseif,else` block so each case doesn't need to do that itself.

app/views/interactive_pages/_show.html.haml
```diff
@@ -2,14 +2,18 @@
   - unless page.name.blank?
     %h4.h4
       = page.name
-  - if page.show_introduction && !page.text.blank?
-    .intro-txt!= page.text
+  - if page.show_header
+    .header-block-mod
+      -# Wrapping plugins don't necessarily wrap items in the assessment block, so they need to be rendered separately.
+      = render partial: 'interactive_pages/list_wrapping_plugins', locals: {wrapping_plugins: main_section_wrapping_plugins(page, @run)}
```
scytacki, line 8: This seems to be rendering the main section wrapping plugins inside of the header block. I think this will mean the main section plugins will be rendered twice.

app/views/interactive_pages/edit.html.haml
```diff
@@ -82,72 +82,25 @@
     - # TODO: Display preview of student completion page.
     =render :partial => "edit_completion"
   - else
-    - if @page.show_introduction
-      = text_editor @page, :text, { header: 'Introduction' }
+    - if @page.show_header
+      %div{:class => 'other full-width'}
+        #header_block.questions
+          %h2 Header Section
+          = render :partial => "edit_section", :locals => { :page => @page, :section_name => 'HEADER_BLOCK', :toggle_option => false }
```
scytacki, line 89: This should use the constant for instead of `'HEADER_BLOCK'`

app/views/interactive_pages/edit.html.haml
```diff
@@ -82,72 +82,25 @@
...
-                      - if e.respond_to? :name
-                        - if !e.name.nil? && !e.name.empty?
-                          #{e.name}
-                        - else
-                          #{e.class.model_name.human.titleize}
-                      - else
-                        #{e.class.model_name.human.titleize}
-                    %span.sortable-icon.ui-icon.ui-icon-arrowthick-2-n-s
-                  = render :partial => "#{e.class.name.underscore.pluralize}/author", :locals => { :embeddable => e, :page => @page, :allow_hide => true }
+          = render :partial => "edit_section", :locals => { :page => @page, :section_name => '', :toggle_option => true }
```
scytacki, line 94: Ah so section_name isn't nil. Unless it causes problems using nil here would be better so it matches how things are modeled.

app/views/interactive_pages/edit.html.haml
```diff
@@ -82,72 +82,25 @@
...
-                    - if interactive.respond_to? :name
-                      - if !interactive.name.nil? && !interactive.name.empty?
-                        #{interactive.name}
-                      - else
-                        #{interactive.class.model_name.human.titleize}
-                    - else
-                      #{interactive.class.model_name.human.titleize}
-                  %span.sortable-icon.ui-icon.ui-icon-arrowthick-2-n-s
-                = render :partial => "#{interactive.class.name.underscore.pluralize}/author", :locals => {:embeddable => interactive, :page => @page, :allow_hide => false}
+          = render :partial => "edit_section", :locals => { :page => @page, :section_name => 'INTERACTIVE_BOX', :toggle_option => false }
```
scytacki, line 103: Should use the constant instead of `'INTERACTIVE_BOX'`

db/migrate/20200622222120_copy_page_intro_to_new_header_textbox.rb
```diff
@@ -0,0 +1,67 @@
...
+                    :interactive_page_id, :embeddable_id, :embeddable_type
+    acts_as_list :scope => :interactive_page
+    belongs_to :interactive_page
+    belongs_to :embeddable, :polymorphic => true, dependent: :destroy
+  end
+
+  def up
+    InteractivePage.find_each(batch_size: 10) do |ip|
+      if ip.text.present?
+        ip.show_header = true
```
scytacki, line 29: This should set show_header to the value of show_introduction. I suspect there are some production activity pages which have hidden introductions but still have some text that the author never cleaned up.

db/migrate/20200622222120_copy_page_intro_to_new_header_textbox.rb
```diff
@@ -0,0 +1,67 @@
...
+    end
+  end
+
+  def down
+    InteractivePage.find_each(batch_size: 10) do |ip|
+      pi_count = 0
+      if ip.text.present?
+        PageItem.where(interactive_page_id: ip.id, section: 'header_block').each do |pi|
+          if pi.embeddable_type == 'Embeddable::Xhtml'
+            textbox = Embeddable::Xhtml.where(id: pi.embeddable_id).first
```
scytacki, line 53: There are more efficient ways to do this, but since we are hopefully not going to need to run this, I don't think it is worth changing it.

db/migrate/20200622222120_copy_page_intro_to_new_header_textbox.rb
```diff
@@ -0,0 +1,67 @@
...
+            if ip.text == textbox.content
```
scytacki, line 54: If we do need to run this, it is likely we'll be doing so after we have already removed the text field. So this test will always be false.  I'd add a test for that case and set the text above this line. Perhaps something like:
```
if ip.text.nil?
 ip.text = textbox.content
end
```

But you should test that to make sure that `ip.text.nil?` is actually true in this case.

db/migrate/20200622222120_copy_page_intro_to_new_header_textbox.rb
```diff
@@ -0,0 +1,67 @@
...
+                       )
+      end
+    end
+  end
+
+  def down
+    InteractivePage.find_each(batch_size: 10) do |ip|
+      pi_count = 0
+      if ip.text.present?
+        PageItem.where(interactive_page_id: ip.id, section: 'header_block').each do |pi|
```
scytacki, line 51: I think the page items have an order. So it would be good to use it here. That way we'll process the first page item first.

db/migrate/20200622222120_copy_page_intro_to_new_header_textbox.rb
```diff
@@ -0,0 +1,67 @@
...
+          if pi.embeddable_type == 'Embeddable::Xhtml'
+            textbox = Embeddable::Xhtml.where(id: pi.embeddable_id).first
+            if ip.text == textbox.content
+              textbox.destroy()
+            end
+          end
+          pi_count = pi_count + 1
+        end
+        if pi_count == 1
```
scytacki, line 60: I'm not sure this pi_count is necessary.  If we've run the migrations and people have started using the header_block to add new stuff. Then if we have have to revert the migrations. We'll still want at least give them their introduction text back.

db/migrate/20200622222120_copy_page_intro_to_new_header_textbox.rb
```diff
@@ -0,0 +1,67 @@
...
+          ip.show_header = false
```
scytacki, line 61: I think this should be `ip.show_introduction = ip.show_header` this way the state of this show_ field will be retained if we have to revert the migrations.

db/migrate/20200625215657_remove_introduction_text_from_interactive_page.rb
```diff
@@ -0,0 +1,9 @@
+class RemoveIntroductionTextFromInteractivePage < ActiveRecord::Migration
+  def up
+    remove_column :interactive_pages, :text
```
scytacki, line 3: There is also the 'show_introduction` column that I think should be removed here.

scytacki:
Looks great. Thanks for all of your work on this.

PR #571, opened 2020-06-18
Add lightbox support to interactive API

kswenson:
- adds lightbox support to `showModal()` /`closeModal()` APIs.
- supports two "modes" (as does the `$.colorbox()` library on which it's based)
  - if `isImage` is true, the specified url is presented in an `` tag
  - otherwise, the specified url is presented in an `<iframe>` tag
- implements a plugin-like architecture for installing handlers in `IFrameSaver`
- modifies our previous modification of the `$.colorbox()` plugin that implements the lightbox so that it no longer assumes that wide images need only be constrained horizontally and tall images need only be constrained vertically.
- fixes some unrelated lint errors on master

Note: all of the issues reported by CodeClimate predate this PR.

### DIFF
--- a/lara-typescript/src/interactive-api-parent/iframe-saver.ts
+++ b/lara-typescript/src/interactive-api-parent/iframe-saver.ts
@@ -1,7 +1,7 @@
 import { ParentEndpoint } from "iframe-phone";
 import * as LaraInteractiveApi from "../interactive-api-client";
 import { IframePhoneManager } from "./iframe-phone-manager";
-import { IFrameSaverPlugin } from "./iframe-saver-plugin";
+import { IFrameSaverPluginDisconnectFn } from "./iframe-saver-plugin";
 import { ModalApiPlugin } from "./modal-api-plugin";
 
 const getAuthoredState = ($dataDiv: JQuery) => {
@@ -106,7 +106,7 @@ export class IFrameSaver {
   private alreadySetup: boolean;
   private iframePhone: ParentEndpoint;
   private successCallback: SuccessCallback | null | undefined;
-  private plugins: IFrameSaverPlugin[];
+  private plugins: IFrameSaverPluginDisconnectFn[];
 
   constructor($iframe: JQuery, $dataDiv: JQuery, $deleteButton: JQuery) {
     this.$iframe = $iframe;
@@ -138,7 +138,7 @@ export class IFrameSaver {
 
     this.iframePhone = IframePhoneManager.getPhone($iframe[0] as HTMLIFrameElement, () => this.phoneAnswered());
 
-    this.plugins = [ModalApiPlugin()];
+    this.plugins = [ModalApiPlugin(this.iframePhone)];
   }
 
   public save(successCallback?: SuccessCallback | null) {
@@ -259,19 +259,6 @@ export class IFrameSaver {
 
     this.addListener("getFirebaseJWT", (request?: IGetFirebaseJwtRequestOptionalRequestId) => {
       return this.getFirebaseJwt(request);
-    });
-
-    // add listeners from "plugins"
-    this.plugins.forEach(plugin => {
-      if (plugin.listeners) {
-        for (const m in plugin.listeners) {
-          if (Object.prototype.hasOwnProperty.call(plugin.listeners, m)) {
-            const msg = m as LaraInteractiveApi.IFrameSaverClientMessage;
-            const listener = msg && plugin.listeners[msg];
-            listener && this.addListener(msg, listener);
-          }
-        }
-      }
     });
 
     if (this.learnerStateSavingEnabled()) {
--- a/lara-typescript/src/interactive-api-parent/iframe-saver.ts
+++ b/lara-typescript/src/interactive-api-parent/iframe-saver.ts
@@ -1,8 +1,8 @@
 import { ParentEndpoint } from "iframe-phone";
 import * as LaraInteractiveApi from "../interactive-api-client";
 import { IframePhoneManager } from "./iframe-phone-manager";
-import { IHintRequest, IShowModal, IShowAlert, ICloseModal, ModalType } from "../interactive-api-client";
-import { addPopup } from "../plugin-api";
+import { IFrameSaverPlugin } from "./iframe-saver-plugin";
+import { ModalApiPlugin } from "./modal-api-plugin";
 
 const getAuthoredState = ($dataDiv: JQuery) => {
   let authoredState = $dataDiv.data("authored-state");
@@ -77,11 +77,6 @@ type IGetAuthInfoResponseOptionalRequestId = LaraInteractiveApi.IGetAuthInfoResp
 type IGetFirebaseJwtRequestOptionalRequestId = LaraInteractiveApi.IGetFirebaseJwtRequest;
 type IGetFirebaseJwtResponseOptionalRequestId = LaraInteractiveApi.IGetFirebaseJwtResponse;
 
-interface ModalState {
-  type: ModalType;
-  $content: JQuery<HTMLElement>;
-}
-
 export class IFrameSaver {
 
   private static instances: IFrameSaver[] = [];
@@ -111,7 +106,7 @@ export class IFrameSaver {
   private alreadySetup: boolean;
   private iframePhone: ParentEndpoint;
   private successCallback: SuccessCallback | null | undefined;
-  private modals: Record<string, ModalState>;
+  private plugins: IFrameSaverPlugin[];
 
   constructor($iframe: JQuery, $dataDiv: JQuery, $deleteButton: JQuery) {
     this.$iframe = $iframe;
@@ -143,7 +138,7 @@ export class IFrameSaver {
 
     this.iframePhone = IframePhoneManager.getPhone($iframe[0] as HTMLIFrameElement, () => this.phoneAnswered());
 
-    this.modals = {};
+    this.plugins = [ModalApiPlugin()];
   }
 
   public save(successCallback?: SuccessCallback | null) {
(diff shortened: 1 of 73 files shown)


lara-typescript/src/interactive-api-parent/iframe-saver.ts
```diff
@@ -278,6 +261,19 @@ export class IFrameSaver {
...
+        for (const m in plugin.listeners) {
+          if (Object.prototype.hasOwnProperty.call(plugin.listeners, m)) {
+            const msg = m as LaraInteractiveApi.IFrameSaverClientMessage;
+            const listener = msg && plugin.listeners[msg];
+            listener && this.addListener(msg, listener);
+          }
+        }
+      }
+    });
+
```
scytacki, line 276: Some plugins will want to send messages and they might also want to remove their listener.
So I'd think it would be better to just initialize the plugins with the iframePhone object and then let them add their own listeners.

However one thing to watch out for (I think) is that iframePhone doesn't allow more than one listener for a message at at time. So one plugin could mess up another.

The interactive api client has a iframephone manager of some kind that manages this better so there can multiple listeners for a message.

lara-typescript/src/interactive-api-parent/modal-api-plugin.ts
```diff
@@ -0,0 +1,145 @@
...
+    const availableHeight = window.innerHeight - kChromeSize.height;
+    const sizeOpts: { width?: number, height?: number, scalePhotos?: boolean } = {};
+    // images are auto-sized correctly without intervention
+    if (isImage) {
+      // prevent scaling up unless requested
+      const requiresUpscaling = (!!size?.width && size.width < availableWidth) &&
+                                (!!size?.height && size.height < availableHeight);
+      sizeOpts.scalePhotos = !requiresUpscaling || options.allowUpscale;
+      return sizeOpts;
+    }
```
scytacki, line 95: I didn't think we were going to support showing an image in this lightbox api. Even the image interactive will need to be an iframe in order to render it custom magnifying glass and caption text.

lara-typescript/src/interactive-api-parent/modal-api-plugin.ts
```diff
@@ -0,0 +1,145 @@
...
+        sizeOpts.width = size.width;
+        sizeOpts.height = size.height;
+      }
+      if (size.width > availableWidth) {
+        // width is constraining dimension
+        sizeOpts.width = availableWidth;
+        sizeOpts.height = availableWidth / aspectRatio;
+      }
+      return sizeOpts;
+    }
```
scytacki, line 114: It might not be necessary at first, but in other cases we try to handle users who resize their browser. So that means this function will need to re-called on resize events.

Also interactive iframes have the ability to request a height or aspect ratio depending on what a user does. In this case it doesn't look like we are connecting iframephone to the iframe in the lightbox.  

However I think that also means that an image interactive that is showing its own magnifying glass to close it self, will have have no way to tell LARA to close itself. So it seems like we will need the iframphone setup.

lara-typescript/src/page-item-authoring/common/components/interactive-iframe.tsx
```diff
@@ -97,9 +97,9 @@ export const InteractiveIframe: React.FC<Props> = (props) => {
   }, [src, resetCount]);
 
   let computedHeight = 300;
-  if(height !== null) {
+  if (height !== null) {
```
scytacki, line 100: looks like we need to have travis run the linter on lara-typescript so I can't get away with this in the future. :)

lara-typescript/src/interactive-api-parent/iframe-saver.ts
kswenson: To my mind, the goal of this effort is purely one of implementation -- make it possible to separate code so that `iframe-saver.ts` doesn't become any more unmaintainable than it already is. This implementation doesn't introduce any additional problems/constraints that weren't already present in the existing inline implementation. Building a robust plugin API here seems beyond the scope of this PR.

That said, I don't have a problem with passing the IFramePhoneManager into the plugin if you'd prefer (I made plugins functions so that there could be arguments) and having the plugin add the listeners itself, but issues like multiple plugins adding multiple listeners for the same message should be treated the way they're treated now -- as programmer error.

If we want to flesh this out into something more robust than simply a technique for separating implementation into multiple files, then we can design such a plugin API and write stories for it.

lara-typescript/src/interactive-api-parent/modal-api-plugin.ts
kswenson: The existing image functionality already supports rendering a magnifying glass and customizable caption text. If it's important to make the magnifying glass customizable as well we can add that capability (although I'm not sure why we'd want different interactives to show different magnifying glass icons).

The only current client of the lightbox API is primarily interested in showing an image. Toward that end, `` tags have a lot of nice auto-sizing capabilities that `<iframe>` tags do not. I suspect there would be a fair amount of work involved to get the `<iframe>` implementation to behave as well as the `` does now -- I started down that path and switched to supporting the image option when the sizing wasn't working. On top of that, the `<iframe>` implementation is more work on the client side as well. I'm not convinced that the additional effort is worth it at this juncture.

lara-typescript/src/interactive-api-parent/modal-api-plugin.ts
kswenson: Resizability seems like a new feature/story. It's not a capability of the existing image question lightbox, for instance. It seems as though we ought to be able to handle browser window resize automatically without additional API.

The existing image lightbox implementation already handles click on the image to close the lightbox, so that functionality is already present. Making it possible for an `<iframe>`d interactive to close the lightbox seems perfectly reasonable but my inclination would be to make it a separate story as well.

lara-typescript/src/interactive-api-parent/iframe-saver.ts
scytacki: Right, it doesn't need to be a full plugin system. I'm just anticipating that the very next plugin will need to send messages as well as receive them. So then reducing the amount of layers that we need to maintain makes this even more simple.

Is frame-saver already using a IFramePhoneManager? It looks like it is just using IFramePhone directly.

lara-typescript/src/interactive-api-parent/modal-api-plugin.ts
scytacki: It kind of seems like making the image into an interactive was kind of a waste then. If LARA is going to directly support showing image light boxes it might as well just render the image itself too. 

My hunch is that once we have the full iframe support required by the drawing tool, then we can remove this image specific api and just use that one. So I guess I'm OK with this approach for now, and I'll wait to see how that dialog api turns out.

Are the `` autosizing features ones provided by Colorbox, or is it just that Colorbox is using these features? 
We have to implement these autosizing features for iframes anyhow. Currently that code is in iframe-sizing.js. I'm trying to get that refactored so at least some parts of it will be usable by React. 

I wonder how far you are going to get with Colorbox. I'd guess it would be better to switch to a React based lightbox/dialog. Since we will need that React lightbox/dialog in the activity player. So if we are putting in more than a little bit of time on Colorbox it is probably not a good thing.

lara-typescript/src/interactive-api-parent/modal-api-plugin.ts
scytacki: Yeah I think it is fine to leave it this way, and we'll see how the dialog stuff works out. My original intention was not to support images just iframes. Which means they'd need to a way to close the lightbox.

The current resizing of iframes doesn't require an additional API for browser size changes. The interactive just provides a height or aspectRatio, and the LARA side of things does the rest of the computations. But at the same since the user can change the content inside of the iframe, and we normally want to avoid putting the scroll bars inside of the iframe itself, so we have an API for the interactive to request more height from LARA.

In any case you will run into all of this when implementing the dialog. Hopefully the interactive sizing will be a little bit cleaner then so it will be easier to understand from the code.

lara-typescript/src/interactive-api-parent/modal-api-plugin.ts
kswenson: >My hunch is that once we have the full iframe support required by the drawing tool, then we can remove this image specific api and just use that one. So I guess I'm OK with this approach for now, and I'll wait to see how that dialog api turns out.

Maybe, but I wouldn't be surprised if full iframe support never reaches full parity with the image-specific API. If all you need to do is show an image, putting an `<iframe>` around an `` tag just adds complexity. Separately, I took a quick look around for react lightbox libraries on npm -- there's a bunch of them that are tailored to work well with images but only one came up that supported non-image content (in addition to images). When we switch to a React-based implementation for the activity player I suspect the image functionality will be straightforward to achieve but the iframe support will be a substantial amount of work there as well.

>Are the `` autosizing features ones provided by Colorbox, or is it just that Colorbox is using these features?

Colorbox simply uses an `` tag when an image is requested and `` tags resize their contents in convenient ways that `<iframe>`s don't.

>I wonder how far you are going to get with Colorbox. I'd guess it would be better to switch to a React based lightbox/dialog. Since we will need that React lightbox/dialog in the activity player. So if we are putting in more than a little bit of time on Colorbox it is probably not a good thing.

Which is exactly why I'm using the part of Colorbox that works without additional modification. I'm assuming that the jQuery-based implementation will only exist until the activity player gets far enough along, so there's no point in spending any more time on it than is required to achieve parity with the existing implementation.

lara-typescript/src/interactive-api-parent/iframe-saver.ts
kswenson: It's just using the `ParentEndpoint` from iFramePhone directly. I just pushed a commit which changes the plugin to take iFramePhone as an argument. It took a while to get the tests stubbed/mocked properly but I think I've wrestled it to the ground. It makes the `post()` method available to plugins.

scytacki:
This looks good to me. If we can hold off merging it until Piotr has a fix for a metadata issue, that will make releasing LaRA 1.52.0 a little easier.

PR #573, opened 2020-06-18
InteractiveAPI: make getFirebaseJwt consistent with PluginAPI#getFirebaseJwt

pjanik:
[#173088322]

1. First commit is unrelated to changes, there were some lint errors breaking build.
2. I moved IJwtResponse types from plugin-api dir to `shared/types`. The same types are used by InteractiveAPI. I think it makes sense to share these types, as we'll guarantee that both methods remain consistent.
2. I've updated api function arguments, so now it accepts `firebaseApp` instead of `firebase_app`. But implementation translates it to firebase_app so over-the-wire message format isn't affected.
3. I removed some random, unused types from `types.ts` and added there a comment with a link to the discussion about naming patterns. Personally, I'd divide it into smaller files. I could definitely see something like `over-the-wire-types.ts`. This is useful for sure, as it's documentation of iframe-phone messages. But e.g. I'm not so sure if we really need to export all the `...Options` interfaces. They are just arguments of API functions. Or we could just export them directly from api.ts file to make the main types file smaller.


pjanik:
One more inconsistency I noticed is getFirebaseJwt arguments.

- Interactive API accepts { firebaseApp: string }
- Plugin API expects just string

This seems less important with TypeScript, but I can change it to string only, and if we need some options one day, we could add second argument with options hash.

scytacki:
Over the wire (message in postMessage), I think should always send objects not strings.
But I think you are taking just about internal functions in the api. Yes in this case it is fine to make it just a string.

pjanik:
I changed signature from `getFirebaseJWT({firebaseApp: string})` to `getFirebaseJwt(firebaseApp: string)`, so it's exactly the same like in PluginAPI and resolved merge conflicts.

PR #575, opened 2020-06-23
Several tweaks to interactive sizing

scytacki:
- figure out if the theme has a pinned header, if it does not then pin the interactive to the top of the screen instead of with a margin. Perviously the interactive would always be pinned with a margin for the header even if the theme didn't have one.
- include the size of the pinned header (if there is one) in the maxHeight calculation in full width layout of interactive box and and assessment block interactives. This maxHeight calcuation is used when an interactive is set to "use all available space" or "default aspect ratio, or set by interactive".
- remove special handling of multiple interactives in the interactive box. Currently when there are multiple interactives, the height limiting is disabled. Note: with the handle of multiple interactives, the height limiting is only based on the window height. So if you embed multiple interactives it is unlikely that the interactive box will be pinned.
- match the width of wrapping plugins to the width of the interactive.
- update the interactive size when a wrapping plugin updates the wrapping content

Additionally this removes a random `<br>` that was at the end of the interactive.
This might be related to the fact that interactive iframe has a display:inline instead of display:block
Removing it fixed some extra whitespace and I couldn't see any negatives.

This includes several updates to comments as well.

[#173456421]

### DIFF
--- a/app/assets/javascripts/runtime-init.js
+++ b/app/assets/javascripts/runtime-init.js
@@ -1,6 +1,22 @@
 /*jslint browser: true, sloppy: true, todo: true, devel: true, white: true */
 /*global $ */
 
+// Some themes pin the header and some do not
+var _headerIsPinnedInThisTheme = null;
+function headerIsPinnedInThisTheme() {
+  // This currently seems to be the best way to check this. Themes just change
+  // CSS so there isn't a way for them add content to the dom or run javascript
+  // This is fragile though because someone might mess with .conten-hdr and then
+  // this will break.
+  // We cache the value since this won't change
+  if (_headerIsPinnedInThisTheme !== null) {
+    return _headerIsPinnedInThisTheme;
+  } else {
+    _headerIsPinnedInThisTheme = $('.site-width > .content-hdr').is(':visible');
+    return _headerIsPinnedInThisTheme;
+  }
+}
+
 $(document).ready(function () {
   var headerTop = null;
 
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/runtime-init.js
```diff
@@ -1,6 +1,22 @@
 /*jslint browser: true, sloppy: true, todo: true, devel: true, white: true */
 /*global $ */
 
+// Some themes pin the header and some do not
+var _headerIsPinnedInThisTheme = null;
+function headerIsPinnedInThisTheme() {
+  // This currently seems to be the best way to check this. Themes just change
+  // CSS so there isn't a way for them add content to the dom or run javascript
```
kswenson, line 8:
```suggestion
  // CSS so there isn't a way for them to add content to the dom or run javascript
```

app/assets/javascripts/runtime-init.js
```diff
@@ -1,6 +1,22 @@
...
+  // This is fragile though because someone might mess with .conten-hdr and then
```
kswenson, line 9:
```suggestion
  // This is fragile though because someone might mess with .content-hdr and then
```

scytacki:
This page demonstrates how the extra header space is no longer there when pinning the interactive:
https://lara-demo.concordqa.org/activities/20007/pages/304541
You can compare that to this page:
https://authoring.concord.org/activities/9620/pages/124107
The extra space you see on production is to handle the pinned navigation you see here:
https://lara-demo.concordqa.org/activities/20008/pages/304557

scytacki:
I just found out that the waters theme does not seem to be pinned, but the computation I'm using to determine the header size is not figuring this out. You can currently see the issue here:
https://lara-demo.concordqa.org/activities/20009/pages/304573

scytacki:
Ethan is going to change how the Waters theme disables the navigation pinning, so it matches the other themes that disable the navigation pinning.  So that should fix the waters issue.

scytacki:
Piotr reviewed this on slack:
> I looked at it today more carefully and it seems good to me. Ofc it would be nice to have even less edge cases, but I guess we just need all that having pinning, wrapping plugins, and all the sizing options. Things seem nicely described in this comment and changes make sense to me.

PR #578, opened 2020-06-26
Support rich text (HTML) hints

kswenson:
See [question-interactives #33](https://github.com/concord-consortium/question-interactives/pull/33) for the corresponding client code.

### DIFF
--- a/app/assets/javascripts/lara-typescript.js
+++ b/app/assets/javascripts/lara-typescript.js
@@ -10239,6 +10239,1234 @@ function toVal(mix) {
 
 /***/ }),
 
+/***/ "./node_modules/dompurify/dist/purify.js":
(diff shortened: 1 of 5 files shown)


kswenson:
>Optionally, I could imagine a simple test that tries to send a hint with a JS embedded and checking if it gets removed.

Good call -- I added some tests.

kswenson:
>I didn't think we needed to support rich text hints, but if we do, this seems OK.

It was mentioned in the PT story.

>It seems DOMPurify doesn't strip style tags,

It can be configured to do so:
```
// leave all as it is but forbid <style>
var clean = DOMPurify.sanitize(dirty, {FORBID_TAGS: ['style']});
```

PR #579, opened 2020-06-26
Add PORTAL_PROTOCOL env to portal-proxy overlay.

knowuh:
The default value is set to `http` as by `docker-compose.yml`

README.md includes information about setting up SSL in dinghy


knowuh:
@scytacki these are the changes we discussed earlier Today

PR #580, opened 2020-06-28
Make `ReactDOMServer` available to clients in the same way as `React` and `ReactDOM`

kswenson:
The rich text editor uses the [renderToStaticMarkup](https://reactjs.org/docs/react-dom-server.html#rendertostaticmarkup) function from [ReactDOMServer](https://reactjs.org/docs/react-dom-server.html) when exporting to HTML. `ReactDOMServer` is published as part of the `ReactDOM` package, but is not bundled with it and so must be imported separately:
```
import ReactDOMServer from "react-dom/server";
```
Like `React` and `ReactDOM`, `ReactDOMServer` has additional debug code in `development` mode that is stripped out in `production` mode. These development-only additions are interdependent so that all of them must be bundled using the same mode (`development`|`production`). This generally isn't a problem as they're usually all being processed by the same bundler/loader. In the case of LARA, however, production versions of the `React` and `ReactDOM` libraries are loaded by the Ruby environment, but prior to this PR `ReactDOMServer` has not been made available. Furthermore, the `lara-typescript` bundle which includes the rich text editor (which requires `ReactDOMServer.renderToStaticMarkup`) bundles everything in `development` mode. Thus, the `renderToStaticMarkup()` function fails when attempting to access some development-only state of `React`.

This PR adds a production version of `ReactDOMServer` to the list of libraries loaded by the Ruby code and made available to the JavaScript runtime environment.


pjanik:
Looks good to me. I'm not sure how/where is this new function being used, though. But I guess it's in some other repository, or will be used soon, right? I'm wondering if we shouldn't use webpack externals since we do it for React and React-DOM:
https://github.com/concord-consortium/lara/blob/master/lara-typescript/webpack.config.js#L82-L83
That way, the other library won't depend on LARA API. It'll just list react-dom-server as an external dependency.

kswenson:
Actually, it is being handled as an external webpack dependency -- see the change to `webpack.config.js` in this PR. The`PageItemAuthoring.renderToStaticMarkup` function was mainly so I could unit-test the fact that it's working, but since it's working I could theoretically remove the function and its test code.

kswenson:
I simplified the unit test and removed the `PageItemAuthoring.renderToStaticMarkup` function.

PR #582, opened 2020-06-29
Optimize interactive state saving

pjanik:
[#173231403]

There are two main changes here:

1. Interactive API `setInteractiveState` delays sending iframe-phone message by 2s. It means that when user is typing, state updates should be accumulated into one message that will be sent 2s after he stops typing. This should greatly limit number of AJAX requests coming from iframe-saver.

2. Interactive API keeps track if interactiveState is "dirty" and if it is, it displays a warning to a user who tries to reload or leave the page. This wasn't implemented before and user could reload the page and loose data. LARA was only trying to prevent data loss when user was using LARA navigation buttons if I interpret the code correctly. 

@scytacki, when we talked about it, you mentioned that "getInteractiveState" request coming from LARA can be accidentally answered by random "setInteractiveState" call from iframe. And we thought about using `requestId`. I was investigating that, but using requestId from parent side isn't easy and I ran out of time. For example, it won't be trivial to differentiate responses to "getInteractiveState" coming from interactives that don't support `requestId` and regular "interactiveState" messages triggered by `setInteractiveState` API call, etc.

I think that iframe-parent communication is almost immediate in contrast to AJAX calls to LARA server. So, iframe-phone calls shouldn't queue in practice and we shouldn't get these false answers. Especially now when there are much less iframe-phone messages being sent. The problem with my data loss (that I could replicate when I simulate slow connection in dev tools) was about the long queue of AJAX requests that were getting lost (and that were enabling page navigation due to iframe-saver / save-on-change code). Of course, this can still happen, but just much less likely without AJAX traffic jam. And I guess LARA SPA will have to deal with this problem anyway.

### DIFF
--- a/lara-typescript/src/interactive-api-client/api.ts
+++ b/lara-typescript/src/interactive-api-client/api.ts
@@ -42,6 +42,8 @@ export const getInteractiveState = <InteractiveState>(): InteractiveState | null
   return getClient().managedState.interactiveState;
 };
 
+let setInteractiveStateTimeoutId: number;
+export const setInteractiveStateTimeout = 2000; // ms
 /**
  * Note that state will become frozen and should never be mutated.
  * Each time you update state, make sure that a new object is passed.
@@ -62,7 +64,11 @@ export const getInteractiveState = <InteractiveState>(): InteractiveState | null
 export const setInteractiveState = <InteractiveState>(newInteractiveState: InteractiveState | null) => {
   const client = getClient();
   client.managedState.interactiveState = newInteractiveState;
-  client.post("interactiveState", newInteractiveState);
+  window.clearTimeout(setInteractiveStateTimeoutId);
+  setInteractiveStateTimeoutId = window.setTimeout(() => {
+    client.post("interactiveState", newInteractiveState);
+    client.managedState.interactiveStateDirty = false;
+  }, setInteractiveStateTimeout);
 };
 
 export const getAuthoredState = <AuthoredState>(): AuthoredState | null => {
--- a/docs/interactive-api-client/README.md
+++ b/docs/interactive-api-client/README.md
@@ -1,6 +1,6 @@
-[@concord-consortium/lara-interactive-api - v0.5.0-pre.2](README.md) › [Globals](globals.md)
+[@concord-consortium/lara-interactive-api - v0.5.0-pre.3](README.md) › [Globals](globals.md)
 
-# @concord-consortium/lara-interactive-api - v0.5.0-pre.2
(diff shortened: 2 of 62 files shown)


lara-typescript/src/interactive-api-client/api.ts
```diff
@@ -62,7 +64,11 @@ export const getInteractiveState = <InteractiveState>(): InteractiveState | null
 export const setInteractiveState = <InteractiveState>(newInteractiveState: InteractiveState | null) => {
   const client = getClient();
   client.managedState.interactiveState = newInteractiveState;
-  client.post("interactiveState", newInteractiveState);
+  window.clearTimeout(setInteractiveStateTimeoutId);
```
scytacki, line 67: If the interactive is constantly changing its state then the state will never be saved. I guess that is OK, because I can't of a case where that would be a normal.

pjanik: Right. There are two things that can help in this case:
- LARA has auto-save feature, it sends "getInteractiveState" every 2 or 5s (I don't remember now)
- it sends "getInteractiveState" before user leaves the page
So, the state still should be saved sooner or later.

pjanik:
@scytacki, I've just published 0.5.0-pre.3 from master. Apparently Dough didn't publish it before (neither did I).

One catch related to state saving - I realized that Scaffolded Question sub-interactive will never receive "getInteractiveState" message. It's not exposed by the new Client and interactive can't really listen to it (without overwriting the default handler). So, scaffolded question can't really pass this message down to subinteractive. It means that subinteractive state saving will rely only on this 2 seconds delay. 

But I tested it and actually it'll still block page unload if you try to reload it or navigate away within these 2 seconds. It'll just use the featured just added in this PR ("onbeforeunlad" native browser alert), instead of typical LARA "getInteractiveState" msg sent just before user tries to navigate away. So, it's a bit less nice experience for user when he uses default navigation (reload is the same), but data shouldn't be lost anyway and we don't need to complicate the code too much. Just raising it here if someone notices this small difference in behavior and is wondering what happens, or I'm not seeing some other potential problems.

dougmartin:
Yes, I didn't publish.  I think we shouldn't publish until code is merged.

pjanik:
I published lots of 0.4.0-pre.x, as npm link wasn't working for me (but it's been fixed by Kirk since then). ;)
Usually, I assume that it's pretty okay to publish `-pre.x` version using `--tag beta` (so other projects don't upgrade accidentally) and master could use non-pre numbers. But it seems we're using only pre releases so far.

PR #584, opened 2020-06-30
173589212 managed interactive form

scytacki:
Some updates to the managed interactives form:
https://www.pivotaltracker.com/story/show/173589212


scytacki:
@dougmartin This isn't the cleanest. If you have suggestions on how to make this better let me know. I was at least going to replace 
```
=> {
 return (<>
```
with just
```
=> <>
```
In a few places.

PR #585, opened 2020-06-30
Add more clearance to make all form elements accessible.

emcelroy:
This fixes the problem reported in this PT story: https://www.pivotaltracker.com/story/show/173541459

I think there may be a better solution, but it would involve more work. I'm not 100% sure, but I suspect the reason why the window shade form needs more clearance is because the button element that appears at the top of the form has a negative `top` value in its CSS. It may be similar to how that causes issues with overlapping in the page authoring form.

### DIFF
--- a/app/assets/stylesheets/partials/authoring/modal_edit.scss
+++ b/app/assets/stylesheets/partials/authoring/modal_edit.scss
@@ -10,10 +10,10 @@
 
   form[class^="edit_"], form.styled-admin-form {
     height: auto;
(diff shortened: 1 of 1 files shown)


emcelroy:
Yeah, it's complex. I think it's mostly because I ran into a Firefox bug/feature where it treats padding differently than other browsers in certain circumstances. Using the `:after` pseudo element was the only way to get around that. A more drastic rewrite of the HTML might lead to a more elegant solution, but I don't think there's time for it.

PR #589, opened 2020-07-09
Don't list wrapping plugins separately from the items they wrap.

emcelroy:
PT story: https://www.pivotaltracker.com/story/show/173737116

### DIFF
--- a/app/views/interactive_pages/_edit_section.html.haml
+++ b/app/views/interactive_pages/_edit_section.html.haml
@@ -30,7 +30,7 @@
     = submit_tag 'Add', {id: 'add_embeddable_button'}
 .embeddables_list.sortable_embeddables
   - section_embeddables_list.each do |e|
-    - if !is_wrapping_plugin?(e)
+    - if e.show_in_edit?
       - tracked = e.respond_to?(:tracked_question) && e.tracked_question
       - tracked_class = tracked ? 'tracked' : ''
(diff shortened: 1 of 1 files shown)


app/views/interactive_pages/_edit_section.html.haml
```diff
@@ -30,17 +30,18 @@
...
   - section_embeddables_list.each do |e|
-    - tracked = e.respond_to?(:tracked_question) && e.tracked_question
-    - tracked_class = tracked ? 'tracked' : ''
-    .authorable.accordion_embeddable{:id => "embeddable_#{e.id}.#{e.class.to_s}", :class => tracked_class}
-      %h3.embeddable_heading
-        %span.embeddable_title
-          - if e.respond_to? :name
-            - if !e.name.nil? && !e.name.empty?
-              #{e.name}
+    - if !is_wrapping_plugin?(e)
```
scytacki, line 33: looking at the header block PR I see this line seemed to have been lost: https://github.com/concord-consortium/lara/pull/569/files#diff-5cad69fb9582aebf98d2ee925ed4f983L107

It looks like `!is_wrapping_plugin?(e)` and `e.show_in_edit?` do the same thing, but show_in_edit is more semantic so I think it is better to use that here. Unless you already tried it :)

emcelroy: No, I didn't try that. Sorry, I somehow missed it while looking for existing ways to exclude wrapping plugins. I agree that's the better option.

PR #590, opened 2020-07-14
Add activity export API, Activity Player preview links

sfentress:
This

1. Adds a new API route, `/api/v1/activities/{:id}.json`, which provides the JSON content of a lightweight activity. This gives the same JSON as `activities/{:id}/export`, but (a) is CORS enabled and (b) doesn't try to download the file. This route does not require authentication, on the assumption that activity structure is not hidden from anonymous users
2. Adds new Preview links to the top of the editing pages, "Preview in Activity Player." These depend on a new environment variable, ACTIVITY_PLAYER_URL.

Note, I may have defined the environment variable in too many places, or not enough -- trying to copy the patterns of the other environment variables. Let me know if I made a mistake. It's also not clear to me which is the correct way for a developer to point to a localhost server -- adding a new file to `docker/dev/...` and including that by default in `.env`?

Also it sounds like I'll need to make a PR in https://github.com/concord-consortium/cloud-formation, but I assume I can do that after this one is in.

Note: This PR relates to https://github.com/concord-consortium/activity-player/pull/13

### DIFF
--- a/app/controllers/api/v1/lightweight_activities_controller.rb
+++ b/app/controllers/api/v1/lightweight_activities_controller.rb
@@ -5,4 +5,14 @@ class Api::V1::LightweightActivitiesController < API::APIController
     activity.destroy
     render :json => { success: true }
   end
+
+  # GET /api/v1/activities/1.json
+  def show
+    activity = LightweightActivity.find(params[:id])
+    lightweight_activity_json = activity.export.to_json
+    respond_to do |format|
+      format.html # show.html.html.haml
+      format.json { render json: lightweight_activity_json }
+    end
+  end
 end
--- a/app/controllers/api/v1/lightweight_activities_controller.rb
+++ b/app/controllers/api/v1/lightweight_activities_controller.rb
@@ -10,9 +10,6 @@ class Api::V1::LightweightActivitiesController < API::APIController
   def show
     activity = LightweightActivity.find(params[:id])
     lightweight_activity_json = activity.export.to_json
-    respond_to do |format|
-      format.html # show.html.html.haml
-      format.json { render json: lightweight_activity_json }
-    end
+    render json: lightweight_activity_json
   end
 end
--- a/app/helpers/lightweight_activity_helper.rb
+++ b/app/helpers/lightweight_activity_helper.rb
@@ -60,10 +60,11 @@ module LightweightActivityHelper
   end
 
   def activity_player_path(activity)
-    return  "#{ENV['ACTIVITY_PLAYER_URL']}/?activity=#{@activity.id}&baseUrl=#{CGI.escape(root_url)}"
+    activity_api_path = "#{api_v1_activity_url(@activity.id)}.json"
+    return  "#{ENV['ACTIVITY_PLAYER_URL']}/?activity=#{CGI.escape(activity_api_path)}"
   end
 
   def activity_player_page_path(activity, page)
-    return  "#{ENV['ACTIVITY_PLAYER_URL']}/?activity=#{@activity.id}&page=#{page.position}&baseUrl=#{CGI.escape(root_url)}"
+    return  "#{activity_player_path(activity)}&page=#{page.position}"
   end
 end
--- a/app/helpers/lightweight_activity_helper.rb
+++ b/app/helpers/lightweight_activity_helper.rb
@@ -64,6 +64,6 @@ module LightweightActivityHelper
   end
 
   def activity_player_page_path(activity, page)
-    return  "#{ENV['ACTIVITY_PLAYER_URL']}/?activity=#{@activity.id}&page=#{page.id}&baseUrl=#{CGI.escape(root_url)}"
+    return  "#{ENV['ACTIVITY_PLAYER_URL']}/?activity=#{@activity.id}&page=#{page.position}&baseUrl=#{CGI.escape(root_url)}"
   end
 end
--- a/app/helpers/lightweight_activity_helper.rb
+++ b/app/helpers/lightweight_activity_helper.rb
@@ -59,12 +59,12 @@ module LightweightActivityHelper
     end
   end
 
-  def activity_player_path(activity)
-    activity_api_path = "#{api_v1_activity_url(@activity.id)}.json"
-    return  "#{ENV['ACTIVITY_PLAYER_URL']}/?activity=#{CGI.escape(activity_api_path)}"
+  def activity_player_url(activity)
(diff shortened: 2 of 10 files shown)


app/controllers/api/v1/lightweight_activities_controller.rb
```diff
@@ -5,4 +5,14 @@ def destroy
...
   end
+
+  # GET /api/v1/activities/1.json
+  def show
+    activity = LightweightActivity.find(params[:id])
+    lightweight_activity_json = activity.export.to_json
+    respond_to do |format|
+      format.html # show.html.html.haml
+      format.json { render json: lightweight_activity_json }
+    end
```
scytacki, line 16:
```suggestion
    render json: lightweight_activity_json
```
Since this is an api I don't think it should support the html format.

app/helpers/lightweight_activity_helper.rb
```diff
@@ -58,4 +58,12 @@ def complete_badge_for(activity)
       return ribbon(t("COMPLETED"),"my-ribbon")
     end
   end
+
+  def activity_player_path(activity)
+    return  "#{ENV['ACTIVITY_PLAYER_URL']}/?activity=#{@activity.id}&baseUrl=#{CGI.escape(root_url)}"
```
scytacki, line 63: Does the activity player need to make additional requests back to lara?
My preference would be to keep this launching argument simple. So it just passes the activityUrl directly instead of the baseUrl and the activity id. This way the activity player doesn't have any lara specific knowledge about how to construct the activity url.

You should be able to get the activity url by calling `lightweight_activity_url(@activity)` you might need the id on there: `lightweight_activity_url(@activity.id)`

config/app_environment_variables.sample.rb
```diff
@@ -20,6 +20,7 @@
 # ENV['C_RATER_FAKE']                    ||= 'true'
 ENV['ROLLBARS_KEY']                      ||= 'XXXXXXX'
 ENV['LABBOOK_PROVIDER_URL']              ||= 'https://labbook.concord.org'
+ENV['ACTIVITY_PLAYER_URL']               ||= 'https://activity-player.concord.org/branch/master'
```
scytacki, line 23: We are trying to move away from using this file. So I'd remove this line to keep things more simple.

docker/prod/app_environment_variables.rb
```diff
@@ -4,6 +4,8 @@
 
 ENV['LABBOOK_PROVIDER_URL']              ||= 'https://labbook.concord.org'
 
+ENV['ACTIVITY_PLAYER_URL']               ||= 'https://activity-player.concord.org/branch/master'
```
scytacki, line 7: It'd be best to just remove this, it follows the pattern of not using this file, and if someone deploys LARA without a configuration will get the master branch which won't be stable.

app/views/interactive_pages/edit.html.haml
```diff
@@ -48,6 +48,8 @@
       = link_to "Preview", preview_activity_page_path(@activity, @page), :target => 'new'
     %span.preview
       = link_to "Preview Teacher Edition", preview_activity_page_path(@activity, @page, mode: "teacher-edition"), :target => 'new'
+    %span.preview
```
scytacki, line 51: I'd recommend wrapping this in a conditional:
`if ENV['ACTIVITY_PLAYER_URL'].present?` that way we can easily enable and hide this link by un-defining the variable or setting it to the empty string.

app/controllers/api/v1/lightweight_activities_controller.rb
sfentress: :+1:

app/helpers/lightweight_activity_helper.rb
sfentress: Hmm. Part of my thinking was that we'd eventually want to support urls that are as clean as the ones in existing LARA, such as `/activity/123/`, and so wanted to separate the activity # from the base url in case we set a global base url in the Activity Player at some point. But since this isn't on the radar (and can always be changed in the future) this maybe isn't necessary.

In terms of LARA-specific knowledge, what it uses is the API route, which I figured was supposed to be something freely useable by anything anticipating working with a LARA-like service.

But I can go ahead and change this to use the url directly. `lightweight_activity_url` doesn't seem to be supported, with either argument. In any case what we need is the `api/` url. Do you know if there's a magic rails route that would go to that?

scytacki: In general you can run `rake routes` to find these magic names. In docker from your host that would be the craxy `docker exec app bundle exec rake routes`.

I'd guess based on the route.rb file that what you need is `activity_url(@activity)`

I just did a little digging and we might be able to support urls like activity-player.concord.org/activities/123 using this approach: https://stackoverflow.com/questions/16267339/s3-static-website-hosting-route-all-paths-to-index-html

But yeah I think it should be a different feature since we'll also want previewing from developer machines and staging to work. And with the url parameter we can also create static test exports which the activity player can load in.

sfentress: :+1:

Updated in fb18e72, Activity Player updated in https://github.com/concord-consortium/activity-player/pull/21.

sfentress:
@scytacki Thanks for the comments. I believe this should be all set.

app/helpers/lightweight_activity_helper.rb
```diff
@@ -58,4 +58,13 @@ def complete_badge_for(activity)
       return ribbon(t("COMPLETED"),"my-ribbon")
     end
   end
+
+  def activity_player_path(activity)
+    activity_api_path = "#{api_v1_activity_url(@activity.id)}.json"
+    return  "#{ENV['ACTIVITY_PLAYER_URL']}/?activity=#{CGI.escape(activity_api_path)}"
```
scytacki, line 64: By using `_path` it won't be a full URL. So then this won't work well when we are testing on authoring.staging.concord.org or lara-master.concordqa.org and production since those will be different domains.

sfentress: It's grabbing `api_v1_activity_url`. It's the variable that is (maybe poorly-) named `_path`.

sfentress: I want ahead and renamed the variable (and helper methods) just to make that totally clear.

scytacki: Sorry, I didn't see that before. Thanks for updating it.

PR #592, opened 2020-07-16
Correct question number order.

emcelroy:
Related PT story: https://www.pivotaltracker.com/story/show/173826188

### DIFF
--- a/app/models/interactive_page.rb
+++ b/app/models/interactive_page.rb
@@ -132,7 +132,21 @@ class InteractivePage < ActiveRecord::Base
   end
 
   def reportable_items
-    visible_embeddables.select { |item| item.reportable? }
+    all_visible_embeddables = []
+    if self.show_header
+      all_visible_embeddables.concat header_block_visible_embeddables.select { |item| item.reportable? }
+    end
+    if self.show_info_assessment
+      all_visible_embeddables.concat main_visible_embeddables.select { |item| item.reportable? }
+    end
+    if self.additional_sections
+      self.additional_sections.each do |as|
+        all_visible_embeddables.concat section_visible_embeddables(as).select { |item| item.reportable? }
+      end
+    end
+    if self.show_interactive
+      all_visible_embeddables.concat interactive_box_visible_embeddables.select { |item| item.reportable? }
+    end
   end
 
   def add_embeddable(embeddable, position = nil, section = nil)
--- a/app/models/interactive_page.rb
+++ b/app/models/interactive_page.rb
@@ -132,21 +132,11 @@ class InteractivePage < ActiveRecord::Base
   end
 
   def reportable_items
-    all_visible_embeddables = []
-    if self.show_header
-      all_visible_embeddables.concat header_block_visible_embeddables.select { |item| item.reportable? }
-    end
-    if self.show_info_assessment
-      all_visible_embeddables.concat main_visible_embeddables.select { |item| item.reportable? }
-    end
-    if self.additional_sections
-      self.additional_sections.each do |as|
-        all_visible_embeddables.concat section_visible_embeddables(as).select { |item| item.reportable? }
-      end
-    end
-    if self.show_interactive
-      all_visible_embeddables.concat interactive_box_visible_embeddables.select { |item| item.reportable? }
-    end
+    all_visible_embeddables = self.show_header ? header_block_visible_embeddables.select { |item| item.reportable? } : []
+    all_visible_embeddables += self.show_info_assessment ? main_visible_embeddables.select { |item| item.reportable? } : []
+    # the following line may need to be changed/augmented if we ever have more than one kind of additional section
+    all_visible_embeddables += self.show_arg_block ? section_visible_embeddables("arg_block").select { |item| item.reportable? } : []
+    all_visible_embeddables += self.show_interactive ? interactive_box_visible_embeddables.select { |item| item.reportable? } : []
   end
 
   def add_embeddable(embeddable, position = nil, section = nil)
--- a/app/models/interactive_page.rb
+++ b/app/models/interactive_page.rb
@@ -154,11 +154,6 @@ class InteractivePage < ActiveRecord::Base
   end
 
   def reportable_items
-    # all_visible_embeddables = self.show_header ? header_block_visible_embeddables.select { |item| item.reportable? } : []
-    # all_visible_embeddables += self.show_info_assessment ? main_visible_embeddables.select { |item| item.reportable? } : []
-    # the following line may need to be changed/augmented if we ever have more than one kind of additional section
-    # all_visible_embeddables += self.show_arg_block ? section_visible_embeddables("arg_block").select { |item| item.reportable? } : []
-    # all_visible_embeddables += self.show_interactive ? interactive_box_visible_embeddables.select { |item| item.reportable? } : []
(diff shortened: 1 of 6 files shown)


emcelroy:
@scytacki What do you think of the cognitive complexity issue raised by Code Climate here? I can reduce the complexity, but then Code Climate complains about the file having too many lines. The increase in lines is too much to easily fix. Does it make sense to maybe move some methods out of the model? It has 31 methods, much more than the 20 method limit.

app/models/interactive_page.rb
```diff
@@ -132,7 +132,11 @@ def interactive_box_visible_embeddables
   end
 
   def reportable_items
-    visible_embeddables.select { |item| item.reportable? }
+    all_visible_embeddables = self.show_header ? header_block_visible_embeddables.select { |item| item.reportable? } : []
```
scytacki, line 135: This approach doesn't seem right. Somehow the reportable_items were ordered correct before. In other words the assessment block items came first, then either the arg_block or interactive box, and then the other one. How did this work before?

Perhaps the way it working before wasn't good, but I'd like to understand that before adding more logic.

emcelroy: I didn't understand how it was working correctly before. It wasn't clear to me that there was anything set up to ensure the correct order. I could be wrong, but I suspected it had only worked before because there were fewer sections and the order just happened to come out correctly. 

Previously the `visible_embeddables` method had one line: `embeddables.select{ |e| !e.is_hidden }`

And `embeddables` was generated by the `embeddables` method which also had one line: `page_items.collect{ |qi| qi.embeddable }`

Sorry, but that's all I know right now. I'll try to look more into it later.

scytacki: I think I found it. It was this ordering configuration here: 
`has_many :page_items, :order => [:section, :position], ...`
I'd suspect that `nil` comes before `header_block` when it is being sorted. `nil` also comes before `interactive_box` so that is why it was working before. And `arg_block` comes before `interactive_box` so that would have also been ordered correctly.

Because the order of the embeddables also determines the export order. And it might be used in other cases too, it would be best if the order of the embeddables continued to match what is expected. And matching the page_item order would be best too.

I'm looking to see if we can customize the `:order` configuration...

scytacki: The `:order` config just passes the result down to SQL `ORDER BY`. So we are limited on how this is ordered.

Option 1:
Rename all of the sections with a prefix like `00_header_block`, `01_assessment_block`, `02_arg_block`, `03_interactive_box`.  This would require a migration. And require updating any code that is using these strings directly. As well as any code that is assuming the assessment_block has a nil section type.

Option 2:
Give up on having page_items ordered correctly. Instead push some of the logic of this PR into the `def emebeddables` method. At least that way code using embeddables will still get the correct list. 

I'd vote for option 2 because it seems like a safer change.  

I would change the code in the PR though so additional_sections was handled better. My suggestion would be to add a class level array like @registered_additional_sections which lists all of the possible sections. This array would then be used by the `def embeddables`. 

When `def self.register_additional_section(s)` is called it would add the section to this list.
A new method would be added 'def self.register_section` which would also add the section to this array.
So then the registration code would look like:

```
  # The order of these register_ calls matters. It determines the order of the embeddables.
  InteractivePage.register_section(HEADER_BLOCK)
  InteractivePage.register_section(nil)

  # Register additional sections of interactive page.
  # This code could (or should) to some config file or initializer, but as we
  # have only one additional section at the moment, let's keep it here.
  InteractivePage.register_additional_section({name:  CRater::ARG_SECTION_NAME,
                                               dir:   CRater::ARG_SECTION_DIR,
                                               label: CRater::ARG_SECTION_LABEL})

  InteractivePage.register_section(INTERACTIVE_BOX)
```

@dougmartin you should track this PR because some of your work was setting or checking section names.

emcelroy: Are the latest changes pretty much what you were thinking, @scytacki? I'm struggling a bit with how to handle the situation where a section is added and populated with items, then disabled. What I did seems to work and makes at least a kind of sense, but I'm not sure it's right. I haven't been able to think of a less convoluted way to do it.

Maybe I should add something to the `visible_embeddables` method that checks if the section the embeddable is in is activated instead?

(I realize there are a bunch of broken tests now. I'll work on making sure those pass after confirming this is moving in the right direction.)

scytacki: At first glance the approach seems good. However, it will break the export. Which is probably why the tests are breaking. The export is expected to include the embeddables even if their section is hidden. I'm not sure it is important to keep that behavior, but we might as well.

So then  embeddables ought to just the registered_sections for ordering and ignore if they visible.  
And visible_ embeddables would need to do what you said which is to check if the section of the embeddable is visible. 

To facilitate that check it would be efficient to store a map of section names to show methods. Something like:
```
@registered_section_show_methods = {}
def self.registered_section_show_methods
  @registered_section_show_methods
end
```

and then in `def self.register_section(s)`
you'd add:
`@registered_section_show_methods[s.name] = s.show_method`

And a helper method to make this easier to use
```
def registered_section_visible(section_name)
  self.send(self.class.registered_section_show_methods[section_name])
end
```

Then in visible_embeddables it'd be changed to:
```
embeddables.select{ |e| 
  !e.is_hidden && registered_section_visible(e.section)
}
```

One more thing to check though is how the reports: both teacher and researcher handle hidden sections. Does the report show the questions even though the section is hidden?  If these reports show the questions, then perhaps we don't want to do any of this hiding.  The reason is because the question numbers need to match the numbers in the teacher and researcher report. So if these reports are showing these questions in hidden sections, then their numbering won't match that of the activity.

emcelroy: I looked at a teacher report locally after the latest commits and things are acting a bit strangely. It's hard to keep track of because the behavior seems random at times. In most respects it seems to do what I would expect. If a section is hidden, the questions in that section don't appear in the report. And the numbering of questions in the report adjusts itself according to how many sections are shown. Things seem to get wonky when you hide all sections. In that case, it will continue to show at least some questions. It's not clear how it chooses which, but they're all from the same section. If you then activate one section, the report breaks with an error in the browser console saying "Cannot read property 'filter' of undefined." Once that happens, things can get weird. For example, I got into a state where activating the header section would make the header questions disappear from the report. If I hid the header section, the header questions would then reappear in the report.

On staging, the teacher reports show all questions all the time no matter which sections are hidden or activated. Is that the desired behavior? If so, does that mean we shouldn't be allowing hiding at all?

FWIW: In staging teacher reports, questions in the header block appear out of sequence.

scytacki: The staging (and probably production) behavior matches what I kind of expected. I think I remember it confusing authors in the past.  I think it might contribute to completion status confusion because there are some questions in sections that are hidden.

Did you also check the staging (or production) researcher report? I'd expect it to function the same way: hidden sections do not effect the list of questions shown.

If we can fix it, we might as well try. 
What might be required is that you need to make sure the activity is republished to the report service and portal, when the section is shown or hidden. I'd suspect that it might not be republished and that is causing some of the behavior you are seeing.

But also it doesn't look like you implemented by suggestion above. The code is still not including embeddables from hidden sections in the embeddables list. This means embeddables  which are in hidden sections will not be exported.  I don't think this would cause the strange behavior you documented above though.

PR #593, opened 2020-07-16
Add script that migrates docstore interactive URLs to CFM

pjanik:
[#173749425]

Depending on the action, this script generates:
- docstore_to_cfm_master.json file that includes list of objects with interactive ID, old URL, and the new URL
- docstore_to_cfm_production.json file that includes list of objects with interactive ID, old URL, and the new URL
- cfm_master_to_cfm_production.json file that includes list of objects with interactive ID, old URL, and the new URL

If action is set to `execute_migration`, it accepts one of these JSON files and just updates URLs. This will let us easily revert changes if necessary and avoid losing old URLs if there's some bug.

### DIFF
--- a/script/transform_codap_and_sage_urls.rb
+++ b/script/transform_codap_and_sage_urls.rb
@@ -1,0 +1,190 @@
+require File.expand_path('../../config/environment', __FILE__)
+require "uri"
+require "rack"
+require "json"
+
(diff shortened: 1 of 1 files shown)


pjanik:
I've tested staging so far and we have only 312 interactives to transform. I'll check production in a moment.

pjanik:
1081 on production, so it doesn't seem it'll be a problem.

PR #596, opened 2020-07-17
Added interactive list api [#172950387]

dougmartin:


app/assets/javascripts/lara-typescript.js
```diff
@@ -29314,7 +29333,9 @@ exports.ManagedInteractiveAuthoring = function (props) {
                         React.createElement("legend", null, "Url Fragment"),
                         React.createElement("textarea", { id: formField("url_fragment").id, name: formField("url_fragment").name, defaultValue: url_fragment, onBlur: handleUrlFragmentBlur })),
                     libraryInteractive.authoring_guidance
-                        ? React.createElement("div", { dangerouslySetInnerHTML: { __html: libraryInteractive.authoring_guidance } })
+                        ? React.createElement("fieldset", null,
+                            React.createElement("legend", null, "Authoring Guidance"),
+                            React.createElement("div", { dangerouslySetInnerHTML: { __html: libraryInteractive.authoring_guidance } }))
```
kswenson, line 29338: Not necessarily part of this PR since it's not a change, but elsewhere we've been using a [renderHTML()](https://github.com/concord-consortium/question-interactives/blob/master/src/shared/utilities/render-html.ts) utility function to replace use of `dangerouslySetInnerHTML`).

app/models/interactive_page.rb
```diff
@@ -143,10 +147,10 @@ def add_embeddable(embeddable, position = nil, section = nil)
     end
   end
 
-  def add_interactive(interactive, position = nil, validate = true)
+  def add_interactive(interactive, position = nil, validate = true, section = INTERACTIVE_BOX)
```
scytacki, line 150: Is this change necessary?

app/models/mw_interactive.rb
```diff
@@ -152,4 +152,9 @@ def update_labbook_options
   def prompt
     name
   end
+
+  def thumbnail_url
+    # mirror managed_interactive has thumbnail_url which proxies to library_interactive
+    nil
```
scytacki, line 158: There is an image_url which should be used here if it is set.

lara-typescript/src/interactive-api-parent/iframe-saver.ts
```diff
@@ -263,6 +266,10 @@ export class IFrameSaver {
       return this.getFirebaseJwt(request);
     });
 
+    this.addListener("getInteractiveList", (request: LaraInteractiveApi.IGetInteractiveListRequest) => {
```
scytacki, line 269: I think this iframe-saver is only used at runtime. The getInteractiveList api will be needed during authoring.

app/models/interactive_page.rb
dougmartin: Yes, this allows us pass `nil` to add the interactive to the assessment block here: https://github.com/concord-consortium/lara/pull/596/files#diff-85f32c83abf3ed472f1a05afc00bcd5eR83

scytacki: It is probably better to call add_embeddable directly in that case.

This method is only used in tests. And it shows the interactive section when it is called.
Ideally we'd remove this method and move it to a test helper.

scytacki:
I'm guessing this isn't a draft PR anymore right?

PR #603, opened 2020-07-27
Fix incorrectly handled non full-width items in header block.

emcelroy:


emcelroy:
OK, I've updated the story. Your understanding about the changes is correct.

PR #604, opened 2020-07-28
Callout text box

emcelroy:
Related PT story: https://www.pivotaltracker.com/story/show/174011825


app/assets/stylesheets/mw-runtime-base.scss
```diff
@@ -438,10 +438,7 @@ div.intra-page-nav {
   .question {
     margin: 20px;
     display: inline-block;
-    // the questions sit in two columns in the full width layout.
-    // The 438px width is computed to maximize their size:
-    // 960/2 = 480, 480 - margins = 440, then there is 2 px of padding somewhere
-    width: 438px;
+    width: 45%;
```
emcelroy, line 441: I changed this to match what works in the CC Runtime base theme. 

In MW Runtime base themes there were line breaks between what should've been side-by-side boxes in full-width layout mode. I tried to track down the cause, but could only get as far as seeing that removing one pixel of width from one of the boxes fixed the problem. There wasn't an obvious reason why and I couldn't come up with an intuitive way to accommodate it.

Using a width of 45% instead of 438px removes about 5 pixels in width from each box. That doesn't seem like much to me, but if it's a problem, I can try to spend more time figuring out where that extra pixel comes from and maybe get the widths back to 438px.

emcelroy:
> Can we use this change to remove the special handling of first text box in the header section?
> What is difference between the non-callout and this first text box?

You're right, there isn't any difference. I had mistakenly thought we needed to keep some of the special handling because the first text box was styled slightly differently, but after doing some testing I saw that it isn't. The different style rules essentially cancelled each other out. So I've completely removed that special handling in the stylesheets in the latest commit.

app/assets/stylesheets/partials/authoring/modal_edit.scss
```diff
@@ -21,6 +21,18 @@
       height: 100px;
     }
 
+    &.edit_embeddable_xhtml {
+      .textblock-heading-field {
+        display: none;
```
scytacki, line 26: I think it'd be better if this didn't completely hide the heading filed, but just disabled it. That way authors would realize they could add a heading by checking the box.

app/views/embeddable/xhtmls/edit.html.haml
```diff
@@ -11,6 +12,15 @@
   .options
     = field_set_tag "Options" do
       .ul
+        .li
```
scytacki, line 15: I'm not that familiar with this form. But I think it'd be better if the heading field was nested under the is callout box. So it is more clear that checking the box will enable the header.

scytacki: We decided to allow a header even when the item is a callout. That simplifies this form and gives the authors more options.

PR #606, opened 2020-07-29
Support editing of open response interactives in ITSI editor

emcelroy:
Related PT story: https://www.pivotaltracker.com/story/show/173380691


emcelroy:
@scytacki The changes here work though they're a bit rough. I was hoping to get your thoughts before proceeding.

Is this generally the right idea?

Should I try to more clearly separate the handling of open response interactives from the embeddable items?

scytacki:
This isn't really what I had in mind. When we were working on this a couple of weeks ago, we changed it so it was using the authoring view of the interactive. That should be less likely to break in the future. 

With your recent changes, it is dependent on the authoring state of the interactive which could change and we wouldn't remember to update this code. 

When we left it a couple of weeks ago we had an issue with the coffeescript not compiling (we thought).

emcelroy:
Thanks. I actually remembered most of that, but misunderstood a little. When I fixed the typo that was preventing the coffeescript from compiling, it still wasn't working and I ended up going down that wrong path. After kind of backing up and looking at it in light of your comments though, I think I'm back on the right track with the latest change, though it still needs a little work.

spec/services/itsi_authoring/editor_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+                                                 name: 'Test Library Interactive',
+                                                 base_url: 'https://concord.org/'
+                                                )}
+  let(:managed_interactive) { FactoryGirl.create(:managed_interactive,
+                                                 library_interactive: library_interactive,
+                                                 url_fragment: "test"
+                                                )}
+
+  before(:each) do
+    page.add_interactive(managed_interactive)
```
scytacki, line 27: Since you've gone this far, you should make a second managed_interactive, and add it to the page as an embeddable. I think that is just using the page.add_embeddable method.
That way this test will also cover liine 61 of editor.rb: 
https://github.com/concord-consortium/lara/pull/606/files#diff-f875b83b4abbc964eb099221f462f109R61

emcelroy: OK, I think that's all set.

PR #609, opened 2020-07-30
Run Cypress tests on Travis [#173706926]

dougmartin:


dougmartin:
@scytacki leaving this along for now after spending about 4 hours on it.  

The code is all in place to start the server and run the tests but we are in a dependency gotcha as you can see with all my test commits.  Ignore the greenlit builds - those actually failed but I didn't have `set -e` in the new script that runs the test so when it failed Travis thought it succeeded.  I've added some inline comments to describe the changes.

.travis.yml
```diff
@@ -3,7 +3,14 @@ language: ruby
 rvm:
   - "2.3.7"
 
-dist: precise
+dist: trusty
```
dougmartin, line 6: I had to upgrade to allow `libatk-bridge2.0-0` to be installed otherwise Cypress dies.

.travis.yml
```diff
@@ -3,7 +3,14 @@ language: ruby
...
+
+jobs:
+  include:
+    - script: bundle exec rake
+    - script: cd lara-plugin-api-V2 && npm i && npm run test
+    - script: cd lara-typescript && npm i && npm run test
+    - script: ./script/cypress-on-travis.sh || travis_terminate 1;
```
dougmartin, line 13: This has a nice side effect of running all the tests in parallel.  Without the Cypress part it cuts about 1 minute from the test run.

.travis.yml
```diff
@@ -45,3 +45,6 @@ branches:
   except:
     - staging
     - production
+
+addons:
+  chrome: stable
```
dougmartin, line 50: I did this to see if it would install dependencies which Cypress also needed - this can be removed as it didn't help.

script/cypress-on-travis.sh
```diff
@@ -0,0 +1,31 @@
...
+bundle exec rails s -b 0.0.0.0 -d
+
+# setup cypress
+cd cypress
+npm i
+bundle exec rake travis:create_user
+cp config/user-config.sample.json config/user-config.json
+
+# DEBUG INFO
+ldd /home/travis/.cache/Cypress/4.11.0/Cypress/Cypress
```
dougmartin, line 19: This is here for debugging to see which libraries are loaded by Cypress.

scytacki:
How about running cypress in docker: 
https://www.cypress.io/blog/2019/05/02/run-cypress-with-a-single-docker-command/#testing-site-on-host

dougmartin:
No longer needed as we use Github Actions now.

scytacki:
We haven't switched the main LARA over to GitHub Actions. But I agree it'd be better to do that switch instead of trying to get cypress working in travis.

PR #612, opened 2020-07-31
Add a default value for PORTAL_PROTOCOL, and error for PORTAL_HOST

scytacki:
PORTAL_PROTOCOL is defined in the environment section of the main docker-compose file,
But that doesn't make it available to the other sections of the file.
Adding a default here `:-http` prevents a warning and SSO errors when it is not defined.

Also this adds a `:?[err]`  to the PORTAL_HOST variable which tells docker it is mandatory

### DIFF
--- a/docker/dev/docker-compose-portal-proxy.yml
+++ b/docker/dev/docker-compose-portal-proxy.yml
@@ -45,7 +45,7 @@ services:
   app:
     environment:
       CONCORD_CONFIGURED_PORTALS: LOCALHOST HAS_STAGING
-      CONCORD_LOCALHOST_URL: ${PORTAL_PROTOCOL}://${PORTAL_HOST}/
+      CONCORD_LOCALHOST_URL: ${PORTAL_PROTOCOL:-http}://${PORTAL_HOST:?Need to define PORTAL_HOST}/
       CONCORD_LOCALHOST_CLIENT_ID: localhost
       CONCORD_LOCALHOST_CLIENT_SECRET: 'unsecure local secret'
       # the code requires at least 2 portals configured this second one to has Staging


docker/dev/docker-compose-portal-proxy.yml
```diff
@@ -45,7 +45,7 @@ services:
   app:
     environment:
       CONCORD_CONFIGURED_PORTALS: LOCALHOST HAS_STAGING
-      CONCORD_LOCALHOST_URL: ${PORTAL_PROTOCOL}://${PORTAL_HOST}/
+      CONCORD_LOCALHOST_URL: ${PORTAL_PROTOCOL:-http}://${PORTAL_HOST:?Need to define PORTAL_HOST}/
```
knowuh, line 48: I wasn't aware of this syntax, and even now, I am having trouble finding documentation for it.

I would expect to find something about this `:default` to be discussed in the docs, but cant find it here:
https://docs.docker.com/compose/compose-file/#variable-substitution

Or googling. 

Do you have a link to some documentation?  In particular I am wondering what the `-` or `?` mean.

knowuh:
This looks fine to me, I like the idea, and I assume it works, but I asked for a link to some docs.

docker/dev/docker-compose-portal-proxy.yml
scytacki: It isn't easy to find, but it is at the link you posted above, just scroll down a bit:
https://docs.docker.com/compose/compose-file/#variable-substitution:~:text=%24%7BVARIABLE%3A%2Ddefault%7D

(Good excuse for me to use chrome's special `:~:text=` argument. :) )

knowuh: Awesome -- and 🤯  `:~:text=` in URLs is amazing.

PR #613, opened 2020-07-31
Add duplicate/export to interactive links [#174063166]

dougmartin:


spec/services/lara_serialization_helper_spec.rb
```diff
@@ -84,6 +122,48 @@
...
     end
+
+    describe "import page with interactives linked to other interactives" do
+      let(:export_data) do
+        {type: "InteractivePage", ref_id: "100-InteractivePage", embeddables: [
+          {embeddable: {type: "MwInteractive", ref_id: "100-MwInteractive", linked_page_items: [{:primary_ref_id=>"100-MwInteractive", :secondary_ref_id=>"101-ManagedInteractive", :label=>"one"}, {:primary_ref_id=>"100-MwInteractive", :secondary_ref_id=>"102-MwInteractive", :label=>"two"}]}, section: "interactive_box"},
+          {embeddable: {type: "ManagedInteractive", ref_id: "101-ManagedInteractive", linked_page_items: []}, section: "interactive_box"},
+          {embeddable: {type: "MwInteractive", ref_id: "102-MwInteractive", linked_page_items: []}, section: "interactive_box"},
+          {embeddable: {type: "ManagedInteractive", ref_id: "103-ManagedInteractive", linked_page_items: [{:primary_ref_id=>"103-ManagedInteractive", :secondary_ref_id=>"100-MwInteractive", :label=>"three"}]}, section: "interactive_box"}
+        ]}
```
scytacki, line 133: Can you clean this up a bit?

Better indenting would be helpful and not mixing `key:` with `:key=?` would also help.

app/services/lara_serialization_helper.rb
```diff
@@ -21,6 +21,18 @@ def export(item)
       # interactive missing. I guess it could be double checked and fixed in the future.
       results[:linked_interactive] = self.export(item.linked_interactive)
     end
+
+    # This is done here and not in the models as we don't want to export the linked embeddables as they are already exported
+    # we only want to save the key so that the references to the embeddables can be restored during import.
+    # The embeddedable references are saved instead of the page_item references as new page_items are created when the
+    # embeddables are imported into the page.
+    if item.respond_to?(:primary_linked_items)
+      results[:linked_page_items] = item.primary_linked_items.map {|pli| {
```
scytacki, line 30: Since this export is what is used by the activity player as its activity structure format. I think it is worth keeping it more independent of the implementation.

So I'd recommend changing this from `results[:linked_page_items]` to `results[:linked_interactives]` I know this is going to be somewhat confusing since there is also linked_interactive. But that is the way it is called in the new field of the initInteractive message, so using the same name here seems better.

Additionally the `primary_ref_id` doesn't seem necessary, and would likely confuse a developer working with the export.  Without the primary_ref_id, then the `secondary_ref_id` could be changed to `interactive_ref_id` which would be consistent with other references to interactives.

dougmartin:
@scytacki I added 3 commits:

1. The first addresses your PR comments: https://github.com/concord-consortium/lara/pull/613/commits/f009fa34bfc259a64401c72da5de88d6da89249b
2. The second cleans up importing to move deleting the `linked_interactives` key when importing the polymorphic class from the model back into the serialization helper import (where other keys are similarly deleted): https://github.com/concord-consortium/lara/pull/613/commits/b7760f06c71c948a6b5f2dce68159135c5a566d1
3. Renamed the renamed `interactive_ref_id` in #1 to `ref_id` so that it isn't confused with the top level interactive_ref_id and it matches the "schema" of each of the other elements that are serialized.

scytacki:
Looks good to me.
One thing I just remembered was preserving the order of the array of linked_interactives. I'd guess it works now, but it might not be guaranteed since I think it is relying on the database ordering. But that can be a story for another day.

PR #615, opened 2020-08-07
Fix a bug in lara-interactive-api hooks causing issues in Safari

pjanik:
This PR fixes an issue in Safari: https://www.pivotaltracker.com/story/show/174154314

`useAuthoredState` (also other `use...State` hooks) was getting the initial value of authoredState from the Client during the first render (the first hook function execution). Later, it was adding authoredState observers in `useEffect(..., [])` callback. Since useEffect is executed asynchronously, probably with some small delay, there was a period of time between getting initial authoredState value and setting up observers when authoredState was not observed. And sometimes the update was happening exactly in this short period (Safari only, around 50% of cases). 

The fix was to ensure that we read the most recent version of authoredState right before setting up observers in the same synchronous function. So there's no possibility now that some updates will be missed.

I hope comments should be clear enough. I've also added a spec that tests against this specific case.

I've updated NPM package version to 0.5.0 without pre- suffix, so we can start using normal version numbers.

### DIFF
--- a/docs/interactive-api-client/README.md
+++ b/docs/interactive-api-client/README.md
@@ -1,6 +1,6 @@
-[@concord-consortium/lara-interactive-api - v0.5.0-pre.4](README.md) › [Globals](globals.md)
+[@concord-consortium/lara-interactive-api - v0.5.0](README.md) › [Globals](globals.md)
 
-# @concord-consortium/lara-interactive-api - v0.5.0-pre.4
+# @concord-consortium/lara-interactive-api - v0.5.0
 
 ## [API documentation](globals.md)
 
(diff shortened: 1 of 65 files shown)


pjanik:
> Most of the comments are copy and pasted and refer to authoredState when really it is interactiveState or globalState.

Ops, right. The bug was only about authoredState, but it could affect interactiveState and globalIntState too. Comments in the hooks.ts are correct I think, but I forgot to fix ones in hooks.test.tsx. I'll fix them.

> Also when referring to an "update" happening in between the useState and the useEffect, do you mean that the interactive using the client is updating the state? It wasn't clear to me what was updating the state.

In this particular case, the update was caused by lara-interactive-api which received `initInteractive` message. But generally it doesn't really matter what is updating authoredState/interactiveState/globalInteractiveState. The problem was that there's a time when this state wasn't observed by the hook. I can add a note that it was client and init interactive message.

pjanik:
I've updated the comments. I've tried to clarify one in the main code, removed copy-paste comments there (since we have tests anyway), and fixed ones in spec file.

PR #619, opened 2020-08-20
Fix carousel

emcelroy:
The latest changes address this comment: https://www.pivotaltracker.com/story/show/173992930/comments/217204382


scytacki:
@emcelroy how is the sizing handled in the carousel?

Since the interactives can request more height while the user is working with them, does the carousel also get taller when that happens? A good way to test this is to add an open response interactive and then drag the text-area bottom left handle to make it bigger.

scytacki:
After discussion with Leslie and Ethan we decided to punt the height issue. I made a new story for it: https://www.pivotaltracker.com/story/show/174426725

scytacki:
@emcelroy is this PR all set now?

emcelroy:
@scytacki Yeah, it's ready. Thanks.

PR #622, opened 2020-08-21
use MutationObserver to only show the plugin question when it adds co…

scytacki:
The story describes several options. I went with the Mutation observer for the best flexibility backward compatibility.

https://www.pivotaltracker.com/story/show/174448841

### DIFF
--- a/app/views/plugins/_show.haml
+++ b/app/views/plugins/_show.haml
@@ -122,16 +122,18 @@
     -# we are observing changes before the plugin can modify the content
     :javascript
       (function ($plugin_output) {
+        var $plugin_question = $plugin_output.closest(".question");
+
         // hide the parent question until the plugin adds content
-        $plugin_output.closest(".question").addClass("hidden");
+        $plugin_question.addClass("hidden");
 
         // show the question when the plugin adds content, hide it again if the plugin
         // removes all of its content
         var observer = new MutationObserver(function (mutationsList, observer) {
           if($plugin_output.contents().length !== 0) {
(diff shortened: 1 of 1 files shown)


scytacki:
@mklewandowski you might want to look at this solution too since you just added plugin support to the activity-player. I haven't tested it or looked at how it handles the padding around teacher-edition plugins when they decide not to render any content. Removing that padding is what this story is about. 

The other case handled by this story is when the items in section are in two columns (half width), the plugin items shouldn't take up a 'slot' if they don't have any content.

scytacki:
@emcelroy good questions. 

Yes it is possible the plugin could remove all of its content from the page. I don't think we have a plugin that does this, but perhaps a plugin would show a global button which toggles its content (instead of the teacher-edition URL flag), so then being able to hide the padding around the content after it has been shown is good to support.

I think output_id makes the code easier to read. The two choices are:
`"##{output_id}"` or `"#{runtime_div_selector}"`
The first is shorter and makes it more clear, we are working with the output.  The runtime_div_selector's main usage is for getting the div to pass to the plugin through its container property here:
https://github.com/concord-consortium/lara/blob/fab03812c84a84222c4c219d7c2d6c2b45d906c9/app/views/plugins/_show.haml#L163

But since you saw this, I think it make senses to use a javascript variable to replace:
`$("##{output_id}")` I'll give that a try...

scytacki:
OK I cleaned it up more to reduce redundancy and hopefully clarify some of the comments.

pjanik:
It looks good to me too 👍

pjanik (approved):

PR #629, opened 2020-09-18
Launch Activity Player in Preview-Mode from LARA

emcelroy:
Related PT story: https://www.pivotaltracker.com/story/show/174718757

[#174718757]


scytacki (approved):
Looks good

PR #631, opened 2020-09-23
Add support for custom messages

kswenson:
- provides API for interactives to listen for custom messages
- provides `useCustomMessages` hook to simplify listener management
- expands supported features API to allow configurable support for custom messages
- adds (as-yet-untested) code to example interactive to display custom messages received

[[#173989045]](https://www.pivotaltracker.com/story/show/173989045)

[ActivityPlayer #60](https://github.com/concord-consortium/activity-player/pull/60) contains the corresponding ActivityPlayer changes.

Note that while there is no plugin code using these features yet I'm opening up the PR for review so as to get feedback on the implementation of the API and whether there's anything I missed.

### DIFF
--- a/lara-typescript/src/example-interactive/src/components/runtime.tsx
+++ b/lara-typescript/src/example-interactive/src/components/runtime.tsx
@@ -1,6 +1,6 @@
 import * as React from "react";
 const { useEffect, useState } = React;
-import { IRuntimeInitInteractive, getFirebaseJwt } from "../../../interactive-api-client";
+import { IRuntimeInitInteractive, getFirebaseJwt, useCustomMessages, ICustomMessage } from "../../../interactive-api-client";
 import { IAuthoredState } from "./types";
 
 interface Props {
@@ -22,6 +22,11 @@ export const RuntimeComponent: React.FC<Props> = ({initMessage}) => {
     }
   }, [authoredState]);
 
+  const [customMessages, setCustomMessages] = useState<ICustomMessage[]>([]);
+  useCustomMessages((msg: ICustomMessage) => {
+    setCustomMessages(messages => [...messages, msg]);
+  });
+
   return (
     <div className="padded">
       <fieldset>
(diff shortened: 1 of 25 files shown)


lara-typescript/src/example-interactive/src/components/runtime.tsx
```diff
@@ -22,6 +22,11 @@ export const RuntimeComponent: React.FC<Props> = ({initMessage}) => {
     }
   }, [authoredState]);
 
+  const [customMessages, setCustomMessages] = useState<ICustomMessage[]>([]);
+  useCustomMessages((msg: ICustomMessage) => {
```
scytacki, line 26: It would be easier for developers (I think), if this hook also took an argument of the custom messages supported by the interactive. Then the hook can take care of adding those to the supportedFeatures.  That way the developer doesn't need to worry about the details of how those supported custom messages are published via the api.

However this automated approach might complicate how setSupportedFeatures is used. For example in the question-interactives repo here: https://github.com/concord-consortium/question-interactives/blob/6c0453518d15ea30e8e88a1ea84c1fa9d6a13485/src/image/components/runtime.tsx#L37

scytacki (changes requested):
Over all looks good. It seems like the example interactive is not specifying its supported messages in supportedFeatures. And I suggested a way to make that easier (and perhaps less error prone). 

However if you want to punt that supportedFeatures change to another PR, I'm fine with merging this as is. Just let me know.

lara-typescript/src/plugin-api/events.ts
```diff
@@ -29,4 +37,15 @@ export const events = {
    * Removes InteractiveAvailable event handler.
    */
   offInteractiveAvailable: (handler: IInteractiveAvailableEventHandler) => offInteractiveAvailable(handler),
+  /**
+   * Subscribes to InteractiveAvailable events. Gets called when any interactive changes its availablity state.
```
scytacki, line 41: This comment and the one below should be updated.

scytacki (changes requested):
I found one comment that was updated after it was copy and pasted. Otherwise it looks great.

lara-typescript/src/plugin-api/events.ts
kswenson: Fixed

scytacki (approved):
Looks good.

lara-typescript/src/example-interactive/src/components/runtime.tsx
kswenson: Done

PR #632, opened 2020-09-24
Fix a bug causing weird behavior when two embeddables with different types have the same ID

pjanik:
More details here:
https://www.pivotaltracker.com/story/show/174967192
It was affecting authoring page, hide/show and remove links.

Also, I've also updated example-interactive.md docs (not related to the issue above, just to skip another branch and PR for this tiny fix).

### DIFF
--- a/app/controllers/interactive_pages_controller.rb
+++ b/app/controllers/interactive_pages_controller.rb
@@ -133,7 +133,7 @@ class InteractivePagesController < ApplicationController
   def remove_embeddable
     authorize! :update, @page
     update_activity_changed_by
-    PageItem.find_by_interactive_page_id_and_embeddable_id(params[:id], params[:embeddable_id]).destroy
+    PageItem.find_by_interactive_page_id_and_embeddable_type_and_embeddable_id(params[:id], params[:embeddable_type], params[:embeddable_id]).destroy
     # We aren't removing the embeddable itself. But we would remove the tracked_question of the embeddable.
     redirect_to edit_activity_page_path(@activity, @page)
   end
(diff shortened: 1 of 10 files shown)


scytacki (changes requested):
Could you change the remove to use the page_item id directly? That will simplify the route.
I think that was the change the @dougmartin made in some other places. 
And perhaps he just missed these cases.

I'll try to find that PR to see what was changed...

scytacki:
@pjanik I think this was the PR that included some of these page_item changes: https://github.com/concord-consortium/lara/pull/601
(it is also the PR adding the linked interactives support, relevant to your current work)
https://github.com/concord-consortium/lara/pull/601

pjanik:
@scytacki, I'm not sure if I see a connection with this other PR. I can see that there has been added `interactive_item_id`, but it won't really be useful here. Or did you mean something else? It's pretty large, so maybe I've missed something.

This PR in fact fixes two separate issues:

1. `app/views/shared/_embedded_editor_links.html.haml` and the related controller was using Page ID and Embeddable ID to toggle visibility or remove page items. This was obviously not enough, Embeddable type was missing.
2. ` app/views/mw_interactives/_author.html.haml` was also broken due to a similar, but different reason - it's been used both by MwInteractive and ManagedInterative, but it was using DOM IDs that were not unique for these two different types.


Fix of the 1st issue can't use `interactive_item_id` as it needs to work with all kinds of embeddables, including old style questions, or even plugins. I've just refactored it to use page_item.id instead, and also updated related controllers, but I'm not sure if you meant this approach. I totally removed page_id from routes, as it was redundant after using `page_item_id`. Someone could post ID of a different page to pass authorization and remove page item from a different page that he has no access too. ;)

Fix of the second issue could use `interactive_item_id`, but actually `embeddable_dom_id` feels better to me (as it's actually DOM ID). I'd say it's the same though. I can update it if this new ID is better for some reason.

scytacki (approved):
Looks great to me. 
Sorry for the confusion about the big PR. I wasn't totally sure if it was related.

PR #634, opened 2020-09-25
Provide linkedInteractives in initMessage in the authoring mode

pjanik:
[#172781754]

`linkedInteractives` was provided only in the runtime so far. It's useful/necessary in the authoring as well, so the authoring interface can access the current linked interactives list and set UI widget values based on that.

Semi-related changes:
- I'm a bit lost in all the interactive hashes, but I tried to move ones that were identical to BaseInteractive class.
- I've removed one unused property in authoring initialization. Now, it seems like `ManagedInteractive#to_interactive` could be easily removed. It's only called by the `#to_authoring_preview_hash`. I was also wondering if `#to_authoring_preview_hash` could be unified in MwInteractive and ManagedInteractive.


scytacki (approved):
Looks good to me

PR #635, opened 2020-10-02
Add snapshot API

pjanik:
This PR implements getInteractiveSnapshot in interactive-api-client. There are some changes to the response message format and types.

Also, I've implemented actual snapshot support in LARA, as it was very simple.


scytacki (changes requested):
Can you update the example interactive so this api can be tested without needing to use the interactives in question-interactives?

Otherwise this looks great to me. 
Thanks for implementing getSnapshot in LARA runtime!
I think then the only missing API in the LARA runtime is the lightbox API.

scytacki:
Docs on the example interactive are here: https://github.com/concord-consortium/lara/blob/master/docs/example-interactive.md

pjanik:
@scytacki, I've updated example-interactive and extended a bit API implementation.

btw, if you're testing this manually, quite often snapshots return blank / white image. It's around 20-30% of attempts. This seems to be a problem with Shutterbug server itself. I think I remember that someone has reported this problem some time ago. I haven't looked at it, as it's a separate problem, and probably a bigger story.

scytacki (approved):
Looks good.

PR #636, opened 2020-10-02
Fix more issues related to non-unique IDs, add note about existing bug

pjanik:
I haven't actually seen a problem related to editing, but it seems it could be theoretically problematic. I've made changes there as it's fairly safe.

Also, I've added a comment in `interactive_run_helper.rb` as this is a bigger problem. This interactive ID definitely doesn't seem to be unique for MwInteractive and ManagedInteractive. I couldn't easily fix it, as it's not that obvious to track all the places that use it. Ones I've found so far:
- click to play
- drawing tool snapshot

The problem is very unlikely to happen in production, so I'm not spending time on it for now. I've just added this comment and a story in the icebox: https://www.pivotaltracker.com/story/show/175098853


scytacki (approved):
Looks good.

PR #638, opened 2020-10-05
Add hostFeatures to InitInteractive message

kswenson:
[[#174886713]](https://www.pivotaltracker.com/story/show/174886713)

Note: Following @pjanik's example, I meant to publish `@concord-consortium/lara-interactive-api@0.6.0-pre.2` with the `beta` tag, but I inadvertently published it without a tag so it was published with the `latest` tag instead. 🤦 To make this kind of mistake less likely in the future, I added scripts for publishing beta versions of the plugin-api and interactive-api. I'll republish without the `-pre.2` once the PR is merged.


pjanik (approved):
Looks good to me 👍

PR #639, opened 2020-10-08
Interactive API: Modal API cleanup + flushStateUpdates

pjanik:
[#175165872]

@scytacki, a couple of changes mostly related to the Interactive API that we've discussed recently:

- `flushStateUpdates` immediately sends interactive state to parent. Useful while dealing with modals. It's exported as part of the API, but also automatically called by `showModal` and `closeModal`, so interactives don't have to remember about it.
- I've removed `DialogState` generics and `mode === "dialog"`. Instead, we're planning to rely on URL params for now, as they cover all of our use cases.
- I've tried to clean up confusion about modal type "lightbox" and modal type "dialog". There are new comments that explain the difference in a few places. To sum up:
   - Lightbox is meant to display any random URL. Possibly image URL. It does NOT assume that this URL is interactive. It doesn’t try to use iframe-phone, nor initialize any iframe.
   - Dialog is meant to display dynamic interactive (or part of its UI based on the dialog context). This can be easily used instead of Lightbox, but requires a bit more work (but not much more). Lightbox is a shortcut to make implementation a bit simpler and make image zooming UI consistent.
- I've made `uuid` optional in all the modal requests. Usually interactive uses only one dialog or lightbox, and doesn't manage this uuid anyway. It's still there, as it might be necessary when we combine dialogs and lightboxes.

I'm using this locally in question-interactives API and it seems to be working fine for me. I've also implemented basic lightbox support in ActivityPlayer. 

@kswenson, you might check this PR, as this will affect `hostFeatures` that you've been working on. I think now we'd need to update it to something like:

ActivityPlayer:
```
hostFeatures: {
  modal: {
    version: "1.0.0",
    dialog: true,
    lightbox: true,
    alert: false
  }
},
```

LARA:
```
hostFeatures: {
  modal: {
    version: "1.0.0",
    dialog: false,
    lightbox: true,
    alert: true
  }
},
```

I'd avoid using `modalDialog` as the dialog is one of our types of modals. Although I'm sure you can solve linguistic problems better. ;)


scytacki (changes requested):
Looks good to me.
My only concern with it is the uuid change. Lets say an interactive opens a modal:dialog and then that dialog wants to show an alert. In that case the interactive will need to use a uuid for the alert or the initial modal:dialog. That doesn't seem too bad.  Like you said I think the main issue will happen if there are multiple interactives opening modals at the same time. 

So I guess my concern is that it will be hard to fix this in the future since I think it means we'll have to update all interactives that are opening modals.  

Would it possible to randomly assign the uuid in the interactive-client if it isn't provided by the interactive?  That way we wouldn't have conflicts in the future.  But in that case I'm not sure how the modal:dialog view of an interactive would know what this id is in order to close itself unless it was passed the id in a URL param. But if the interactive-api-client is randomly assigning this id the interactive code won't know what this random id is in order to add to as a URL param.

kswenson (approved):
👍 LGTM other than the issue raised by Scott about potential for conflicting IDs if there were ever modals from multiple interactives using the same `defaultId`.

pjanik:
Thanks. Right, I might have not checked LARA implementation too precisely. The last commit refactors modal-api-plugin.ts.

Generally, I'm still leaving `uuid` optional. I think it should be the host responsibility to:
- ensure that an interactive can't open modal with conflicting UUID. It'd say that UUID is more like ID in fact, and we shouldn't rely on the interactive to make it globally unique.
- ensure that an interactive can't close modal that belongs to another interactive, even it provides a matching UUID or UUID is skipped.

I'm not renaming it to ID not to break backward compatibility. And one day we might want to change the rules and let interactives to control modals that belong to other interactives. It doesn't seem useful for me, but who knows, there might be some use case.

Note that these conditions are easy to implement - the host has separate iframe-phones for each interactive. So it always knows from which interactive the message is coming. They don't really mix up. So, the host can keep a separate modal list for each iframe interactive or generate UUID based on the interactive instance. It's an implementation detail. It's like that in ActivityPlayer, but didn't realize LARA was keeping a global list of modals.

The last commit changes this implementation and makes modals list local to plugin / iframe-saver instance. So, no conflicts are possible anymore. I've slightly changed the `IFrameSaverPlugin` interface so it can return more things (`#hasModal` method that is only used by tests).

>  In that case the interactive will need to use a uuid for the alert or the initial modal:dialog. That doesn't seem too bad. 

Right. My idea was that as long as you use one modal, you don't have to care about UUIDs. When you open multiple ones, but you rely on the host close action (e.g. clicking background or a default X icon), you still don't have to care about UUID. It'll be only necessary when you need custom controls that close the modal.

scytacki:
So in the activity player if an interactive opens a dialog, and then wants to manually close itself. Does it need to know this id?
And if it the ids are scoped to the interactive do the two 'instances' of the interactive (inline and dialog) share this scope? 

I could imagine a case where the interactive in the dialog uses a new instance of some component, so then the there are two scopes, one for the inline interactive and one for the dialog interactive. 

I suspect you tested that, and it would be a change in the activity player.

Assuming that all works correctly this approach of scoping the ids seems great to me. 
We can open up cross interactive dialog management later as it is needed.

scytacki (approved):
👍

pjanik:
@scytacki,
> So in the activity player if an interactive opens a dialog, and then wants to manually close itself. Does it need to know this id?

No, it doesn't need to know ID if it uses only one modal. And that's what I'd recommend for now (one modal only), as:
- LARA supports only one lightbox **globally** (for all the interactives) and multiple alerts **per interactive**
- AP supports one lightbox and one dialog **per interactive**, and no alerts at all

It doesn't seem to have multiple dialogs and/or lightboxes as they pretty much the whole screen. So, it won't be the best UX. It's more likely we'll combine modal and alerts, but then it's not that likely that we need to close alerts manually.

> And if it the ids are scoped to the interactive do the two 'instances' of the interactive (inline and dialog) share this scope?

Yes, modals are actually implemented in the parent of `IframeRuntime` component (called `ManagedInteractive`). And this parent component moves`IframeRuntime` between inline rendering and dialog. But even if it renders two `IframeRuntime` at the same time, the context will be shared.

> I could imagine a case where the interactive in the dialog uses a new instance of some component, so then the there are two scopes, one for the inline interactive and one for the dialog interactive.

That's fine (see above). We'll still have just one `ManagedInteractive` component with dialogs and lightboxes inside.

scytacki:
That all sounds great. Thanks!

PR #640, opened 2020-10-12
Interactive API updates

pjanik:
1. Add .notCloseable option to modal dialog options. Not sure if this name is the best. But the idea is that host should disable dialog closing when user clicks backdrop and also it shouldn't render X in the corner. The only option to close the dialog is by calling `closeModal` manually from the interactive.
2. I've refactored host features a bit to reflect better the current status. Actually, no interactive use them now, as I've also refactored image interactive to always use lightbox (so no decision making based on host features). BTW, I'm not sure if this idea is going to work well. If this `hostFeatures` hash keeps growing, it might become annoying for some "light" hosts like Scaffolded question or authoring preview runtime. The last one is especially tricky, as we don't want interactives to detect its poor feature list and change rendering - as it's designed to be an accurate preview. 
3. I've added ImageQuestion Authoring and Runtime Metadata types necessary for exporting to report service and Portal.


scytacki:
I was thinking the host features would help with the authoring preview runtime. Right now the interactive doesn't know what it can do or not in that view. So it likely results in broken UI.  I think as long as the authoring preview doesn't support the features the interactive needs to know so there is less broken-ness. However if interactives then start changing they views so they look different when there are missing features, then we'd have to add some new mode or property so the interactive knows it should render the same and just not try to call these actions.

For now, though I think it should be sufficient if Interactives always try to render the same and just use the supported features to show messages when users try to use some rendered button or link that isn't supported.

Is there an example now where we want an interactive to render differently based on the supported features?

scytacki (approved):
Looks good to me.

notClosable seems fine

PR #641, opened 2020-10-13
Publish interactives that pretend to be an image question to Portal and Report Service

pjanik:
There are two things that might be a bit confusing here:
- drawing prompt vs answer prompt - it seems that the author of the new interactive question named this property that way. It seems to make more sense, as it's placed next to the answer textarea. I left it that way, but now we have quite a few places where drawing_prompt == answer_prompt. I guess it's fine.
- This code can't be tested in practice right now. New Image Question interactive doesn't work in LARA. But unit tests were easy to add, as the related code is tested. I thought it makes sense to keep the implementation complete and don't have to think about which host (LARA, AP) does what transformations.

[#174824371]
[#174069081]


scytacki (approved):
Looks good to me.

PR #642, opened 2020-10-19
Copy activity/sequence and convert core question types to interactives

emcelroy:
- Update import and sequences API controllers.
- Add conversion script URL and template activity ID env vars.
- Add "convert" links in appropriate places.

[#174824504]

### DIFF
--- a/app/helpers/lightweight_activity_helper.rb
+++ b/app/helpers/lightweight_activity_helper.rb
@@ -67,4 +67,11 @@ module LightweightActivityHelper
   def activity_player_page_url(activity, page)
     return  "#{activity_player_url(activity)}&page=#{page.position}"
   end
+
+  def activity_player_conversion_url(activity)
+    if @sequence
+      return "#{ENV['CONVERSION_SCRIPT_URL']}?resource_name=#{url_encode(@sequence.title)}&lara_root=#{url_encode('https://app.lara.docker')}&lara_resource=#{url_encode(api_v1_activity_url(@sequence.id))}.json&template=#{url_encode(api_v1_activity_url(ENV['CONVERSION_TEMPLATE_ID']))}"
+    end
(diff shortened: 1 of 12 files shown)


docker-compose.yml
```diff
@@ -22,10 +22,14 @@ services:
       SHUTTERBUG_URI:
       # Enables logging to stdout instead of file.
       RAILS_STDOUT_LOGGING: "true"
-      #set to https in .env
+      # set to https in .env
```
emcelroy, line 25: @scytacki I'm guessing I should have left this out of the commit. Is that right?

app/helpers/lightweight_activity_helper.rb
```diff
@@ -67,4 +67,19 @@ def activity_player_url(activity)
...
+    else
+      params[:lara_resource] = "#{url_encode(api_v1_activity_url(activity.id))}.json"
+      params[:resource_name] = activity.name
+    end
+
+    url = ENV['CONVERSION_SCRIPT_URL']
+    uri = URI.parse url
+    uri.query = URI.encode_www_form URI.decode_www_form(uri.query || "").concat(params.to_a)
+    return uri.to_s
+  end
```
emcelroy, line 84: @scytacki Is this pretty much what you intended for adding the params to the conversion script URL?

scytacki: The way this is done mostly in the portal is like this:
https://github.com/concord-consortium/rigse/blob/60fdf4f894316382af1ab16925c3ef0eb1385ca1/rails/app/controllers/external_activities_controller.rb#L57-L62

That seems like the best way to go.

scytacki (changes requested):
Just the one change needed with the parsing of the URL.

I also noticed this contains code that is duplicated in the other open PR. So once that one is merged it'd probably be best to merge master into this one and resolve any conflicts.

scytacki (approved):
This looks good to me.
It is still marked draft though, is there more changes you want to add to it?

It could use some more tests on `activity_player_conversion_url()`, but I don't think it is big deal if they are missing in this case.

scytacki (approved):
Looks good to me.

PR #643, opened 2020-10-22
Add Preview in Activity Player link to sequence authoring

emcelroy:
Related PT story: https://www.pivotaltracker.com/story/show/175289804

- add new API for exporting sequence JSON
- add preview link to sequence edit page

[#175289804]

### DIFF
--- a/app/helpers/sequence_helper.rb
+++ b/app/helpers/sequence_helper.rb
@@ -1,0 +1,6 @@
+module SequenceHelper
+  def activity_player_sequence_url(sequence)
+    sequence_api_url = "#{api_v1_sequence_url(@sequence.id)}.json"
+    return  "#{ENV['ACTIVITY_PLAYER_URL']}/?sequence=#{CGI.escape(sequence_api_url)}&preview"
+  end
+end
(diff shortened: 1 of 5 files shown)


app/helpers/sequence_helper.rb
```diff
@@ -0,0 +1,6 @@
+module SequenceHelper
+  def activity_player_sequence_url(sequence)
+    sequence_api_url = "#{api_v1_sequence_url(@sequence.id)}.json"
+    return  "#{ENV['ACTIVITY_PLAYER_URL']}/?sequence=#{CGI.escape(sequence_api_url)}&preview"
+  end
```
emcelroy, line 5: Would it be best to consolidate this with the very similar `activity_player_url` method in the lightweight activity helper? If so, where would be the best place for the consolidated method? The application helper maybe?

scytacki: I think it is OK to leave these separated.

scytacki (changes requested):
Looks good.
It is missing a test of the new show action though.
There is an example of testing this for the activity version of this here:
https://github.com/concord-consortium/lara/blob/6d1e5c196d9b9e39e806cadeee55f3e57adacc57/spec/controllers/api/v1/lighweight_activity_controller_spec.rb#L10

scytacki (approved):
looks good

PR #644, opened 2020-10-27
Improve cypress import command

scytacki:
Hopefully these changes will make it easier to realize when the test has been misconfigured.
- it removes some extraneous info from the log
- adds an explicit log statement for the import command that includes he URL
- don't follow redirects on the request
- throw an error if the response status isn't what we expect

I tried using Cypress 5.5 but many specs were failing. When I went back down to 4, then it worked better.


scytacki:
@emcelroy the reason the cypress tests weren't working (at least for me) was because they were configured with http instead of https. The error message was far from obvious, so this PR is intended to help with that.

emcelroy (approved):
Looks good to me 👍

PR #645, opened 2020-10-28
Image proxy: handle CORS requests

pjanik:
[#175167437]

I tested it locally and it seems to be working.

I'm wondering if we can get that small change in production relatively soon? Otherwise, it might cause some issues while testing and working with image question.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -79,6 +79,7 @@ module LightweightStandalone
       allow do
         origins '*'
         resource '/api/*', :headers => :any, :methods => [:get, :post, :put, :options]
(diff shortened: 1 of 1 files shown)


scytacki (approved):
looks good to me

scytacki:
Once it passes the tests I'll merge it. That will get it up on lara-master.concordqa.org. Can you temporarily use that in the image-question code and then when it gets to production we can update the image-question?

It'll probably take a week or two before it would get on LARA production.

I made a new story specifically about this change in LARA to make it easier to track the release: https://www.pivotaltracker.com/story/show/175487122

pjanik:
OK, I changed proxy to lara-master. It works for me when I test it locally.

PR #646, opened 2020-10-29
Fix removal of linked interactives list

pjanik:
[#175165850]

PR #647, opened 2020-11-03
Linked Interactives API refactoring

pjanik:
JS code was issuing XHR request to LARA server immediately after receiving "setLinkedInetractives" message via iframe-phone. Setting of linked interactives was handled by `controller/api/v1/interactive_page_controller`. The result was that linked interactives were saved immediately after making any change in the authoring form. "Save" and "Cancel" buttons were ignored.

To fix that:
1. I've updated authoring JS code to embed one more hidden input with linked_interactives param instead of sending XHR request immediately.
2. I've moved linked interactives request handling from `controller/api/v1/interactive_page_controller` to `controller/interactive_controller#update`. It's a common base for `controller/managed_interactive` and `controller/mw_interactive`.
3. I've moved tests too, and extracted common pieces for `controller/managed_interactive` and `controller/mw_interactive` to make that a bit cleaner. I removed some tests that didn't test anything anyway.

A few random, more detailed comments:
- I think that `controller/managed_interactive` doesn't implement any authorization. There was a comment in tests that suggested there's some form of authorization, but I don't think it's true looking at the code. I didn't try to fix it here not to make this PR bigger.
- You can notice that I've removed `flash[:warning]` calls as flash messages are not rendered anyway. Instead, I let LARA server throw an exception and fail. This should not happen in any typical situation, so I think it's better if we get that logged and the author sees that the update fails. Previously, errors were swollen silently.
- I've simplified LARA Interactive API `setLinkedInteractives` implementation. `sourceId` isn't necessary, as it can be easily obtained by the server.
- This implementation fixes removal of the linked interactives too. So, we can skip this PR: https://github.com/concord-consortium/lara/pull/646


app/controllers/interactive_controller.rb
```diff
@@ -6,20 +6,19 @@ def edit
...
+        set_linked_interactives(JSON.parse linked_linteractives)
+      end
     end
+    @interactive.update_attributes! input_params
     respond_to do |format|
-      if @page
-        update_activity_changed_by(@activity)
-        format.html { redirect_to edit_activity_page_path(@activity, @page) }
-      else
-        format.html { redirect_to :back }
```
scytacki, line 21: Was this redirect_to back not needed anymore?

spec/controllers/image_interactives_controller_spec.rb
```diff
@@ -52,11 +52,10 @@
           expect(int.url).to eq(new_values_hash[:url])
         end
 
-        it 'returns to the edit page with a message indicating success' do
+        it 'returns to the edit page' do
           new_values_hash = { :caption => 'I made this up', :url => 'http://mw.concord.org/modeler/_assets/img/mw.png' }
           post :update, :id => int.id, :page_id => page.id, :image_interactive => new_values_hash
           expect(response).to redirect_to(edit_activity_page_path(activity, page))
-          expect(flash[:notice]).to eq('Your image was updated.')
```
scytacki, line 59: I'm surprised this changed in this PR. Was this test randomly failing?

spec/controllers/video_interactives_controller_spec.rb
```diff
@@ -51,11 +51,10 @@
           expect(int.poster_url).to eq(new_values_hash[:poster_url])
         end
 
-        it 'returns to the edit page with a message indicating success' do
+        it 'returns to the edit page' do
           new_values_hash = { :caption => 'I made this up', :poster_url => 'http://mw.concord.org/modeler/_assets/img/mw.png' }
           post :update, :id => int.id, :page_id => page.id, :video_interactive => new_values_hash
           expect(response).to redirect_to(edit_activity_page_path(activity, page))
-          expect(flash[:notice]).to eq('Your video was updated.')
```
scytacki, line 58: Strange that this was changed in this PR.

scytacki (changes requested):
I found a few places that I wasn't sure about.

app/controllers/interactive_controller.rb
pjanik: I think it was a dead code, as it would be executed only if `@page` was `nil`. I think we have only one place to edit interactives and the editor is providing page ID.

spec/controllers/image_interactives_controller_spec.rb
pjanik: No, it failed because I removed setting of the `flash[:notice]` in `InteractivesController` and `ImageInteractivesController` extends `InteractivesController`. I removed `flash` related code, as these messages were never displayed anyway and it was creating a false impression of error handling. ;)

spec/controllers/video_interactives_controller_spec.rb
pjanik: The same situation as in ImageInteractivesController above.

app/controllers/interactive_controller.rb
scytacki: I have a hunch that there is another place where interactives are edited. But hopefully I'm wrong.

spec/controllers/image_interactives_controller_spec.rb
scytacki: ok

spec/controllers/video_interactives_controller_spec.rb
scytacki: ok

scytacki (approved):

PR #651, opened 2020-11-17
Add IDataset interface

pjanik:
[#175739194]


lara-typescript/src/interactive-api-client/types.ts
```diff
@@ -404,3 +404,26 @@ export interface ILinkedInteractiveStateResponse<LinkedInteractiveState> {
...
+
+/**
+ * Interface that can be used by interactives to export and consume datasets. For example:
+ * - Vortex interactive is exporting its dataset in the interactive state
+ * - Graph interactive (part of the question-interactives) can observe Vortex interactive state
+ *   (via linked interactive state observing API) and render dataset columns as bar graphs.
+ */
+export interface IDataset {
+  type: "dataset";
+  version: 1;
```
kswenson, line 416: I take it we don't envision using version numbers like "1.1"?

lara-typescript/src/interactive-api-client/types.ts
```diff
@@ -404,3 +404,26 @@ export interface ILinkedInteractiveStateResponse<LinkedInteractiveState> {
...
+  properties: string[];
+  xAxisProp: string;
+  rows: Array<Array<(number | string | undefined | null)>>;
```
scytacki, line 419: When the rows are `JSON.stringify`d the undefineds will turn into nulls. That could be confusing for someone expecting undefined to be preserved.

lara-typescript/src/interactive-api-client/types.ts
scytacki: I guess that can't hurt. The downside I could see would be any automatic migration logic or version comparison would have to include a semantic versioning module so it would know how to handle different version.

What would 1.1 mean in a format like this? A new property is added to support a data producing interactive, but that new property would not break old data consuming interactives. Or likewise if a newer data consuming interactive needs some new property if the version bump is minor then it means the data consuming interactive would still work even if that property was not set by a data producing interactive.

pjanik: Good question. I think most of our projects use only a major version number. But we rarely update data formats and write migrations I think.

Not sure if I can imagine all the scenarios, but it seems to me that it makes sense to bump this version only when we make a **breaking** change. So, it's necessary to migrate data in the client app or just skip handling completely. I'd say it's pretty much a major version number and minor is not really useful.

If we just keep extending this interface with new properties, we don't really have to change the version number from "1" to "1.1.0", "1.2.0", etc. as long as we keep them optional (`?`). That way we don't break backward compatibility and I think no interactive would care if it's 1.0.0 or 1.1.0. It can just check for the optional prop instead.

scytacki (changes requested):

lara-typescript/src/interactive-api-client/types.ts
pjanik: Good point, I've fixed that in the last commit.

pjanik:
@scytacki, I've removed `undefined`. But do we want to change version to follow semver? I think we commented almost in the same time, so not sure if you've seen my answers to @kswenson question.

I can't see any benefits of the minor version numbers. I feel like we want to bump the version number here only after making breaking changes. But maybe I don't see some use cases.

lara-typescript/src/interactive-api-client/types.ts
scytacki: Yup seems good to me.

scytacki (approved):

PR #653, opened 2020-11-19
Add TE link to new Activity Player

emcelroy:
https://www.pivotaltracker.com/story/show/175724569

[#175724569]

* Move preview options to select menu
* Add option to preview teacher edition in Activity Player


scytacki (approved):
This seems fine to me. It looks like it will conflict with the other work on selecting a runtime. So it should probably wait until that work is merged and this PR will need to be updated.

db/schema.rb
```diff
@@ -432,6 +432,7 @@
     t.boolean  "student_report_enabled",                 :default => true
     t.text     "last_report_service_hash"
     t.boolean  "show_submit_button",                     :default => true
+    t.boolean  "activity_player_only"
```
scytacki, line 435: Is this boolean supposed to be here? This PR doesn't seem to have a migration which adds it.

db/schema.rb
```diff
@@ -681,6 +682,7 @@
     t.string   "publication_hash",         :limit => 40
     t.string   "imported_activity_url"
     t.text     "last_report_service_hash"
+    t.boolean  "activity_player_only"
```
scytacki, line 685: Ditto from above.

app/helpers/lightweight_activity_helper.rb
```diff
@@ -59,6 +59,15 @@ def complete_badge_for(activity)
     end
   end
 
+  def activity_player_url(activity, mode="")
+    activity_api_url = "#{api_v1_activity_url(activity.id)}.json"
+    return  "#{ENV['ACTIVITY_PLAYER_URL']}/?activity=#{CGI.escape(activity_api_url)}&preview&mode=#{mode}"
```
scytacki, line 64: It'd be safer to use a URL utility here instead of just combining strings. Perhaps at some point the ACTIVITY_PLAYER_URL will have have its own parameters in it which would then break the code above.

scytacki (changes requested):
I found a couple of things to update.

db/schema.rb
emcelroy: That snuck in from a different branch. I've removed it.

app/helpers/lightweight_activity_helper.rb
emcelroy: Sorry, I should've remembered that from your previous suggestion about using a URL utility elsewhere. Fixed.

scytacki (approved):
looks good. 👍

PR #654, opened 2020-11-20
Interactive API: update hostFeatures, add getFirebaseJwt

pjanik:
[#175787385]

This PR adds `getFirebaseJwt` to `hostFeatures`. I've used a name that perfectly matches the client method name, so it's clear whether this method can be used or not.

Also, I've modified `client.getFirebaseJwt` to check `hostFeatures` before it's even executed. That way the interactive wouldn't have to do it itself. Client method will work or it will immediately fail. Interactive doesn't have to care about init message, `hostFeatures` properties, and know what hostFeature properties mean. This logic could be encapsulated in client methods instead. I think this might be much more usable in practice. Other methods could even check `hostFeatures` vs their arguments, e.g.`showModal` could check whether lightbox, dialog, or alert are actually supported (I haven't done it yet, as not sure if you like the approach). 

I've added a test that checks hostFeatures for `getFirebaseJwt`. There are more updates in spec file, as `testRequestResponse` was broken - it wasn't testing anything (missing async/await and Promise.all), plus it had some other problems.

### DIFF
--- a/app/assets/javascripts/lara-typescript.js
+++ b/app/assets/javascripts/lara-typescript.js
@@ -28274,6 +28274,9 @@ var IFrameSaver = /** @class */ (function () {
                     dialog: false,
                     lightbox: true,
                     alert: true
+                },
+                getFirebaseJwt: {
+                    version: "1.0.0",
                 }
             },
             authoredState: this.authoredState,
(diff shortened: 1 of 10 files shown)


scytacki (approved):
Looks great to me.

pjanik:
Great, I'm updating the package version number to 7.0.0, and I'll publish it.

PR #657, opened 2020-11-25
Interactive API: IDataset - make xAxisProp optional, release version 0.7.1

pjanik:
[#175787385]

When xAxisProp is undefined, row indices will be used by the graph interactive.


scytacki (approved):
Looks good to me.

PR #658, opened 2020-12-03
Fix inconsistent edit menus for sequences and activities

emcelroy:
https://www.pivotaltracker.com/story/show/175995433

[#175995433]

* Change "Run" to "Preview" on sequence edit page
* Change run/preview URL to point to activity within sequence
* Remove trashcan icon from "Remove" link on sequence edit page

### DIFF
--- a/app/views/sequences/_activity_list.html.haml
+++ b/app/views/sequences/_activity_list.html.haml
@@ -8,7 +8,7 @@
           = truncate(activity.name_with_id, :length => 20, :omission => "…")
           %ul.menu
             %li.edit= link_to "Edit", edit_activity_path(activity) if can? :update, activity
-            %li.run= link_to "Run", activity_path(activity)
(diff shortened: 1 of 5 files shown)


app/views/sequences/_activity_list.html.haml
```diff
@@ -8,7 +8,7 @@
           = truncate(activity.name_with_id, :length => 20, :omission => "…")
           %ul.menu
             %li.edit= link_to "Edit", edit_activity_path(activity) if can? :update, activity
-            %li.run= link_to "Run", activity_path(activity)
+            %li.run= link_to "Preview", sequence_path(@sequence) + activity_path(activity)
```
emcelroy, line 11: @scytacki There isn't a `preview` method for sequences, so I just joined the sequence path to the activity path to form a URL for running the activity within the sequence. Is that good enough do you think? Or should we add an actual preview for sequences?

scytacki (approved):
looks good to me.

PR #662, opened 2020-12-14
Interactive API: don't throw an error outside iframe

pjanik:
[#176004704]

Old behavior wouldn't work for interactives that are meant to work both inside LARA and standalone.

Obviously, most of the features of LARA Interactive API aren't available outside LARA-like host, but interactive can use `useInteractiveState` as its state container and it works fine in both cases. And just call functions that are ignored if it's outside LARA, without having to have two separate code paths.

I've noticed that when I was working on CZI Gravity sim.


scytacki (approved):
seems fine to me

PR #664, opened 2020-12-17
175823492 interactive decoration

mklewandowski:
This PR adds the ability to send and receive text decoration messages to and from a client interactive using the LARA interactive client API.
- add `addDecorateContentListener ` and `useDecorateContent` hook so client can receive messages containing text decoration information from server
- add `decoratedContentEvent` which the client can use to send messages back to the server to handle text decoration events (right now it's really just designed to handle a click but this could be expanded in the future)
- expand the example-interactive to use these new text decoration features

Note: this PR is related to the Glossary Plugin and Activity PRs below:
https://github.com/concord-consortium/glossary-plugin/pull/79
https://github.com/concord-consortium/activity-player/pull/95


lara-typescript/src/example-interactive/src/components/runtime.tsx
```diff
@@ -31,6 +32,27 @@ export const RuntimeComponent: React.FC<Props> = ({initMessage}) => {
...
+
+    const decoratedContentClickListener = {
+      type: "click",
+      listener: (evt: Event) => {
+        const wordElement = evt.srcElement as HTMLElement;
+        if (!wordElement) {
+          return;
+        }
+        const clickedWord = (wordElement.textContent || "").toLowerCase();
+        decoratedContentEvent({type: "click", text: clickedWord});
```
kswenson, line 50:
```suggestion
        decoratedContentEvent({type: "click", text: clickedWord, bounds: wordElement.getBoundingClientRect()});
```
Based on my suggestion on the glossary-plugin PR, here's where you could pass the bounds of the element.

lara-typescript/src/interactive-api-client/api.ts
```diff
@@ -175,6 +185,10 @@ export const setHeight = (height: number | string) => {
   getClient().post("height", height);
 };
 
+export const decoratedContentEvent = (msg: IDecoratedContentEventMessage) => {
+  getClient().post("decoratedContentEvent", msg);
+};
```
kswenson, line 190: The name `decoratedContentEvent` makes me think it _is_ an event, as opposed to a function which sends an event. Perhaps `postDecoratedContentEvent()`?

lara-typescript/src/interactive-api-client/types.ts
```diff
@@ -297,6 +299,11 @@ export interface ICloseModal {
   uuid?: string;
 }
 
+export interface IDecoratedContentEventMessage {
+  type: string;
+  text: string;
+}
```
kswenson, line 305: As per my earlier suggestion, this could include:
```typescript
  bounds?: DOMRect;
```

kswenson (approved):
👍 Looks good -- just a couple of ideas to consider.

lara-typescript/src/interactive-api-client/types.ts
mklewandowski: I like it.  Added.

lara-typescript/src/interactive-api-client/api.ts
mklewandowski: I like this name better.  I was a bit uncertain about a few of the name choices - I think this makes things clearer.  Fixed.

lara-typescript/src/example-interactive/src/components/runtime.tsx
mklewandowski: I like it.  Added.

lara-typescript/src/interactive-api-client/client.ts
```diff
@@ -132,6 +132,14 @@ export class Client {
     return this.removeListener("customMessage");
   }
 
+  public addDecorateContentListener(callback: ICustomMessageHandler) {
```
scytacki, line 136: I wonder why this is using `ICustomMessageHandler`.  Since this is listening for its own event it seems it should have its own callback type.

mklewandowski:
Similar to the glossary PR, I'm going to merge this in hopes of moving this work forward.  Note, that I did have conflicts with the auto-generated index.js file which resulted in a force push on this branch to resolve the conflicts and get the commits up.  Normally, I would not do that once a PR is opened.

lara-typescript/src/interactive-api-client/types.ts
```diff
@@ -297,6 +299,12 @@ export interface ICloseModal {
   uuid?: string;
 }
 
+export interface IDecoratedContentEventMessage {
```
scytacki, line 302: Based on other Message Interfaces, the `Message` suffix should be dropped from this interface name. The name of interface seems to match the name of the message itself.  Unless it is a message which requires a response then it seems to be have Request and Response suffixes.

scytacki (changes requested):
Some name changes.

lara-typescript/src/interactive-api-client/client.ts
mklewandowski: Yeah, I think that is wrong.

lara-typescript/src/interactive-api-client/types.ts
mklewandowski: Got it.  Will change.

lara-typescript/src/interactive-api-client/client.ts
mklewandowski: I'm going to have to study this a little more.  I'm quite confused by the types.

mklewandowski: Should it just be an iframePhone.ListenerCallback?  Or should it use something that fits the message?
```
export interface ITextDecorationInfo {
  words: string[];
  replace: string;
  wordClass: string;
  eventListeners: IEventListeners;
}
```

mklewandowski: addListener inside the client wants an argument of type iframePhone.ListenerCallback for the second argument.  So maybe I'm not understanding how to set up a listener with a different type.

mklewandowski: Ah, so I assume you mean the type that is passed into the callback itself.

lara-typescript/src/interactive-api-client/types.ts
mklewandowski: Fixed.

lara-typescript/src/interactive-api-client/client.ts
mklewandowski: Ok, I think I sorted it out.  The last commit should have added types.

scytacki: Sorry I don't know the details either. It just struck me as wrong that an `ICustomMessageHander` was being passed to a `addDecoratedContentListener`.   customMessage is a special iframe message name/type that is different than the a decorateContent or decoratedContentEvent iframe messages.

mklewandowski: Check out the last commit.  I have to jump on a meeting, but I think it's sorted out.

lara-typescript/src/interactive-api-client/types.ts
```diff
@@ -297,6 +299,21 @@ export interface ICloseModal {
...
+  type: string;
+  text: string;
+  bounds?: DOMRect;
+}
+
+export interface ITextDecorationInfo {
+  words: string[];
+  replace: string;
+  wordClass: string;
+  eventListeners: any;
```
scytacki, line 312: This looks better, but what about these listeners?

If I understand this right, an interactive would add a ITextDecorationHandler to the client to be notified when the host (AP) wants to decorate content. So then I'd guess the eventListeners here are a kind of 'proxy listener' which need to be notified when the user clicks (or whatever action) on the decorated content. It is a 'proxy listener' because what it does is forward this click action up to the host (AP). 

So it seems to good that the listeners here should be typed too. So some interactive which supports decorated content knows what it should send to these listeners.

scytacki (changes requested):
needs a little more typing

lara-typescript/src/interactive-api-client/types.ts
mklewandowski: Good point.  When the AP passes this down, we really only care about the `type` that is inside of the object.  As discussed elsewhere, we can't actually pass functions through post message so we're actually adding the listeners locally based on the type that we get inside of the `eventListeners` object.  I'm thinking that what this really should be is something like:
`listenerTypes: string[]`
At present all we handle is "click", but we could imagine a `listenerTypes` array also containing a type for hover or active or double click or whatever else someone wants to do.  It would be up to the client to create the listener, but it would know that if it sent a message back up to the server when the event gets handled, that the server would do something with it.

mklewandowski: Although maybe we should preserver the function so the type can be exported and used elsewhere.  Not sure.  It would have to be optional.

mklewandowski: Although since this type is being created here for use in the messaging, adding a listener function (that will always be empty) seems wrong.  Again, I think it should just be an array of strings that represent what event types are supported:
`listenerTypes: string[]`
Thoughts?

scytacki: The goal is to make this as easy as possible for interactives to use. This is a client library which interactives include to reduce the work they need to do to support the interactive api. 

If the interactives are going to be using https://github.com/concord-consortium/text-decorator/blob/master/src/text-decorator.ts to do the actual decoration. Then matching these types up with what that library needs seems best. 
However this seems hard since that library supports different options: HTML or React. What the interactive needs to pass to the library varies depending on what kind of content the interactive is decorating.

In any case it is possible to have the eventListeners as functions here since interactive-api-client is passing this info. So the interactive-api-client could create a listener function that way the interactive wouldn't have to do that itself.  I don't know if that really makes since. It would depend if that makes it easier or harder for interactives to use this.

Basically I'm suggesting that this listener be moved into the interactive-api-client: https://github.com/concord-consortium/lara/blob/035d13ca19693d1ecf2450814a5deba2dd6cf3aa/lara-typescript/src/example-interactive/src/components/runtime.tsx#L42-L54

But I realize that is probably not going to be helpful for an interactive that is using React components instead of html. I haven't yet looked at what your changes look like in question interactives.

mklewandowski: I haven't pushed anything to question-interactives yet, but I was basically adding what we see above: a listener function that gets created in question-interactives.  I think I need clarification on what you want me to do next.  You want me to keep the `eventListeners` in the `ITextDecorationInfo` type?  And then move the creation of the handler and the decoration into the interactive-api-client rather than having client handle the message and decide what to do with it?

mklewandowski: "If the interactives are going to be using https://github.com/concord-consortium/text-decorator/blob/master/src/text-decorator.ts to do the actual decoration. Then matching these types up with what that library needs seems best."
Which seems a little odd since the `ITextDecorationInfo ` type is being used to type a message from the AP that can't have a function, but the types used in the TextDecorator calls require functions.
https://github.com/concord-consortium/text-decorator/blob/master/src/text-decorator.ts#L96
So you want this?
```
interface IEventListener {
  type: string;
  listener: (evt: Event) => void;
}

type IEventListeners = IEventListener | IEventListener[];

export interface ITextDecorationInfo {
  words: string[];
  replace: string;
  wordClass: string;
  eventListeners: IEventListeners;
}
```

Which then would not work with the AP unless `listener` is optional in `IEventListener` (or the AP passes down a stub function?).

lbondaryk: I am not understanding @scytacki - it seems like we have a working implementation that we could, in theory, refine further but at what cost? Is it really necessary to change the style of the implementation at this point?

scytacki: I'm not talking about changing the style. Just moving something from one side of the abstraction line to the other. I'll make a quick branch off of this to demonstrate.

I think it would also be fine to go ahead with this as is. And we can try to refine it later. I just like to fix things when we can.  Once they get into production it is a lot hard to update them when they are at the intersection of several components.

scytacki: Here is what I was thinking: https://github.com/concord-consortium/lara/pull/666

mklewandowski: Quick follow up on types.  I did go ahead and type this:
```
export interface ITextDecorationInfo {
  words: string[];
  replace: string;
  wordClass: string;
  eventListeners: IEventListeners;
}
```
Which still feels wrong since `IEventListeners` is an `IEventListener` or an array of `IEventListener` objects which include `listener: (evt: Event) => void;` which is not what anyone can actually send down via post message since the function gets stripped out.  I still think this should be `listenerTypes: string[]` which would, for example, be something like `["click"]`.  Keeping it as an `IEventListeners` type seems very misleading.

mklewandowski:
@scytacki @kswenson there are another round of changes in the last handful of commits that do the following:
- use DecorateChildren` React component from the text decorator library instead of using the `decorateDOMClasses` function from the text decorator library
- add a `useGlossaryDecoration` custom hook which returns a `decorateOptions` object that gets passed to the DecorateChildren` component above and a CSS class that is used to decide which sections of the DOM get decorated.  It also adds listeners manually using `addEventListeners` from the text decorator library 
- sets up the click handler (and any other handlers) inside `addDecorateContentListener` inside `client.ts` - so per @scytacki's request this should remove the responsibility to set up the listeners from anyone who chooses to use the `useGlossaryDecoration` hook
- adds types to places where `any` was used.  I'm still not happy with the typing of the fields in `ITextDecorationInfo` - see note above

lara-typescript/src/interactive-api-client/types.ts
scytacki: Yeah the type going over the wire (postMessage) seems better if was listenerTypes. It might be best for it to be an array of objects incase we want to add more details to each listener in the future so the type would be `listenerTypes: Array<{type: string}>`. I didn't do that in my suggestion because I was trying to minimize the changes to the various parts. 

This does mean there are 2 types one for what goes over the wire(postMessage).  And another for what is provided to useDecorateContent. I think that is fine to have those two types. I think we have something like that is some other parts of the API.

mklewandowski: Yes, I think you're right that we would need two types for this.  I'll look at that today.

mklewandowski: @scytacki also, if you get a chance, take a look at the current state of this PR.  I think the last round of commits are in line with your thinking from last Friday.

scytacki: Yes your changes look good. I'm going to resolve this conversation. If we need more conversation about the types. Lets start a new one.

scytacki (approved):

scytacki:
In order more easily test the issue with decorations not working for multiple interactives in the AP. I suggest that we do a pre release of the interactive-api-client, so the AP PR can point at that. 

You bump the version number `npm run publish:interactive-api-client:beta`

scytacki:
If it makes anything easier I'm OK merging this PR at this point. But if leaving it in branch isn't holding things up, then I'd prefer to leave it un merged until we resolve the issue with the AP. My hunch is that something here in the interactive-api-client might have to change to fix that issue.

mklewandowski:
> If it makes anything easier I'm OK merging this PR at this point. But if leaving it in a branch isn't holding things up, then I'd prefer to leave it un merged until we resolve the issue with the AP. My hunch is that something here in the interactive-api-client might have to change to fix that issue.

I still want to put in the type adjustments that you mentioned (have them locally, just need to test with QI and AP). @kswenson and I just discussed the other issue (need to send text decoration message after interactive is done loading), and we think there is a solution (maybe not a long term one, but one that works in the short term) where we can use the supported features message as a handshake from the client that the init interactive message has been received and the content loaded.  That path would only require changing the AP and the QI.

mklewandowski:
> If it makes anything easier I'm OK merging this PR at this point. But if leaving it in a branch isn't holding things up, then I'd prefer to leave it un merged until we resolve the issue with the AP. My hunch is that something here in the interactive-api-client might have to change to fix that issue.

I updated the types per your earlier comment.  Changes are in last commit.

lara-typescript/src/interactive-api-client/types.ts
```diff
@@ -308,13 +308,20 @@ export interface IDecoratedContentEvent {
...
+  wordClass: string;
+  listenerTypes: Array<{type: string}>;
+}
+
+export interface ITextDecorationHandlerInfo {
   words: string[];
   replace: string;
   wordClass: string;
   eventListeners: IEventListeners;
 }
```
kswenson, line 322: You could DRY up the above:
```typescript
interface ITextDecorationBaseInfo {
  words: string[];
  replace: string;
  wordClass: string;
}

export interface ITextDecorationInfo extends ITextDecorationBaseInfo {
  listenerTypes: Array<{type: string}>;
}

export interface ITextDecorationHandlerInfo extends ITextDecorationBaseInfo {
  eventListeners: IEventListeners;
}
```
Also, I'm wondering if there might be more descriptive naming options -- maybe `TextDecorationStaticInfo` and `TextDecorationRuntimeInfo`? Just throwing some options out there to get the creative juices flowing.

kswenson (approved):
👍 Looks good -- just a couple of tweaks to consider

lara-typescript/src/interactive-api-client/types.ts
kswenson: Finally, I think you could still call the field `listeners` in both cases -- the interpretation would just be that it's the actual listener functions in one case and static descriptions of the listeners (currently including only the type but potentially extensible to other information down the road).

mklewandowski: I added the DRY suggestions.  Still not certain on names.  Let's think it over some more.

mklewandowski:
@scytacki @kswenson I'm fine waiting to merge this until after we all have time to test things a bit further.

scytacki:
I think this is good. I'm going to go ahead and bump the version and try to release it.

PR #665, opened 2020-12-17
Large interactive prompts caused issue when publishing to the portal

emcelroy:
https://www.pivotaltracker.com/story/show/176215104

[#176215104]

* Truncate and sanitize prompt content when it is used to replace empty name value on publication.
* Add related spec test.


scytacki (approved):
looks good to me

PR #666, opened 2020-12-19
sketch of moving the listener inside of the client

scytacki:
this hasn't been compiled or tested it is just to illustrate a change that
should make it easier for interactive developers to use the client
and support decorating their content.

This might work without changing the AP PR. I think that will already be sending
the `eventListeners` array.
I'm guessing what is coming through postMessage right now is something like
{eventListeners: [{type: "click", listener: null}]}
since JSON.stringify seems to turn functions into null.


scytacki:
This was incorporated into a different PR.

PR #667, opened 2020-12-28
Added page id to activity export [#176197155]

dougmartin:
Used by this Activity Player PR to implement deep linking to a page via id (https://github.com/concord-consortium/activity-player/pull/97).  The Activity Player PR will gracefully degrade if this PR isn't yet merged/deployed.

### DIFF
--- a/app/models/interactive_page.rb
+++ b/app/models/interactive_page.rb
@@ -280,7 +280,8 @@ class InteractivePage < ActiveRecord::Base
 
   def export
     helper = LaraSerializationHelper.new
-    page_json = self.as_json(only: [:name,
+    page_json = self.as_json(only: [:id,
+                                    :name,
                                     :position,
                                     :layout,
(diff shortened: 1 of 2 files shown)


scytacki (approved):
This seems OK to me. 

If you are also going to support the links in the portal-report to these pages. Then LARA should set the URL for pages in the activity structure it sends to the report-service to match the URL expected by the activity player. 

LARA activities now know if they are authored for the activity player or the LARA runtime, so the report-service activity structure export code can know if it should make a url like `https://authoring.concord.org/pages/123` or one like `https://activity-player.concord.org/?activity=https://authoring.concord.org/api/v1/activities/456&page=page_123`

That doesn't need to be part of this PR. I just wanted to point it out incase you were heading there next. I think that approach is better than having the portal-report know how to construct these activity-player page URLs.

PR #668, opened 2021-01-11
Fixed getFirebaseJWT post parameter [#174286329]

dougmartin:
Fixed dev time debug parameter that replaced the correct request parameter.


scytacki (approved):
I like the debugging choice of `lol`. :)

Looks good to me.

PR #669, opened 2021-01-15
Add export of approved script info [#176481121]

dougmartin:
This will be used by the Activity Player to determine the url to load for the plugin instead of using hardcoded urls.


scytacki (approved):
Looks good to me.

PR #670, opened 2021-01-19
Bundle Plugin API and Interactive API Client types into one file

pjanik:
This is necessary, as some types are shared between separate NPM packages (plugin-api and interactive-api-client), so they need to be bundled into one file, following module resolution. Default typings created by TS compiler are created per-file and retain original directory structure. Imports are not resolved. So, when plugin-api or interactive-api-client NPM packages import some files from the parent folder, these files were not getting published to NPM and package was half-broken. See: https://www.pivotaltracker.com/story/show/176518974

Why rollup when we already have webpack?
There are few tools to generate bundled .d.ts, but all of them had some problems:
  - Webpack plugins didn't seem production ready or had issues with TS lazy imports.
  - API Extractor and dts-bundle-generator also don't support TS lazy imports (https://github.com/microsoft/rushstack/issues/2140).

rollup-plugin-dts seemed like the only tool that handled correctly both plugin-api and interactive-api-client.


lara-typescript/package.json
```diff
@@ -68,6 +69,7 @@
   },
   "homepage": "https://github.com/concord-consortium/lara/tree/master/lara-typescript#readme",
   "devDependencies": {
+    "@microsoft/api-extractor": "^7.13.0",
```
kswenson, line 72: How is API Extractor being used at this point?

kswenson (approved):
👍 Kudos for tracking down this thorny issue and coming up with a viable solution. Microsoft's API Extractor is worth considering down the road, but this seems like the right solution for the time being.

pjanik:
Ah, api-extractor shouldn't be listed in package.json. I've forgot to delete it. Fixed in the last commit, thanks.
Right, it's interesting project, but [this bug](https://github.com/microsoft/rushstack/issues/2140) is a blocker at the moment (at least for the .d.ts rollup).

pjanik:
Since @kswenson approved this PR, I went ahead and released:
- interactive api client version 1.0.1
- plugin api version 3.2.1

And updated Question Interartives and Teacher Edition Plugin to use these new packages. Things were building fine and tests ran okay locally. Waiting for TravisCI to finish builds. 

If everything is okay, I'll merge this PR unless @scytacki has some feedback.

PR #674, opened 2021-01-20
Interactive API Client: fix empty string handling

pjanik:
Some interactives don't define any authored state. In this case, an empty string was passed to JSON.parse and crashing interactive runtime. That's why Dan didn't see user state in Gravity interactive.


scytacki (approved):
Looks good to me.

PR #675, opened 2021-01-25
Added page links to Activity Player export [#176197304]

dougmartin:


scytacki (approved):
Looks good to me.

If you wanted to make it a little better, the ap_url function could use URL parsing to add the activity and page params. That'd make the code a little easier to read and would allow us to add params inside of `ENV["ACTIVITY_PLAYER_URL"]`.  But the code wasn't doing that before, so I don't think this PR should be held up for that.

scytacki (approved):
Looks good to me

PR #676, opened 2021-01-28
#176708540 Modify publish_runs rake task

knowuh:
This task was updated so it only publishes runs that match specific remote_endpoints. It should speed things up.

If the value for `REPORT_PUSH_RUN_SELECT_REMOTE` is set in the environment, the task will find runs
that match that value.


scytacki (approved):
Looks good to me.

scytacki (approved):
looks good to me

PR #680, opened 2021-02-26
A onetime script for updating resource URLs in the portal

scytacki:


script/update_activity_player_urls.rb
```diff
@@ -0,0 +1,12 @@
+# This is intended to be run in the Portal.
+# It was needed to handle a change in the URL format published by LARA version 1.60.0 
+tool = Tool.where(name: "ActivityPlayer").first
+ExternalActivity.where(tool: tool).find_each{ |ea|
+  resource_url = ea.url
+  uri = URI.parse(resource_url)
+  query = Rack::Utils.parse_query(uri.query)
+  uri.query = Rack::Utils.build_query(query)
```
dougmartin, line 8: So it looks like this script just rewrites the existing query parameter and doesn't add/remove anything.  Is this just to fix escaping or the query string?

dougmartin (approved):
It sounds like from the comment that you already ran this - I was just curious in the comment I added as to why it was needed.

script/update_activity_player_urls.rb
dougmartin: Nevermind - I saw the comment in the other PR adding preview links as to why the query string rewrite was needed.

scytacki: Nope, this is independent from that new PR. 
Here is the story: https://www.pivotaltracker.com/story/show/177080964

There was a change in LARA recently to use build_query to make the AP URL. 
This resulted in the activity or sequence parameter being escaped. (`URI.escape` doesn't do parameter escaping it is the build_query that does it)
When lara re-publishes resources to the portal, the portal uses a basic string match on the URL to find the associated external_activity. So this escaping was breaking the republishing of resources that were published before the change to escaping.

Here is when it was switched to start escaping:
https://github.com/concord-consortium/lara/commit/26fafcc61144e55656847ab2eaa6030af81166d6
Again, note that `URI.escape` was basically doing nothing in this commit, it is the build_query that was actually escaping the parameter.

PR #681, opened 2021-02-26
this escape wasn't really doing anything

scytacki:
build_query automatically escapes the values in the hash
URI.escape only escapes characters that aren't valid in a URI (as opposed to characters to avoid in a parameter)
so for example it will not escape `=`, `&`,  `?`, or `/`


scytacki:
this is now part of another PR.

PR #682, opened 2021-02-28
add preview_urls to the report-structure

scytacki:
this also returns the urls of pages back to being LARA urls
the portal-report will need to be updated to check for this preview_url and use that
if it exist.
without this change the portal-report page preview links will be broken

Additionally this adds a new route and action so it is easier to see what an activity or sequence will send to the report-service. For example: https://app.lara.docker/api/v1/activities/10/report_structure

It also removes `URI.escape` which probably wasn't doing what was expected.
build_query automatically escapes the values in the hash
URI.escape only escapes characters that aren't valid in a URI (as opposed to characters to avoid in a parameter)
so for example it would not escape =, &, ?, or /

### DIFF
--- a/app/controllers/api/v1/lightweight_activities_controller.rb
+++ b/app/controllers/api/v1/lightweight_activities_controller.rb
@@ -12,4 +12,12 @@ class Api::V1::LightweightActivitiesController < API::APIController
     lightweight_activity_json = activity.export.to_json
     render json: lightweight_activity_json
   end
+
+  # GET /api/v1/activities/1/reeport_structure.json
+  def report_structure
+    activity = LightweightActivity.find(params[:id])
+    self_url = "#{request.protocol}#{request.host_with_port}"
+    json = activity.serialize_for_report_service(self_url).to_json
+    render json: json
+  end
 end
--- a/app/controllers/api/v1/sequences_controller.rb
+++ b/app/controllers/api/v1/sequences_controller.rb
@@ -12,4 +12,12 @@ class Api::V1::SequencesController < API::APIController
     sequence_json = sequence.export
     render json: sequence_json
   end
+
+  # GET /api/v1/sequences/1/reeport_structure.json
+  def report_structure
+    sequence = Sequence.find(params[:id])
+    self_url = "#{request.protocol}#{request.host_with_port}"
+    json = sequence.serialize_for_report_service(self_url).to_json
+    render json: json
+  end
 end
--- a/app/models/sequence.rb
+++ b/app/models/sequence.rb
@@ -170,25 +170,28 @@ class Sequence < ActiveRecord::Base
       url: local_url
     }
     if self.runtime == "Activity Player"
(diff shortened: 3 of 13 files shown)


dougmartin (approved):
Looks good :+1:

app/controllers/api/v1/lightweight_activities_controller.rb
```diff
@@ -12,4 +12,12 @@ def show
     lightweight_activity_json = activity.export.to_json
     render json: lightweight_activity_json
   end
+
+  # GET /api/v1/activities/1/reeport_structure.json
```
kswenson, line 16:
```
  # GET /api/v1/activities/1/report_structure.json
```

app/controllers/api/v1/sequences_controller.rb
```diff
@@ -12,4 +12,12 @@ def show
     sequence_json = sequence.export
     render json: sequence_json
   end
+
+  # GET /api/v1/sequences/1/reeport_structure.json
```
kswenson, line 16:
```
  # GET /api/v1/sequences/1/report_structure.json
```

app/models/sequence.rb
```diff
@@ -169,18 +169,29 @@ def serialize_for_report_service(host)
       name: self.title,
       url: local_url
     }
+    if self.runtime == "Activity Player"
+      data[:preview_url] = self.activity_player_sequence_url(host, preview: true)
+    end
+
+    # FIXME: the urls and preview_urls in these activities do not include the
+    # the sequence, so links in the report to the activities or activity pages
```
kswenson, line 177:
```
    # sequence, so links in the report to the activities or activity pages
```

spec/controllers/api/v1/lighweight_activity_controller_spec.rb
```diff
@@ -25,6 +25,25 @@
...
+    it 'recognizes and generates #report_structure' do
+      expect({:get => "api/v1/activities/1/report_structure.json"}).to route_to(
+        :controller => 'api/v1/lightweight_activities',
+        :action => 'report_structure',
+        :id => "1",
+        :format => "json"
+      )
+    end
+
+    it "when user is anonymous, shows an the json sent to the report structure" do
```
kswenson, line 38:
```
    it "when user is anonymous, shows the json sent to the report structure" do
```

spec/controllers/api/v1/sequences_controller_spec.rb
```diff
@@ -7,22 +7,41 @@
...
+    it 'recognizes and generates #report_structure' do
+      expect({:get => "api/v1/sequences/1/report_structure.json"}).to route_to(
+        :controller => 'api/v1/sequences',
+        :action => 'report_structure',
+        :id => "1",
+        :format => "json"
+      )
+    end
+
+    it "when user is anonymous, shows an the json sent to the report structure" do
```
kswenson, line 37:
```
    it "when user is anonymous, shows the json sent to the report structure" do
```

PR #688, opened 2021-03-19
Label for link target options

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/177211486

[#177211486]

These changes fix a bug where the Target drop-down menu options and Cancel button in the rich text editor's Insert Link modal were essentially invisible.

Currently the Insert Link modal looks like this on production:
image

These changes will make it look like this:
image


knowuh (approved):
Looks good. Thanks.

PR #691, opened 2021-03-22
Fix teacher edition window shades in header taking up space when hidden

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/175025869

[#175025869]


knowuh (approved):
Looks good, thank you.

PR #697, opened 2021-04-14
Fix convert links for sequences

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/177747131

[#177747131]

These changes ensure that the convert link URLs for sequences use a sequence URL for the `lara_resource` parameter instead of an activity URL.


scytacki (approved):
looks good to me.

PR #699, opened 2021-04-23
Fix completion page not set on import

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/177895067

[#177895067]


scytacki (approved):
Seems good to me.

PR #700, opened 2021-04-28
Fix question wrapper can't be added to non library interactive question

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/177952201

[#177952201]


scytacki (approved):
Seems fine to me.

PR #702, opened 2021-05-05
Reduce duplicated library interactives

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/177968044

[#177968044]

### DIFF
--- a/app/models/library_interactive.rb
+++ b/app/models/library_interactive.rb
@@ -82,7 +82,7 @@ class LibraryInteractive < ActiveRecord::Base
 
   def generate_export_hash
     export = export()
-    export.except("authoring_guidance", "description", "name")
+    export.except!("authoring_guidance", "description", "name")
     Digest::SHA1.hexdigest(export.to_json)
   end
 
--- a/app/models/library_interactive.rb
+++ b/app/models/library_interactive.rb
@@ -90,4 +90,8 @@ class LibraryInteractive < ActiveRecord::Base
     self.export_hash = self.generate_export_hash()
   end
 
+  def use_count
+    ManagedInteractive.where(library_interactive_id: self.id).length
+  end
+
 end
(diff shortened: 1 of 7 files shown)


app/models/library_interactive.rb
```diff
@@ -81,11 +81,17 @@ def export
   end
 
   def generate_export_hash
-    Digest::SHA1.hexdigest(export().to_json)
+    export = export()
+    export.except("authoring_guidance", "description", "name")
```
scytacki, line 85: this won't actually remove these properties from export it just returns a new value that has them removed if you use the bang version instead: `except!(...)` that will remove them from the export variable.

scytacki (changes requested):
I'd recommend adding some spec tests of the new generate_export_hash function. They ought to expose the error in the code I found.

For example you could make a test that builds two library interactives where the only difference is their description. The value of generate_export_hash should match. And if the two library interactives have different base_urls then the generate_export_hash should not match.

app/models/library_interactive.rb
scytacki: For reference I tested this in the rails console with:
```
irb(main):001:0> example = {hello: "hi", bye: "bye"}
=> {:hello=>"hi", :bye=>"bye"}
irb(main):003:0> examplej = example.as_json
=> {"hello"=>"hi", "bye"=>"bye"}
irb(main):004:0> examplej.except("hello")
=> {"bye"=>"bye"}
irb(main):005:0> examplej
=> {"hello"=>"hi", "bye"=>"bye"}
irb(main):006:0> examplej.except!("hello")
=> {"bye"=>"bye"}
irb(main):007:0> examplej
=> {"bye"=>"bye"}
```

emcelroy:
@scytacki Thanks. Are the tests that I've added enough, do you think?

scytacki (approved):
Looks good to me.

scytacki:
@emcelroy can you add the next LARA release version label to the associated story.

emcelroy:
@scytacki Sure. I've added the latest version label.

PR #710, opened 2021-05-10
Allow admins to migrate interactives

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/178060576

[#178060576]

These changes provide an option to migrate library interactives used by questions in existing activities from one version to another.

### DIFF
--- a/spec/controllers/library_interactives_controller_spec.rb
+++ b/spec/controllers/library_interactives_controller_spec.rb
@@ -111,6 +111,23 @@ describe LibraryInteractivesController do
       end
     end
 
+    describe "GET #migrate" do
+      it "returns a success response" do
+        library_interactive = LibraryInteractive.create! valid_attributes
+        get :migrate, {:id => library_interactive.to_param, :library_interactive => valid_attributes}
+        expect(response).to be_successful
+      end
+    end
+
+    describe "PUT #migrate" do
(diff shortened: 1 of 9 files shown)


spec/controllers/library_interactives_controller_spec.rb
```diff
@@ -111,6 +111,23 @@
...
+      end
+    end
+
+    describe "PUT #migrate" do
+      it "changes all references to one library_interactive to another specified library_interactive and redirects to the library_interactive index" do
+        library_interactive1 = LibraryInteractive.create! valid_attributes
+        library_interactive2 = LibraryInteractive.create! valid_attributes
+        put :update, {:id => library_interactive1.to_param, :new_library_interactive_id => library_interactive2.to_param}
+        expect(response).to redirect_to(library_interactives_url)
+      end
```
scytacki, line 128: I think this test is not actually covering the migration function. Code climate is reporting it as not covered. My guess is that the `put :update` needs to be changed to `put :migrate`

scytacki (changes requested):
I think one of the tests is covering the wrong action.

spec/controllers/library_interactives_controller_spec.rb
emcelroy: Right. Thanks for catching that. It's fixed.

scytacki (approved):
Looks great!

PR #711, opened 2021-05-10
Reset preview drop-down menu after selection made

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/178060414

[#178060414]


scytacki (approved):
looks good to me.

PR #712, opened 2021-05-19
Add ID to activity export method.

emcelroy:
Related PT Story: https://www.pivotaltracker.com/story/show/178131977

[#178131977]

Adding the activity ID to activity exports will make it possible for the Activity Player to identify an activity within a sequence when the activity ID is included as a param in the Activity Player sequence's URL.


scytacki (approved):
looks good. I looked to see if this code is already tested but didn't find existing tests, so I'm OK not including them in this PR.

PR #713, opened 2021-05-25
Fix TE interactive wrappers displaying default text for correct tab.

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/178276526

[#178276526]

Adds a migration to remove placeholder text from Teacher Edition question wrappers.


scytacki (approved):
looks good to me

PR #714, opened 2021-05-28
Have migration skip plugins with invalid JSON values for author_data.

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/178276526

[#178276526]


scytacki (approved):
looks good

PR #715, opened 2021-05-28
Adding a new rake task to create a lara admin user in dev environment only

tejal-shah:
This will help in importing activities into LARA when running the integration suite of portal-lara tests.


scytacki (approved):
Looks good to me.

PR #717, opened 2021-06-01
extra step needed to fix lara and portal communication (making https)

sri-tscube:
site url should be https otherwise portal and lara cannot communicate while setting up dev environment


knowuh (approved):
Thank you @sri-tscube

tejal-shah (approved):

scytacki (approved):
Thanks!

PR #718, opened 2021-06-01
Adding a / at the end of the Activity Player URL so that it doesn't f…

tejal-shah:
…ail when parameters are added to it


scytacki (approved):
looks good

PR #719, opened 2021-06-04
Fix activities' project-based customizations not working correctly

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/178201284

[#178201284]

Corresponding Activity Player changes can be found in [this PR](https://github.com/concord-consortium/activity-player/pull/254).


spec/factories/projects.rb
```diff
@@ -3,7 +3,9 @@
 FactoryGirl.define do
   factory :project do
     title "MyString"
-    logo "MyString"
+    project_key "MyString"
```
scytacki, line 6: This should probably use a generator so it makes a unique string each time it is called. 
Here is an example:
https://github.com/concord-consortium/lara/blob/526c2f0e382b7561ca883cba8672d0c3066882ce/spec/factories/plugins.rb#L4-L10

The sequence call creates a generator named `:word` that then returns a string. It takes the `n` param. So in this case you could call it `:project_key` and return something like "test-key-#{n}". Then you use it in the factory with `project_key { generate(:project_key) }`

spec/models/project_spec.rb
```diff
@@ -24,4 +24,37 @@
...
+        url: "https://concord.org/projects/test-project-1/"
+      }
+      project = Project.create(project_params)
+      expect(project.export.to_json).to eq(project_params.to_json)
+    end
+  end
+
+  describe "Project.find_or_create" do
+    project_data = {project_key: "default-project", title: "Default Project"}
+    it "finds a project when it exists" do
```
scytacki, line 47: Based on the test coverage report. It seems like this test isn't actually finding a project, but it is instead creating a new one.

scytacki (changes requested):
Overall looks great. I found a few things to look at.

app/models/project.rb
```diff
@@ -51,4 +52,34 @@ def image_file_exists?(filename)
...
+                                help: project_data[:help],
+                                logo_ap: project_data[:logo_ap],
+                                logo_lara: project_data[:logo_lara],
+                                project_key: project_data[:project_key],
+                                title: project_data[:title],
+                                url: project_data[:url])
+      new_project.save
+      return new_project
+    else
+      return existing_project
```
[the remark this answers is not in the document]
scytacki: looks like it is working now.

spec/models/project_spec.rb
scytacki: looks like this works now

spec/factories/projects.rb
scytacki: Fix looks good.

app/models/project.rb
```diff
@@ -1,14 +1,15 @@
...
+  attr_accessible :footer, :logo_lara, :logo_ap, :title, :url, :theme_id, :about, :help, :project_key
+  validates :project_key, uniqueness: true
   has_many :sequences
   has_many :lightweight_activities
   belongs_to :theme
 
   protected
   def self.create_default
-    self.create(:title => DefaultName, :logo => '', :url => 'https://concord.org/')
+    self.create(:title => DefaultName, :logo_lara => '', :url => 'https://concord.org/', :project_key => 'default-project')
```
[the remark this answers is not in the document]
scytacki: Fix looks good.

scytacki (approved):
Looks good.

PR #721, opened 2021-06-10
Update aws-ses gem so site can use SES Signature V4.

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/177879684

[#177879684]


scytacki (approved):
looks good

PR #724, opened 2021-07-02
Added getFirebaseJwt to authoring mode in interactive api client [#178284894]

dougmartin:


lara-typescript/src/page-item-authoring/common/components/interactive-authoring.tsx
```diff
@@ -48,6 +48,17 @@ export const InteractiveAuthoring: React.FC<Props> = (props) => {
     version: 1,
     error: null,
     mode: "authoring",
+    hostFeatures: {
+      modal: {
+        version: "1.0.0",
+        dialog: false,
+        lightbox: true,
+        alert: true
+      },
```
scytacki, line 57: I don't see new code supporting modals during authoring, was this supported before and just not added to the hostFeatures?

scytacki (changes requested):
This looks good overall. 
The only problem I found was that it implies that modals are supported in authoring, but I don't think they are.
I'll try to track this down myself, since Doug is on vacation now.

lara-typescript/src/page-item-authoring/common/components/interactive-authoring.tsx
scytacki: I removed this, because I didn't see any modal support during authoring.

scytacki:
Besides the changes to the tests, I bumped the version to `1.0.4` since @kswenson got in a PR before this one.

I also made 2 other changes:
- removed the check for the mode and now the client just looks at the hostFeatures to see if the host supports JWTs
- removed the code in this PR before which was saying the modal api was supported by the authoring interactive host

lara-typescript/src/interactive-api-client/api.ts
```diff
@@ -218,7 +218,8 @@ export const getAuthInfo = (): Promise<IAuthInfo> => {
 export const getFirebaseJwt = (firebaseApp: string): Promise<IJwtResponse> => {
   return new Promise<IJwtResponse>((resolve, reject) => {
     getInitInteractiveMessage().then(initMsg => {
-      if (initMsg && initMsg.mode === "runtime" && initMsg.hostFeatures?.getFirebaseJwt?.version === "1.0.0") {
+      if (initMsg
+          && initMsg.hostFeatures?.getFirebaseJwt?.version === "1.0.0") {
```
kswenson, line 222: Can be simplified to:
```suggestion
      if (initMsg?.hostFeatures?.getFirebaseJwt?.version === "1.0.0") {
```

kswenson (commented):
👍 Looks good

kswenson (approved):
👍 Looks good

PR #726, opened 2021-07-15
Attachments API

kswenson:
[[#178526615]](https://www.pivotaltracker.com/story/show/178526615)

Proposed Client API:
```
writeAttachment: (request: IWriteAttachmentRequest) => Promise<Response>;
readAttachment: (name: string) => Promise<Response>;  // fetch Response object
getAttachmentUrl: (name: string) => Promise<string>;  // url for use in <img> tag, for instance
```

Proposed PostMessage API:
```
postMessage("getAttachmentUrl", attachmentUrlRequest);
=> postMessage("attachmentUrl", attachmentUrlResponse);
```

Also did a round of updating dependencies and updated the version number in preparation for an upcoming 1.1 release.


lara-typescript/src/interactive-api-client/api.spec.ts
```diff
@@ -89,6 +90,23 @@ describe("api", () => {
...
+      words: [],
+      replace: "",
+      wordClass: "word"
+    };
+    mockedPhone.fakeServerMessage({type: "decorateContent", content});
+    api.postDecoratedContentEvent({ type: "type", text: "text" });
+    expect(callback).toHaveBeenCalledTimes(1);
+    api.removeDecorateContentListener();
+    api.postDecoratedContentEvent({ type: "type", text: "text" });
+    expect(callback).toHaveBeenCalledTimes(1);
```
scytacki, line 107: This seems to be documenting broken behavior. If the listener is removed wouldn't the callback be expected to not be called?
I think it'd be fine to keep this, but adding a comment explaining it would be helpful.

lara-typescript/src/interactive-api-client/api.spec.ts
```diff
@@ -373,6 +391,111 @@ describe("api", () => {
...
+      const url = "https://concord.org/foo";
+      const apiResponse = { url };
+      const fetchResponse = { ok: true, text: () => "foo" };
+      fetchMock.mockReturnValue(Promise.resolve(fetchResponse));
+      await testReadAttachmentResponse({
+        requestContent: [{ name: "name.ext" }],
+        responseContent: [apiResponse],
+        resolvesTo: [fetchResponse]
+      });
+      expect(fetchMock.mock.calls[0][0]).toBe(url);
```
scytacki, line 471: Maybe you want to verify the method was a GET here.

lara-typescript/src/interactive-api-client/api.ts
```diff
@@ -388,3 +391,67 @@ export const getInteractiveSnapshot = (options: IGetInteractiveSnapshotOptions)
 export const getLibraryInteractiveList = (options: IGetLibraryInteractiveListRequest) => {
   THROW_NOT_IMPLEMENTED_YET("getLibraryInteractiveList");
 };
+
+export const writeAttachment = (params: Omit<IWriteAttachmentRequest, "requestId" | "operation">) => {
+  return new Promise<void>((resolve, reject) => {
+    const client = getClient();
+    const { name, content, options } = params;
+    const request: IAttachmentUrlRequest = { name, operation: "write", requestId: client.getNextRequestId() };
+    client.addListener("attachmentUrl", async (response: IAttachmentUrlResponse) => {
```
scytacki, line 400: I think this needs to be only listening for responses that have the matching requestId. I can't remember if there is some existing addListener mechanism to make this easier.

The reason for this is that the interactive might add multiple attachments and the responses could come out of order.

lara-typescript/src/interactive-api-client/api.ts
```diff
@@ -388,3 +391,67 @@ export const getInteractiveSnapshot = (options: IGetInteractiveSnapshotOptions)
...
+    client.post("getAttachmentUrl", request);
+  });
+};
+
+export const readAttachment = (name: string) => {
+  return new Promise<Response>((resolve, reject) => {
+    // set up response listener
+    const client = getClient();
+    const request: IAttachmentUrlRequest = { name, operation: "read", requestId: client.getNextRequestId() };
+    client.addListener("attachmentUrl", async (response: IAttachmentUrlResponse) => {
```
scytacki, line 423: Same issue here about using the requestId.

lara-typescript/src/interactive-api-client/api.ts
```diff
@@ -388,3 +391,67 @@ export const getInteractiveSnapshot = (options: IGetInteractiveSnapshotOptions)
...
+    client.post("getAttachmentUrl", request);
+  });
+};
+
+export const getAttachmentUrl = (name: string) => {
+  return new Promise<string>((resolve, reject) => {
+    // set up response listener
+    const client = getClient();
+    const request: IAttachmentUrlRequest = { name, operation: "read", requestId: client.getNextRequestId() };
+    client.addListener("attachmentUrl", async (response: IAttachmentUrlResponse) => {
```
scytacki, line 447: And here too.

lara-typescript/src/interactive-api-client/api.ts
```diff
@@ -388,3 +391,67 @@ export const getInteractiveSnapshot = (options: IGetInteractiveSnapshotOptions)
...
+      }
+      else {
+        reject(new Error(response.error || "error writing attachment"));
+      }
+    }, request.requestId);
+    client.post("getAttachmentUrl", request);
+  });
+};
+
+export const readAttachment = (name: string) => {
```
scytacki, line 418: It seems good to explicitly define the return type here, since if it is changes code using this client will break.

lara-typescript/src/interactive-api-client/api.ts
```diff
@@ -388,3 +391,67 @@ export const getInteractiveSnapshot = (options: IGetInteractiveSnapshotOptions)
...
+export const writeAttachment = (params: Omit<IWriteAttachmentRequest, "requestId" | "operation">) => {
+  return new Promise<void>((resolve, reject) => {
+    const client = getClient();
+    const { name, content, options } = params;
+    const request: IAttachmentUrlRequest = { name, operation: "write", requestId: client.getNextRequestId() };
+    client.addListener("attachmentUrl", async (response: IAttachmentUrlResponse) => {
+      if (response.url) {
+        try {
+          await fetch(response.url, { ...options, method: "PUT", body: content });
+          resolve();
```
scytacki, line 404: Perhaps this can return the result of the put? That way if there is some info in result clients would have access to it. Similar to how it is returned in readAttachment.

lara-typescript/src/interactive-api-client/api.ts
```diff
@@ -388,3 +391,67 @@ export const getInteractiveSnapshot = (options: IGetInteractiveSnapshotOptions)
...
+      }
+      else {
+        reject(new Error(response.error || "error reading attachment"));
+      }
+    }, request.requestId);
+    client.post("getAttachmentUrl", request);
+  });
+};
+
+export const getAttachmentUrl = (name: string) => {
```
scytacki, line 442: Because these URLs will expire. It would be useful if an optional timeout (in seconds) was passed as an argument.  That way an interactive that wants a URL that it can reuse during a session could specify a few hours. While URL that is sensitive and going to be used immediately could have a short timeout of a few seconds.

lara-typescript/src/interactive-api-client/types.ts
```diff
@@ -349,6 +351,19 @@ export interface IAuthInfo {
   email?: string;
 }
 
+export interface IAttachmentUrlRequest extends IBaseRequestResponse {
+  name: string;
+  operation: "read" | "write";
```
scytacki, line 356: As mentioned above, it would be good to add a `timeout` or  `expires` field here. I'm not sure it is needed for writes because in those cases the client should write the content quickly and this could just be hardcoded to a few minutes. But for reads it is useful for the interactive to have some flexibility.

scytacki (changes requested):
Overall looks good. I found a few things that should probably be changed, but I don't feel strongly about them.

lara-typescript/src/interactive-api-client/api.spec.ts
kswenson: We were already called once before the removal, so the test is verifying that we weren't called again after the removal. I added a comment to clarify.

lara-typescript/src/interactive-api-client/api.spec.ts
kswenson: We're not passing a `method` explicitly here, but I've added a test that `fetch` was called with exactly one parameter, i.e. only a url and no additional options.

lara-typescript/src/interactive-api-client/api.ts
kswenson: The `requestId` is passed after the callback in line 413.

lara-typescript/src/interactive-api-client/api.ts
kswenson: The `requestId` is passed after the callback in line 437.

lara-typescript/src/interactive-api-client/api.ts
kswenson: The `requestId` is passed after the callback in line 454.

lara-typescript/src/interactive-api-client/api.ts
kswenson: I was thinking that that should be a host decision (in the activity-player draft PR it defaults to five minutes for writes and two hours for reads) but I'm happy to make it client-configurable if you think that makes sense.

scytacki (approved):
Looks good.

pjanik:
I know it's already merged, but I've tried to get familiar with these changes and they look good to me 👍  Some parameters aren't 100% clear (for example a reason for `expires` in `getAttachementUrl`), but I think I'll understand it better when I see the final implementation in LARA / AP.

PR #727, opened 2021-07-26
179007673 add labbook runtime metadata

knowuh:
This simply adds metadata for Labbook Question runtime and authoring types, specifically the `question_type` and `answer_type` fields.

What I saw in the Question Interactives repo is that the implementation details of the interactives are kept in that repo, and extend the metadata types. So I followed that same pattern. 

Please let me know if I am supposed to do anything more than that here.

### DIFF
--- a/docs/interactive-api-client/globals.md
+++ b/docs/interactive-api-client/globals.md
@@ -17,6 +17,7 @@
 * [IAuthoringImageQuestionMetadata](interfaces/iauthoringimagequestionmetadata.md)
 * [IAuthoringInitInteractive](interfaces/iauthoringinitinteractive.md)
 * [IAuthoringInteractiveMetadata](interfaces/iauthoringinteractivemetadata.md)
+* [IAuthoringLabbookMetadata](interfaces/iauthoringlabbookmetadata.md)
 * [IAuthoringMetadataBase](interfaces/iauthoringmetadatabase.md)
 * [IAuthoringMultipleChoiceChoiceMetadata](interfaces/iauthoringmultiplechoicechoicemetadata.md)
 * [IAuthoringMultipleChoiceMetadata](interfaces/iauthoringmultiplechoicemetadata.md)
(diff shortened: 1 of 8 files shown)


scytacki:
I commented on this in Slack. I don't think this change is necessary. I'm commenting here for completeness, but if there is more discussion to be had lets do it in Slack and summarize here:

> I don’t think you need to add any new types to lara-interactive-api. Those types are really just so the portal-report and researcher reports can provide built in UI or columns for the interactive.  I think in this case the labbook can be just a generic interactive that doesn’t require any built in UI in portal-report codebase.

> Other examples like this that don’t have types (as far as I know) in the lara-interactive-api are the Drawing Tool, Carousel and Scaffolded Question interactives in the question interactives repo.

I'll leave the PR open in case there is something I'm missing.

pjanik (commented):
Generally, I can confirm everything that @scytacki said. Re:
>It’s not 100% clear when to use a custom answer-type vs.  interactive_state.

I think this is mostly related to reporting. Note that in the old report we had custom reports for a few questions types like open response, multiple-choice, and image question. These types are quite specific and our report can provide a much nicer experience if it knows that a given question implements one of these interfaces. Otherwise, it’ll render iframe and the question itself needs to provide a report / read-only view. This is generic and will always work, but it can be slower and overkill for basic questions types. For example, it makes no sense to render an iframe to display text only answer. That’s why we have these metadata interfaces.

Note that any new interactive can pretend to be multiple choice / open response / or image question and use this special type of reporting.

Actually, I don’t know what’s the plan for Labook reporting. But there are a few questions that you can answer to decide which types should be used.

1. Are we going to have custom, non-iframe-based Labbook reporting in Portal Report? Do we need that to have some statistics, aggregate view, custom UI? If yes, then we would need new metadata types that you just added. Otherwise, we don’t need them.
2. Does it make sense to reuse one of the existing meta types while developing Labbook? Given that it’s based on images and drawing tool, I could imagine that “image_question” type could be a good candidate. In this case, you don’t have to add anything to LARA, but just extend image question metadata types in question-interactives repo. But at the same time, Labbook is built from multiple images, so probably it’s not enough to show only one image anyway.
3. If answer to 1 and 2 is “no”, then you can simply extend the basic / generic metadata types. Labbook will have to provide a custom report view. My guess is that’s what we want. Most of the questions in question-interactives repo reuse `Runtime` component for that by blocking any interactions / making it read-only. It can work here too or not, depending on the reporting ideas. We can always use a custom Report component too.

It’s a bit long explanation but hopefully will make it clearer. :wink:

scytacki:
based on slack conversation I'm going to close this.

PR #728, opened 2021-07-27
Background Images AP

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/178990327

[#178990327]

These changes add a `background_image` property to activities and sequences. The value is only editable for resources set to use the Activity Player runtime environment. The value is included in resources' JSON for use by Activity Player in setting a background image for the resources' pages. 

Related Activity Player changes are at https://github.com/concord-consortium/activity-player/pull/260

### DIFF
--- a/app/models/lightweight_activity.rb
+++ b/app/models/lightweight_activity.rb
@@ -138,10 +138,10 @@ class LightweightActivity < ActiveRecord::Base
                                         :editor_mode,
                                         :student_report_enabled,
                                         :show_submit_button,
-                                        :runtime ])
+                                        :runtime,
+                                        :background_image ])
     activity_json[:version] = 1
     activity_json[:theme_name] = self.theme ? self.theme.name : nil
-    activity_json[:background_image] = self.background_image ? self.background_image : nil
     activity_json[:project] = self.project ? self.project.export : nil
     activity_json[:pages] = []
     self.pages.each do |p|
--- a/app/models/lightweight_activity.rb
+++ b/app/models/lightweight_activity.rb
@@ -169,7 +169,8 @@ class LightweightActivity < ActiveRecord::Base
       time_to_complete: activity_json_object[:time_to_complete],
       layout: activity_json_object[:layout],
       editor_mode: activity_json_object[:editor_mode],
-      runtime: activity_json_object[:runtime]
+      runtime: activity_json_object[:runtime],
+      background_image: activity_json_object[:background_image]
     }
 
   end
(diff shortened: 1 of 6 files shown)


app/models/lightweight_activity.rb
```diff
@@ -138,7 +138,8 @@ def export
                                         :editor_mode,
                                         :student_report_enabled,
                                         :show_submit_button,
-                                        :runtime ])
+                                        :runtime,
+                                        :background_image ])
```
scytacki, line 142: It looks like this background_image property is not imported correctly. I'm not sure but I think the correct way to do that is to add it to the extract_from_hash method.

app/models/sequence.rb
```diff
@@ -99,7 +99,8 @@ def export
                                         :logo,
                                         :display_title,
                                         :thumbnail_url,
-                                        :runtime
+                                        :runtime,
+                                        :background_image
```
scytacki, line 103: I think this has the same issue of not being imported correctly.

app/views/lightweight_activities/edit.html.haml
```diff
@@ -26,6 +26,10 @@
       .field
         = f.label :theme_id, 'Theme'
         = f.select :theme_id, options_from_collection_for_select(Theme.all, 'id', 'name', @activity.theme_id), { :include_blank => "None (inherit from sequence, project, or site default)" }
```
scytacki, line 28: It would be best to disable the theme setting if the runtime is AP. As far as I know the theme will have no effect in the AP, so disabling this will make it more clear.

app/views/sequences/_form.html.haml
```diff
@@ -32,6 +32,10 @@
     .field
       = f.label :theme_id, "Theme"
       = f.select :theme_id, options_from_collection_for_select(Theme.all, 'id', 'name', @sequence.theme_id), { :include_blank => "None (inherit from project, or site default)" }
```
scytacki, line 34: Same approach of disabling the theme for sequences when the AP runtime is selected seems best.

scytacki (changes requested):
Overall looks good.

I think I found an issue with importing activities and sequences.
I also think the theme field should be disabled to keep the confusion down.

app/models/sequence.rb
emcelroy: @scytacki Thanks for catching that. I've fixed it.

app/views/sequences/_form.html.haml
emcelroy: @scytacki I made the disable Theme changes as well.

scytacki (approved):
Looks good to me.

PR #729, opened 2021-07-30
Add content type argument to getAttachmentUrl()

kswenson:


lara-typescript/src/interactive-api-client/types.ts
```diff
@@ -359,6 +359,7 @@ export interface IAuthInfo {
 export interface IAttachmentUrlRequest extends IBaseRequestResponse {
   name: string;
   operation: "read" | "write";
+  type?: string;    // defaults to application/json
```
scytacki, line 362: I think this should be called `contentType` just to help devs know what this better from the name.

scytacki (changes requested):
Looks good. I just think the `type` should be changed to `contentType`

kswenson:
I'm going to go ahead and merge this as the issue @scytacki raised on review has been addressed.

PR #731, opened 2021-08-03
Final(?) attachments tweak: always set Content-Type when writing attachments

kswenson:
…rather than only when client passes a `contentType`

### DIFF
--- a/lara-typescript/src/interactive-api-client/api.ts
+++ b/lara-typescript/src/interactive-api-client/api.ts
@@ -396,12 +396,13 @@ type WriteAttachmentParams = Omit<IWriteAttachmentRequest, "requestId" | "operat
 export const writeAttachment = (params: WriteAttachmentParams): Promise<Response> => {
   return new Promise<Response>((resolve, reject) => {
     const client = getClient();
-    const { content, contentType, options = {}, ...others } = params;
+    const { content, options = {}, ...others } = params;
+    const { contentType = "text/plain" } = params;
     const request: IAttachmentUrlRequest = { ...others, operation: "write", requestId: client.getNextRequestId() };
     client.addListener("attachmentUrl", async (response: IAttachmentUrlResponse) => {
       if (response.url) {
         const headers: Record<string, string> = { ...(options.headers as Record<string, string>) };
-        headers["Content-Type"] = contentType || "text/plain";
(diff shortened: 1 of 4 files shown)


lara-typescript/src/interactive-api-client/api.ts
```diff
@@ -396,14 +396,13 @@ type WriteAttachmentParams = Omit<IWriteAttachmentRequest, "requestId" | "operat
 export const writeAttachment = (params: WriteAttachmentParams): Promise<Response> => {
   return new Promise<Response>((resolve, reject) => {
     const client = getClient();
-    const { content, contentType, options, ...others } = params;
+    const { content, contentType, options = {}, ...others } = params;
     const request: IAttachmentUrlRequest = { ...others, operation: "write", requestId: client.getNextRequestId() };
```
scytacki, line 400: Shouldn't the contentType be passed in the request object?
It is needed to create the presigned URL.

lara-typescript/src/interactive-api-client/api.ts
```diff
@@ -396,14 +396,13 @@ type WriteAttachmentParams = Omit<IWriteAttachmentRequest, "requestId" | "operat
...
     client.addListener("attachmentUrl", async (response: IAttachmentUrlResponse) => {
```
scytacki, line 401: Because this is going close around the content variable which might be large. Did you check if the listener is automatically removed from the client after the message is received? Otherwise I think this will cause a memory leak.

scytacki (changes requested):
I think I found a couple of more issues.

lara-typescript/src/interactive-api-client/api.ts
kswenson: Yes, it is automatically removed [here](https://github.com/concord-consortium/lara/blob/0f33f2d2fc8e0fd15207c315f5e66a2cb1cd6a5c/lara-typescript/src/interactive-api-client/client.ts#L105-L107) for any message that contains a `requestId`.

lara-typescript/src/interactive-api-client/api.ts
kswenson: 🤦 Fixed

scytacki (approved):
Looks good. Yay!

PR #732, opened 2021-08-06
update local lara and portal integration documentation

scytacki:
This simplifies the documentation to just refer to the Portal documentation, so we aren't maintaining it in two places.

It also removes the portal-https-proxy overlay. This isn't necessary anymore because you can just
set the PORTAL_PROTOCOL variable in LARA's .env file

I've also updated the Portal documentation about all of this:
https://github.com/concord-consortium/rigse/pull/1063


knowuh (approved):
Thank you, Scott.

Makes sense to me.

PR #736, opened 2021-08-11
Clone interactive state so client doesn't get frozen objects

kswenson:
The bug ([[#172312303]](https://www.pivotaltracker.com/story/show/172312303)) found by @tejal-shah in which CODAP shared documents didn't restore properly with the CFM's `InteractiveApiProvider` turned out to be because of this:

>Note that state will become frozen and should never be mutated.
>Each time you update state, make sure that a new object is passed.

CODAP deserialization relies on the ability to change the state object it receives from the CFM which doesn't work unless the state object is cloned somewhere along the line.

This PR is essentially a proposal-for-discussion of the idea that the library should take care of cloning these state objects to prevent future clients from falling into the same trap. Presumably the argument against is performance -- that for some clients we'll be cloning unnecessarily, but I'm not sure how to weigh that against the developer affordance.

Note: in the process of testing this out, I published a `beta` version of the library but forgot to add `.pre.1` to the version (🤦 ) so now version 1.1.3 has been published under the `beta` tag. For clarity, I propose that the next published version of the library should be 1.1.4.

### DIFF
--- a/docs/interactive-api-client/globals.md
+++ b/docs/interactive-api-client/globals.md
@@ -885,22 +885,6 @@ ___
 
 ▸ **setInteractiveState**‹**InteractiveState**›(`newInteractiveState`: InteractiveState | null): *void*
 
-Note that state will become frozen and should never be mutated.
-Each time you update state, make sure that a new object is passed.
-
-Good:
-```
-setInteractiveState(Object.assign({}, previousState, {someProperty: "a new value"}));
-// or
-setInteractiveState({...previousState, someProperty: "a new value"});
-```
-
-Bad:
-```
-previousState.someProperty = "a new value";
-setInteractiveState(previousState);
-```
-
 **Type parameters:**
 
 ▪ **InteractiveState**
--- a/lara-typescript/package-lock.json
+++ b/lara-typescript/package-lock.json
@@ -1833,6 +1833,12 @@
         "@babel/types": "^7.3.0"
       }
     },
+    "@types/clone-deep": {
+      "version": "4.0.1",
+      "resolved": "https://registry.npmjs.org/@types/clone-deep/-/clone-deep-4.0.1.tgz",
+      "integrity": "sha512-bdkCSkyVHsgl3Goe1y16T9k6JuQx7SiDREkq728QjKmTZkGJZuS8R3gGcnGzVuGBP0mssKrzM/GlMOQxtip9cg==",
+      "dev": true
+    },
     "@types/deep-freeze": {
       "version": "0.1.2",
       "resolved": "https://registry.npmjs.org/@types/deep-freeze/-/deep-freeze-0.1.2.tgz",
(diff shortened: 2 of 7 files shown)


pjanik (commented):
I have mixed feelings. I'd still strongly agree with the docs that you removed that show good and bad practices. This applies to many similar hooks. If you look at `useState` hook provided by React itself, you might also run into many problems if you try to store a complex object as a state and mutate its properties. I guess `useState` won't try to clone the state for you. At the same time, it probably won't try to punish you for doing so by freezing the state object as I did in this library. :wink: 

1. Did things fail silently and CODAP just couldn't modify the object? Or was there a clear exception being thrown / error / warning at least?
2. Can't CODAP clone the state object whenever it needs to? Probably less often than here. And if we talk about performance hit, I guess CODAP might be the most affected app (having the biggest state usually).

If freezing causes a silent fail that can cause terrible debugging experience, I definitely agree that it needs to be addressed here somehow. Altough I'd rather try to improve it by making an error clear.

If freezing + modifications in the client app cause an exception being thrown, personally I'd leave it here as it is, and clone it in the client app when necessary.

kswenson:
>Did things fail silently and CODAP just couldn't modify the object? Or was there a clear exception being thrown / error / warning at least?

It fails silently. You can run/step through the `obj.foo = bar` statement without incident, but if you inspect the object in the debugger you'll see that it didn't actually change. The result in this case is you get a failure well down the road deep in the bowels of CODAP's document import code.

>Can't CODAP clone the state object whenever it needs to? Probably less often than here. And if we talk about performance hit, I guess CODAP might be the most affected app (having the biggest state usually).

Sure, the cloning can happen client-side. In this case the CFM's `InteractiveApiProvider` can do it so CODAP doesn't have to worry about it. I've submitted a [CFM PR](https://github.com/concord-consortium/cloud-file-manager/pull/222) with that change for comparison in case we decide to table this PR.

My reason for raising the issue via this PR is that I think this is fundamentally an error-prone API. I just lost about a half-day to it and I'd be willing to bet that more time will be lost to it in the future. The restriction is documented, but I didn't realize that it applied to my situation until I had already been debugging it for several hours. The LARA interactive API exists in large part to simplify things for its clients, and a less error-prone API would be simpler for clients, all else being equal. But, of course, all other things _aren't_ equal. As you suggest, having the library do the cloning would invariably mean that objects are being cloned unnecessarily in some situations. I'm not sure how to weigh the costs/benefits of developer convenience/robustness vs. additional run-time copying.

>If you look at `useState` hook provided by React itself, you might also run into many problems if you try to store a complex object as a state and mutate its properties.

True, because `useState` with a complex object is also an error-prone API that costs developers time as evidenced by the legion of blog posts and StackOverflow questions that discuss the matter. Given the frequency with which `useState` is used with simple values and its centrality to the use of React hooks, the React team made the decision to prioritize performance over robustness, which makes sense given their use case but doesn't necessarily speak to our use case.

Given what I know now, if I were to design this API from scratch I'd be inclined to make it safe-by-default but potentially with an option for clients to opt into the more performant behavior, e.g. a `{ clone: false }` option. That said, the API exists and other clients are using it so the path of least resistance is probably to close this PR and merge the corresponding CFM PR, but I thought it was worth raising the issue for discussion.

scytacki:
Thanks for raising this @kswenson. I agree that for now we shouldn't do the cloning in the lara-interactive-api library.

I wish there as a better solution here. It looks like if 'use strict'  is used then an exception would be thrown: https://stackoverflow.com/questions/9119655/is-there-a-way-to-make-an-object-frozen-object-throw-warnings-when-an-attempt

It seems like webpack adds 'use strict'. So this might only be a problem in older projects which don't use webpack or a build system that adds 'use strict'.

pjanik:
> It fails silently. You can run/step through the obj.foo = bar statement without incident, but if you inspect the object in the debugger you'll see that it didn't actually change. The result in this case is you get a failure well down the road deep in the bowels of CODAP's document import code.

So this is really bad, I fully agree. In this case, I think we should do something in this library to make sure that that error is being thrown. I think I've added freezing hoping for an error being thrown. Maybe my test project was using `use strict` as @scytacki mentioned.

Also, we could just remove freezing. I'd need to check if it has some deeper meaning here and if it's related to some event emitter. If not, I guess it's an option. We won't prevent apps from following questionable patterns, but that's more of a typical approach in most libraries (including react and `useState`).

I think it's worth adding a PT story and prioritizing it somehow. I'm happy to take a look at it after CFM work / when there's a good time.

> Given what I know now, if I were to design this API from scratch I'd be inclined to make it safe-by-default but potentially with an option for clients to opt into the more performant behavior, e.g. a { clone: false } option. T

Re cloning - I think this also has some other issues. I can see scenarios where hidden cloning can even break some apps. If I get a new object **each** time I request my interactive state (well, in practice on each component render), this effectively prevents me from any simple changes detection. I quite often use `useEffect(stateHasBeenUpdatedHandler, [ state]);` pattern to detect state changes and react to them.

PR #739, opened 2021-08-17
Authors cannot change interactive types in the ITSI authoring form

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/178901294

[#178901294]


scytacki (approved):
Looks good to me.

PR #740, opened 2021-08-21
handle saving managed and mw interactive updating from ITSI form

scytacki:
On the ITSI form there is no page, so there is no page_id parameter
So when changing an interactive in ITSI authoring you see an error message
about "save failed".

I initially took the conservative approach of continuing to use the page_id when it is available. But this broke some tests, so then I just decided to make the full change. This simplifies the routes, and makes the interactives behave more like the other embeddables. 

Note: if there is a way for an interactive to be created and not added to the page, then this approach will fail. However I could not find any code that does that.

This PR also removes some dead code: `toggle_visibility`.
That action on the interactive controllers was not being used anywhere.

Instead the edit form was calling `/hideshow_page_item/[id]`. This route calls the action `interactive_pages#toggle_hideshow_page_item`

### DIFF
--- a/app/controllers/interactive_controller.rb
+++ b/app/controllers/interactive_controller.rb
@@ -38,6 +38,9 @@ class InteractiveController < ApplicationController
     if params[:page_id]
       @page = InteractivePage.find(params[:page_id], :include => :lightweight_activity)
       @activity = @page.lightweight_activity
+    elsif @interactive
+      @page = @interactive.page
+      @activity = @interactive.activity
     end
   end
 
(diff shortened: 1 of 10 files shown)


scytacki:
@knowuh I tagged you on this story since you are maintaining a long running branch. Hopefully this doesn't overall with the work you've down so far.

knowuh (approved):
It looks good to me.  

There may be merge conflicts to resolve in my future, but I am not too worried at the moment.

emcelroy (approved):
Looks good 👍

PR #743, opened 2021-08-23
convert old path helpers

scytacki:


emcelroy (commented):
Looks good to me. I tested the changes out locally and everything worked as expected.

PR #745, opened 2021-08-30
Prevent warning message from overlapping Preview button

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/179367834

[#179367834]

This change is a follow up to changes already made for the above story and  will prevent warning messages about activity usage from overlapping the Preview button on the ITSI editing form.


scytacki (approved):
Looks good

PR #749, opened 2021-08-31
179376193 -- fixing tests (WIP)

knowuh:
A PR to track testing …

Not to be merged yet.

### DIFF
--- a/app/models/interactive_page.rb
+++ b/app/models/interactive_page.rb
@@ -34,7 +34,7 @@ class InteractivePage < ActiveRecord::Base
   # has_many :page_items, :order => [:old_section, :position], :dependent => :destroy, :include => [:embeddable]
 
   has_many :sections, :order => :position, :dependent => :destroy, :include => [:page_items]
-  has_many :page_items, through: :sections, order: :position
+  has_many :page_items, through: :sections, order: "sections.position, position"
 
   def toggle_info_assessment
     self[:toggle_info_assessment].nil? ? true : self[:toggle_info_assessment]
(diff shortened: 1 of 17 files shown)


knowuh:
44 test are still failing. But  I think @dougmartin you should review this work as is. If we can get this into master enough of the test suite should be running to unblock https://www.pivotaltracker.com/story/show/179380583.  You should be able to make focused tests pass, even thought the rest of the suite is still being ficed.

I will continue my work in this branch, and add more PRs as more tests start to pass.

FYI @emcelroy ⬆

app/models/interactive_page.rb
```diff
@@ -31,9 +31,10 @@ class InteractivePage < ActiveRecord::Base
   validates :sidebar, :html => true
 
   # PageItem is a join model; if this is deleted, it should go too
-  has_many :page_items, :order => [:old_section, :position], :dependent => :destroy, :include => [:embeddable]
+  # has_many :page_items, :order => [:old_section, :position], :dependent => :destroy, :include => [:embeddable]
```
dougmartin, line 34: We can delete this line in a follow on PR.

app/models/interactive_page.rb
```diff
@@ -167,18 +169,41 @@ def reportable_items
...
+    section_identifier ||= Section::DEFAULT_SECTION_TITLE
+
+    # Local function to test whether section_identifier is a numeric value
+    def numeric (x)
+      Float(x) != nil rescue false
     end
-    section = self.sections.find { |s| s.id == section_id } || self.sections.first
+
+    # look for a section specified by title or ID
+    if !numeric(section_identifier)
```
dougmartin, line 184: Probably best in the future to flip the `!` case since there is an else for this conditional.

dougmartin (approved):
Looks good :+1:

PR #750, opened 2021-09-02
UI: "Add Item" button - and search Library Interactives

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/178595247

[#178595247]

### DIFF
--- a/lara-typescript/src/section-authoring/components/section-item-picker.tsx
+++ b/lara-typescript/src/section-authoring/components/section-item-picker.tsx
@@ -1,45 +1,158 @@
-import * as React from "react";
+import React, { useState } from 'react';
+
+import "./section-item-picker.css"
 
 export interface ItemType {
-  name: string
+  name: string,
+  use_count: number,
+  date_added: number
 }
 
 export interface IProps {
   availableItemTypes: ItemType[],
-  quickAddItems: ItemType[]
+  quickAddItems: ItemType[],
+  allItems: ItemType[],
+  matchingItems: ItemType[]
 }
 
 export const SectionItemPicker: React.FC<IProps> = (props) => {
-  const { quickAddItems } = props;
-  return (
-    <div>
-      <header>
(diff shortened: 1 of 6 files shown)


lara-typescript/src/section-authoring/components/section-item-picker.tsx
```diff
@@ -1,45 +1,158 @@
...
+  const [itemSelected, setItemSelected] = useState(false);
+  const [currentSelectedItem, setCurrentSelectedItem] = useState("");
+  const [allItemsList, setAllItemsList] = useState(allItems);
+  const [isSearching, setIsSearching] = useState(false);
+
+  const setItemClasses = (isSelectedItem: boolean) => {
+    const classes = isSelectedItem ? "assessmentItemOption selected" 
+                                   : currentSelectedItem !== "" 
+                                     ? "assessmentItemOption disabled"
+                                     : "assessmentItemOption";
```
kswenson, line 30: This is the kind of code that warrants the use of `classNames()`:
```suggestion
    const classes = classNames("assessmentItemOption", { selected: isSelectedItem, disabled: currentSelectedItem !== "" });
```
It looks like `classNames()` is not already being included, but I'd be inclined to add it since it's small and likely to be useful elsewhere in the `lara-typescript` project.

emcelroy: @ksweson Thanks 👍  Seems like a good idea.

knowuh (approved):
Ethan this looks good to me.  👍 

We will eventually have to:

- add ID values to the ItemType
- use the IDs for checking selection status
- provide an optional `setItemFunc?:` component property (set outside the component to actually hit the API for adding items)
- (maybe?) provide an optional `searchFunc?:` component property (set outside the component to actually hit the API for searching items)

There is no harm in merging this as it is IMHO. I understand you want to refactor the CSS a bit.  Maybe we can even do that in a future PR.

PR #752, opened 2021-09-03
Rename interactive-api-parent to interactive-api-lara-host

pjanik:
[#178367125]

I think keeping `interactive-api-parent` could be slightly confusing since I've added `interactive-api-host`. I've renamed it to `interactive-api-lara-host` following @kswenson's suggestion. :)


lara-typescript/src/interactive-api-client/hooks.spec.tsx
```diff
@@ -1,7 +1,7 @@
 import * as React from "react";
 import { act, renderHook } from "@testing-library/react-hooks";
 import { render } from "@testing-library/react";
-import { mockIFramePhone, MockPhone } from "../interactive-api-parent/mock-iframe-phone";
+import { mockIFramePhone, MockPhone } from "../interactive-api-lara-host/mock-iframe-phone";
```
scytacki, line 4: It doesn't have to be now, but it seems like it would be good to move the `mock_iframe-phone` out of the lara-host folder. It is confusing that the interactive-api-client has a dependency on the lara-host.

scytacki (approved):
looks good to me.
I just left one comment that can be be addressed later.

PR #753, opened 2021-09-07
Interactive API Host: don't mutate function argument

pjanik:
[#178367125]

Generally, I think it's a good practice not to mutate the function arguments (unless that's a main task of the function).

In this particular case, `handleGetAttachmentUrl` was updating `answerMeta` fields and it gave me quite a bit of headache while debugging LARA. Iframe saver was comparing new and old objects before sending it to the server. Since the old one was already modified, it was never happening.

Btw, I think this PR is also introducing a new bug that I'll fix in a subsequent PR. I'll leave an inline comment about that.


lara-typescript/src/interactive-api-host/attachments-api/handle-get-attachment-url.ts
```diff
@@ -16,35 +18,37 @@ export const handleGetAttachmentUrl =
...
         options.writeOptions.onAnswerMetaUpdate(answerMeta);
       }
     }
     else if (operation === "read") {
-      if (attachmentsFolder && attachments && attachments[name]) {
+      if (answerMeta.attachmentsFolder && answerMeta.attachments && answerMeta.attachments[name]) {
         // TODO: this won't work for run-with-others where we won't have a readWriteToken
-        const attachmentInfo = { ...attachments[name], folder: attachmentsFolder };
-        response.url = await attachmentsMgr.getSignedReadUrl(attachmentInfo, { expiresIn });
+        response.url = await attachmentsMgr.getSignedReadUrl(answerMeta.attachments[name], { expiresIn });
```
pjanik, line 51: I tried to simplify things, but I think this is actually breaking readWriteToken case. Based on the TS types/interfaces, this line:
`const attachmentInfo = { ...attachments[name], folder: attachmentsFolder };`
didn't make too much sense, so I just removed it. But now I'm seeing that `attachmentsFolder` might include read write token that is never part of the `answerMeta.attachments[name]`. I'll try to clarify interfaces so TS compiler can find this bug too.

pjanik: This is fixed here: https://github.com/concord-consortium/lara/pull/754

scytacki (approved):
This seems good to me. However since it would cause a regression I don't want to merge it until the PR with the readWriteToken fix is ready to go.

pjanik:
The other fix is merged into this PR, so I'm also merging in this one.

PR #754, opened 2021-09-07
Simplify Attachments API metadata and types, add run with others support

pjanik:
[#178367125]

Previously attachments API was saving following metadata:
```
attachmentsFolder: {
 id: ...,
 readWriteToken: ... // optional
},
attachments: {
  "file.json": {
    folder: { 
      // pretty much copy of the top-level attachmentsFolder, but without readWriteToken (not sure why?)
      id: ... 
    },
    publicPath: ...
  }
}
```
Now it's reduced to:
```
attachmentsFolder: {
 id: ...,
 readWriteToken: ... // optional
},
attachments: {
  "file.json": {
    publicPath: ...
  }
}
```

It doesn't seem it makes sense to duplicate folder info in attachments hash. All the attachments will share it anyway. Am I missing here some design decision?

Also, some types and lines of code are clarified. For example, I've removed `IWritableAttachmentsFolder` as it didn't seem to be useful. First, `readWriteToken` isn't used only to write. It's also used to read attachments. Second, this type was a bit hidden (not defined explicitly, some casting with `isWritableAttachmentsFolder`, etc.) and I was slightly bitten by that during random refactoring (I broke support for anonymous users). So, I've just moved `readWriteToken` to IAttachmentFolder type as an optional property.


scytacki:
The case I would be worried about is with run with others. Currently this can't be done anonymously, but perhaps we'd support it in the future. But if we just look at the case of logged in users: 

Let's say student A & B work together and upload an attachment. Student A launched the activity. So the resource in the token service will be owned by this user. I'm guessing this id is the id in the folder structure. So following the run with others approach this interactive state will be copied to student B too. To support run with others we'll allow student B to get read only credentials from the resource in the token service. These credentials would be used to get a singed url from the public url of the attachments. All of this is OK so far with the simplified structure.

But then student B runs independently. And the interactive they are using supports multiple attachments. They upload a second attachment. In this case they will need their own resource in the token service which gives them write access to a folder to upload this second attachment. The first attachment would still be located in the old folder. So in this scenario student B needs to get two sets of credentials one for the first attachment and one for the second. To support that the folder structure needs to be set on each attachment. 

Alternatively, if we want to keep the simplified structure then we'd student be to copy all of the attachments into their own folder so they could continue to read them after the top level folder information was changed. Because attachments can be large (for example movies), I think it is worth while to avoid this copying.

pjanik:
We discussed that in person on Zoom. The old implementation also wouldn't handle it. So, it seems the best way would be to extend the implementation to support run with others and at the same time clean up folder metadata. I've added a PT story for it: https://www.pivotaltracker.com/story/show/179499657. I can push more changes here, or we can close this PR and I'll open a new one (I'll reuse the same branch anyway).

scytacki:
It seems good to me to just push more changes to this existing PR.

pjanik:
@scytacki, I've updated the implementation as we discussed. Tests of `findFolder` helper should explain how it works. `attachmentFolder` meta is now completely removed.

scytacki (approved):
Looks good to me. 

@pjanik I'll let you merge this as appropriate.

lara-typescript/src/interactive-api-host/attachments-api/handle-get-attachment-url.ts
```diff
@@ -29,25 +52,26 @@ export const handleGetAttachmentUrl =
...
       const urlOptions: ISignedWriteUrlOptions = { ContentType: contentType, expiresIn };
-      const [writeUrl, attachmentInfo] =
-        await attachmentsMgr.getSignedWriteUrl(answerMeta.attachmentsFolder, name, urlOptions);
+      const [writeUrl, attachmentInfo] = await attachmentsMgr.getSignedWriteUrl(folder, name, urlOptions);
       response.url = writeUrl;
       // public path changes with sessionId
-      if (!answerMeta.attachments[name] || answerMeta.attachments[name].publicPath !== attachmentInfo.publicPath) {
+      if (!answerMeta.attachments[name] ||
+          answerMeta.attachments[name].publicPath !== attachmentInfo.publicPath ||
+          answerMeta.attachments[name].folder.id !== attachmentInfo.folder.id) {
```
kswenson, line 68: May want to update the comment in line 65 now that `folder.id` is involved.

kswenson (commented):
👍  LGTM -- late to the party since I was on vacation last week -- just a couple of minor suggestions that can be dealt with later (if desired) now that the PR has been merged.

>It doesn't seem it makes sense to duplicate folder info in attachments hash. All the attachments will share it anyway. Am I missing here some design decision?

As @scytacki suggested in his response, this was my best-guess attempt at what might be needed to support run-with-others. My thinking was that sometimes the contents of an attachment would be pulled from a user's own attachments folder and sometimes it would need to be pulled from another user's attachments folder. I freely admit, however, that I never took the time to understand the actual requirements of run-with-others so I'm happy to have you change it now that you and @scytacki have apparently thought about the problems more deeply.

PR #756, opened 2021-09-07
Add Attachments API support

pjanik:
[#178367125]

This PR adds Attachments API support. Note that I'm not referencing files in the sibling `interactive-api-host` folder, but I just installed `@concord-consortium/interactive-api-host` NPM package. This should make it easier to develop new versions of the host package without worrying about LARA host env and to migrate it to a separate repo in the future.

`InteractiveRunState#report_service_hash`  has become a bit confusing. We have there two kinds of metadata. One is encoded in the interactive state, another is set separately in the new `metadata` field. Not sure if there's a better way to handle that.

Theoretically, attachments could be also stored in the interactive state too. Or we could migrate all the other metadata to be stored out of it. But this would require a new Interactive API to save this metadata. Or just live with two different ones. The positive thing is that the current division makes it as easy as possible for the interactive developer (I think). It's pretty convenient to set `submitted: true` in the interactive state instead of calling some API to do it. At the same time, Attachments API hides the metadata from the developer completely.


app/helpers/interactive_run_helper.rb
```diff
@@ -48,6 +50,7 @@ def interactive_data_div(interactive,run)
     data['interactive-id'] = interactive.id
     data['interactive-name'] = interactive.name
     data['linked-interactives'] = interactive.linked_interactives_list.to_json
+    data['token-service-env'] = ENV["RAILS_ENV"] == "production" ? "production" : "staging"
```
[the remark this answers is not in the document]
pjanik: Thanks, it makes sense. I've forgotten that ENV["RAILS_ENV"] will be equal to production on staging too. I've fixed that in the last commit. I'm using the same approach as the glossary-plugin now, so it's at least a bit consistent.

pjanik:
So, I've addressed the inline comment, but I'm stuck with the failing tests. The Capybara tests are now failing. There's a JS parse error. I guess it's caused by the new NPM packages that I installed and that are now bundled in `lara-typescript.js`. Probably the old Poltergeist/PhantomJS driver doesn't like it. I guess it can be quite a few things, when I was commenting out some code and packages I've seen an unsupported regex format (but not able to tell from), but otherwise, I can just see JS `Parse Error`.

I tried the following:
- upgrade Capybara and Poltergeist minor versions -> didn't help
- use a new Capybara driver `apparition` -> it's meant to be backward compatible with Poltergeist, but it requires Capybara 3.12+, and we can't install it because of the rack/rails version we're locked to
- Downgrade TS target to ES3 -> compilation fails, as some of the code we use requires target ES5.

scytacki (approved):
Looks good to me. 👍

lara-typescript/src/interactive-api-lara-host/iframe-saver.ts
```diff
@@ -206,6 +230,53 @@ export class IFrameSaver {
     });
   }
 
+  public saveMetadata(metadata: object) {
+    if (!this.learnerStateSavingEnabled()) { return; }
+
+    if (JSON.stringify(metadata) === JSON.stringify(this.metadata)) {
```
kswenson, line 236: `JSON.stringify` is not idempotent, i.e. it's not guaranteed to stringify two different objects with the same contents identically. In particular, the order of properties is not guaranteed. If you mean for this to handle the case of two different equivalent objects (i.e. when `metadata !== this.metadata`), then you may want to use something like [deep-equal](https://www.npmjs.com/package/deep-equal) instead.

kswenson (commented):
>Theoretically, attachments could be also stored in the interactive state too.

Interactive state is stored in firestore documents which are limited to 1MB. The attachments API exists principally to work around that limitation.

pjanik:
> Interactive state is stored in firestore documents which are limited to 1MB. The attachments API exists principally to work around that limitation.

Yeap. I've meant attachments metadata, not attachments themselves.

lara-typescript/src/interactive-api-lara-host/iframe-saver.ts
pjanik: True. I think I'll just remove this check, as it doesn't seem to be worth it anyway. There are no use cases when this optimization would be used in practice.

PR #758, opened 2021-09-08
Added API endpoint to fetch list library interactives [#179376065]

dougmartin:


knowuh (approved):
Synchronous review with Doug

PR #760, opened 2021-09-09
Fix assets compilation (transform regexp named capture groups)

pjanik:
I didn't find a cleaner solution than this. babel-preset-env includes this plugin, but I tried to use it alone. Hopefully, one day TS compiler will do something similar, or we don't need to use Rails assets compilation.

### DIFF
--- a/lara-typescript/package-lock.json
+++ b/lara-typescript/package-lock.json
@@ -53,11 +53,11 @@
       }
     },
     "@aws-sdk/abort-controller": {
-      "version": "3.25.0",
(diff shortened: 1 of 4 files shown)


scytacki (approved):
Looks good to me.

I looked to see if ts-polyfill had a solution for this. That is what we use in other projects.
Slack Thread about it: 
https://concord-consortium.slack.com/archives/C0303QUNU/p1616987480044400

But it does not:
https://github.com/ryanelian/ts-polyfill/issues/1

PR #763, opened 2021-09-10
Use Portal Report instead of the summary page

pjanik:
[#179499612]

The URL generation is partially based on AP code:
https://github.com/concord-consortium/activity-player/blob/master/src/utilities/report-utils.ts
I didn't hardcode the report URL, but added a new ENV variable instead.

Some URL params are encoded, some are not. I've followed AP pattern. I don't want to break Report assumptions about that. But maybe all the URLs should be encoded? 

Also, I'm adding `activityIndex` URL param when there's a sequence. The old LARA summary was always showing only one activity. Based on the places where these links are, I think this makes more sense. AP doesn't seem to do it, but not sure if it's intended or it's just missing.

I'll open a separate PR with the dead code removed (=> https://github.com/concord-consortium/lara/pull/764).

### DIFF
--- a/.env-osx-sample
+++ b/.env-osx-sample
@@ -41,3 +41,6 @@ PORTAL_HOST=app.portal.docker
 REPORT_SERVICE_URL=https://us-central1-report-service-dev.cloudfunctions.net/api
 # Update to be the development token. This can be found in the Cloud Formation lara-ecs-staging stack parameters.
 #REPORT_SERVICE_TOKEN=
+
+# Report URL is used to generate student report links.
+# REPORT_URL=https://portal-report.concord.org/branch/master/index.html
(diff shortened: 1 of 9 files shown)


scytacki (changes requested):
Mostly looks good. I just found one potential issue with the source_key calculation.

I agree the escaping should be unified. I'm pretty sure that you could always escape the parameters and the portal report could handle it. 

Ideally you'd use a URL utility to construct the params, instead of strings. This way the REPORT_URL could have parameters too and also they would be automatically be escaped. An example is this:
https://github.com/concord-consortium/lara/blob/78189582f98a6d98b48e3c9afef7174a0b3b5d5d/app/helpers/lightweight_activity_helper.rb#L71-L76

But it seems fine to leave what you've got.

pjanik:
Thanks. I've addressed both issues.

scytacki (approved):
Looks good to me.

PR #764, opened 2021-09-13
Remove code related to the activity summary page

pjanik:
[#179499612]

I'm not 100% sure about changes in `app/views/runs/dirty.html.haml` and `#resubmit_answers` code. It seems that the summary page was also used by admins - by the code that was related to view dirty runs and resubmit answers. I've assumed that this is not critical, as it's only used by admins and they can inspect student work in other ways.


app/controllers/lightweight_activities_controller.rb
```diff
@@ -122,16 +122,6 @@ def print_blank
     setup_show
   end
 
-  def summary
```
scytacki, line 125: It is possible there will be some old summary links out there that students sent to teachers. So it would be better if this action would redirect to the new portal-report links.

scytacki (changes requested):
It is wonderfully to remove all of this code.

My only requested change is to redirect from old summary report urls to the new report urls. When teachers want to use LARA outside of the portal, we instruct them to have their students "turn in" these summary urls when they are done with their assignment. 
So if we release this in the middle of the school year those turned in URLs would break.

pjanik:
It makes sense. I've implemented this redirect.

scytacki (approved):
Looks good.

PR #768, opened 2021-09-14
Run with others: copy interactive run state metadata

pjanik:
[#179578041]

This is necessary if we want to support run with others + interactives that use Attachments API.


scytacki (approved):
looks good

PR #770, opened 2021-09-14
Update lightweight_activity_controller_spec

pjanik:


scytacki (approved):
looks fine to me.

I wonder if the order of the parameters is going to be kind of random with the use of `URI.query`, but hopefully not.

PR #775, opened 2021-09-16
Experiments with React Query and API Contexts

knowuh:
This PR includes a refactor of how application state is managed, as well as several smaller reorganizations (eg: moving most API primitive types into `api-types.ts`). It also includes a few new LARA routes, controllers, and specs for deleting & creating pages.

The main change is that API implementation has a façade interface called `IAuthoringAPIProvider` so that the API implementation can be switched out.  For examples, see: `query-providers.stories.tsx`

Here is an example of how ReactQuery is used to create a mock API for the app:

```html
    <APIContext.Provider value={mockProvider}>
      <QueryClientProvider client={queryClient}>
        <Content />
      </QueryClientProvider>
    </APIContext.Provider>
```

The `<Content />` component can then consume the Queries and Mutations offered by the API like so:

```
    ...
    const api = usePageAPI(); 
    const addSection = () => api.addSectionMutation.mutate(p.id);
   ...
```

This PR has gotten quite large. Future work will be removing all the prop-drilling mutation methods on the components into local `usePageAPI()` calls.


knowuh:
@dougmartin I don't want to merge this code, but I would like you to look at it to see what you think. 

The big idea is that React Query doesn't care what the functions are that do the data fetching.  So we can just fake backend calls with one API implementation, and actually make the calls with the other. 

We can set the implementation using Contexts, so we just set it once in an outer wrapper.

kswenson (commented):
Seems like ReactQuery could be a good fit for this purpose.

knowuh:
@dougmartin I added quite a few commits to this sense your last look.  I would love another review if you get a chance.

db/schema.rb
```diff
@@ -432,7 +432,6 @@
     t.boolean  "student_report_enabled",                 :default => true
     t.text     "last_report_service_hash"
     t.boolean  "show_submit_button",                     :default => true
-    t.boolean  "activity_player_only"
```
emcelroy, line 435: Was removing this intentional?

knowuh:
@emcelroy OGM thanks for catching that error in the schema change.  😱  I will remove that change from the PR, thank you.

lara-typescript/src/section-authoring/api/lara-api-provider.ts
```diff
@@ -0,0 +1,88 @@
+import {
+  IPage, PageId, ISection,
+  APIPageGetF, APIPagesGetF, IAuthoringAPIProvider,
+  ICreatePageItem
+} from "./api-types";
+
+const APIBase = "/api/v1";
+
+export const getLaraPageAPI = (host: string = "", activityId: string): IAuthoringAPIProvider => {
```
dougmartin, line 9: It is kind of weird that the first parameter to `getLaraPageAPI` is optional and has a default value but the second one is required.

lara-typescript/src/section-authoring/api/lara-api-provider.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  const prefix = `${host}/${APIBase}`;
```
dougmartin, line 11: Just a note - this is going to generate double slashed urls like `https://app.lara.docker//api/v1` since `APIBase` starts with a `/`

lara-typescript/src/section-authoring/api/lara-api-provider.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  const getPages: APIPagesGetF = () => {
+    return sendToLara({url: getPagesUrl});
+  };
+
+  const getPage: APIPageGetF = (id: PageId) => {
+    return sendToLara({url: getPageUrl(id)});
+  };
+
+  const createPage = () => {
```
dougmartin, line 57: Just curious why there isn't a corresponding `APIPagesPostF` type for the post functions?

lara-typescript/src/stories/query-providers.stories.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { usePageAPI } from "../section-authoring/api/use-api-provider";
+import { QueryClient, QueryClientProvider } from "react-query";
+import { ISection } from "../section-authoring/api/api-types";
+
+export default {
+  title: "Query Providers Test",
+  component: App,
+} as ComponentMeta<typeof App>;
+
+// TODO Parameterize host and PageID
```
dougmartin, line 18: Is this a TODO for later or one you wanted to do for this PR?

dougmartin (approved):
Looks good - just a few non-merge blocking comments.

knowuh:
Nice Catches @dougmartin -- Resolved in a668c14

PR #776, opened 2021-09-16
Add external_report_url to InteractiveRunState JSON, update Plugin API to use it as reporting_url

pjanik:
[#179466138]

I've moved Portal Report links generation to `ReportService` module, as it seems like a logical place and let me share this logic easily. However, that forced me to use REPORT_SERVICE_SELF_URL to construct activity or sequence URLs (this module doesn't have access to helpers that are available in the controller/helper context). But this should be fine, right?
I've reorganized tests a bit not to duplicate them too much (use stubs instead).


scytacki (approved):
Looks good to me

PR #777, opened 2021-09-17
Implement PluginAPI#setAnswerSharedWithClass, update InteractiveRunState controller to merge in new metadata

pjanik:
[#179621794]

This PR adds Plugin API function: `setAnswerSharedWithClass`.
Also, the InteractiveRunState controller is slightly updated to merge metadata instead of replacing it.

### DIFF
--- a/app/controllers/api/v1/interactive_run_states_controller.rb
+++ b/app/controllers/api/v1/interactive_run_states_controller.rb
@@ -27,7 +27,7 @@ class Api::V1::InteractiveRunStatesController < ApplicationController
         @interactive_run_state.learner_url = params['learner_url']
       end
       if params.has_key?('metadata')
-        # Note that metadata is always merged, not overwritten, as multiple apps and clients can set its own metadata.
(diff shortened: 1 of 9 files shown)


app/controllers/api/v1/interactive_run_states_controller.rb
```diff
@@ -26,7 +27,11 @@ def update
         @interactive_run_state.learner_url = params['learner_url']
       end
       if params.has_key?('metadata')
-        @interactive_run_state.metadata = params['metadata']
+        # Note that metadata is always merged, not overwritten, as multiple apps and clients can set its own metadata.
```
kswenson, line 30:
```suggestion
        # Note that metadata is always merged, not overwritten, as multiple apps and clients can set their own metadata.
```

pjanik: Thanks. I'm starting to think that you have some Gmail plugin that scans all your emails from Github and checks grammar :wink:

scytacki (approved):
Looks good.

PR #778, opened 2021-09-20
Integrate add item frontend to backend [#179376054]

dougmartin:


dougmartin:
@knowuh there is still an issue with the modal class so I left in the `.react-modal` class renaming so the modal shows.  I think this is best fixed when/if Ethan updates the modals.

knowuh (approved):
looks good to me.  Thank you.

PR #781, opened 2021-09-21
179658460 clean up post merge

knowuh:
These are mostly all white-space / linting issues that were fixed by running `npm run lint:fix` -- but there were some hand tweaks too.


knowuh:
@emcelroy these should all be safe changes, but it would be nice to get your eyes on them too.  If you are swamped I can ask Doug.

emcelroy (approved):
Looks good to me 👍 One little thing I'm not sure is even an issue is the `console.log` on line 230 of authoring-section.tsx. Should that stay?

knowuh:
@emcelroy nice catch, removed in 484c78f

PR #782, opened 2021-09-22
Send answers from anonymous users to the report service [#179623345]

dougmartin:
- Removed check for remote_endpoint from run.rb@queue_for_portal
- Removed unused answer parameter from run.rb@queue_for_portal
- Added missing run_spec.rb spec tests
- Fixed existing test by disabling delayed jobs around the rub_spec.rb dirty_answers test
- Added fallback to run.key for platform_user_id
- Added (untested) rake task to publish anonymous runs to the report service

### DIFF
--- a/app/models/run.rb
+++ b/app/models/run.rb
@@ -259,7 +259,7 @@ class Run < ActiveRecord::Base
   def queue_for_portal
     # Don't check remote_endpoint so we can allow anonymous runs to be enqueued and sent to the report service.
     # The submit_dirty_answers will enqueue a job that will eventually call #send_to_portal which will drop
-    # it on the floor silenty but will then continue and send the answer to the report service.
+    # it on the floor silently but will then continue and send the answer to the report service.
     return false if answers.empty?
     if dirty?
       # no-op: only queue one time
--- a/app/models/run.rb
+++ b/app/models/run.rb
@@ -257,7 +257,9 @@ class Run < ActiveRecord::Base
   end
 
   def queue_for_portal(answer)
-    return false if remote_endpoint.nil? || remote_endpoint.blank?
+    # Don't check remote_endpoint so we can allow anonymous runs to be enqueued and sent to the report service.
+    # The submit_dirty_answers will enqueue a job that will eventually call #send_to_portal which will drop
+    # it on the floor silenty but will then continue and send the answer to the report service.
     return false if answers.nil?
     if dirty?
       # no-op: only queue one time
--- a/app/models/run.rb
+++ b/app/models/run.rb
@@ -256,11 +256,11 @@ class Run < ActiveRecord::Base
     is_success
   end
 
-  def queue_for_portal(answer)
+  def queue_for_portal
     # Don't check remote_endpoint so we can allow anonymous runs to be enqueued and sent to the report service.
     # The submit_dirty_answers will enqueue a job that will eventually call #send_to_portal which will drop
     # it on the floor silenty but will then continue and send the answer to the report service.
-    return false if answers.nil?
+    return false if answers.empty?
     if dirty?
       # no-op: only queue one time
     else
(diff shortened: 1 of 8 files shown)


spec/models/run_spec.rb
```diff
@@ -454,6 +475,32 @@
...
+        it "returns the enqueued job when there is no remote_endpoint" do
+          run.open_response_answers << or_answer
+          expect(run.queue_for_portal).to be_instance_of(Delayed::Backend::ActiveRecord::Job)
+        end
+
+        describe "and there is a remote_enpoint" do
+          let(:remote_endpoint) { valid_remote_endpoint }
+
+          it "returns the enqueued job" do
+            stub_http_request(:post, remote_endpoint).to_return(:body   => "OK", :status => 200)
```
scytacki, line 497: If the delayed jobs are stopped, is it necessary to stub this?

dougmartin: Delayed jobs are not stopped in this suite of tests - only for the suite that checks the dirty flag (as the job clears that flag before the test can check it).

app/models/run.rb
```diff
@@ -256,9 +256,11 @@ def send_to_portal(answers, start_time = Time.now)
     is_success
   end
 
-  def queue_for_portal(answer)
-    return false if remote_endpoint.nil? || remote_endpoint.blank?
-    return false if answers.nil?
+  def queue_for_portal
+    # Don't check remote_endpoint so we can allow anonymous runs to be enqueued and sent to the report service.
+    # The submit_dirty_answers will enqueue a job that will eventually call #send_to_portal which will drop
+    # it on the floor silenty but will then continue and send the answer to the report service.
```
kswenson, line 262:
```suggestion
    # it on the floor silently but will then continue and send the answer to the report service.
```

lib/tasks/reporting.rake
```diff
@@ -83,6 +83,22 @@ namespace :reporting do
     send_all_resources(runs, ReportService::RunSender, opts)
   end
 
+  desc "publish anonymous runs to report service"
+  task :publish_anonymous_runs => :environment do
+    runs = Run.where('remote_endpoint is null')
+    # allow caller do specify the limit and offset of the query
+    limit_env_value = ENV["REPORT_PUSH_RUN_LIMIT"]
+    if limit_env_value && limit_env_value.size > 0
```
scytacki: You could use `limit_env_value.present?` instead of checking for the size of the string.

dougmartin: I initially copied that from the original rake task but have since deleted it completely as it doesn't play well with the `find_in_batches` methods own limit/offset calculation

scytacki: I was wondering about that. So maybe it should be limited by a date. Or you could add an option to send_all_runs which limits the number of batches.

dougmartin: Unfortunately limits where only added to find_in_batches in Rails 5.1:

https://www.bigbinary.com/blog/rails-5-1-adds-support-for-limit-in-batch-processing

Lara is on 3.2.22 right now.

dougmartin: I added min/max id options to the anonymous run rake task in this commit (plus I fixed a few typos in the existing code):

https://github.com/concord-consortium/lara/pull/782/commits/16491ed9bbadb3d9a861b37367c77a11a2b03354

db/schema.rb
```diff
@@ -463,7 +463,7 @@
     t.datetime "updated_at",   :null => false
   end
 
-  add_index "linked_page_items", ["primary_id", "secondary_id", "label"], :name => "index_linked_page_items_unique", :unique => true
+  add_index "linked_page_items", ["primary_id", "label"], :name => "index_linked_page_items_unique", :unique => true
```
scytacki, line 466: Is this intentional?

scytacki (changes requested):
Seems like schema change should be removed from the PR.

db/schema.rb
dougmartin: Nope.

scytacki (approved):
looks good

PR #790, opened 2021-09-23
Fix section editing fonts are incorrect when viewed in LARA

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/179677152

[#179677152]

As far as I could tell, the only issues with the fonts were the curved separators (actually a border issue) and some elements not having the right font family applied. The curved separators are fixed by setting `border-radius` to 0 for the header buttons.


knowuh (approved):
Looks good, thank you

PR #791, opened 2021-09-24
switch sso client id localhost to authoring

scytacki:
this is closer to what the name is on production,
and then the same client id can be be used for both local and cloud deployments

This is needed for these portal changes:
https://github.com/concord-consortium/rigse/pull/1086


scytacki:
@dougmartin I added you as a reviewer to this one, so you are aware of it as you are working on LARA and Portal stuff.

dougmartin:
👍 Thanks for the mention - this would probably have bitten me the next time I updated master.

PR #792, opened 2021-09-24
Page Settings Menu

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/179597704

[#179597704]

Adds a page settings menu. Also adds a header for the page editing view that includes the page title and settings menu toggle button.

### DIFF
--- a/lara-typescript/src/page-settings/components/page-settings-dialog.tsx
+++ b/lara-typescript/src/page-settings/components/page-settings-dialog.tsx
@@ -59,7 +59,6 @@ export const PageSettingsDialog: React.FC<IPageSettingsDialogProps> = ({
   };
 
   const handleUpdateSettings = () => {
-    console.log(title);
     updateSettingsFunction(
       title,
       isCompletion,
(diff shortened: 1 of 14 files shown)


lara-typescript/src/page-settings/components/page-settings-dialog.tsx
```diff
@@ -0,0 +1,164 @@
...
+    hasStudentSidebar = !hasStudentSidebar;
+  };
+
+  const handleHasTESidebarChange = () => {
+    hasTESidebar = !hasTESidebar;
+  };
+
+  const handleUpdateSettings = () => {
+    console.log(title);
+    updateSettingsFunction(
```
knowuh, line 63: I am going to checkout this code locally and look closer, but is there an example of passing this property somewhere?

knowuh: I found it -- sorry.

knowuh (changes requested):
This looks good, I think we should use consistent camelCasing on the name attributes for those form fields though.

knowuh (approved):
Looks great, thanks!

PR #772, opened 2021-09-15
Deploy storybook

scytacki:
This uses the s3-deploy-action to deploy storybook to models-resources:
https://github.com/concord-consortium/s3-deploy-action/pull/86

This LARA PR currently is using a branch of the s3-deploy-action code. I'm hoping I can get that s3-deploy-action branch merged shortly, and then I'll update this PR to refer to a tagged version of the s3-deploy-action code.

PR #793, opened 2021-09-27
[WIP] Create a sample interactive for state/linked state [#179726367]

dougmartin:
- Renamed example-interactive to example-interactives


scytacki (changes requested):
This looks good, but shouldn't there be some more changes in the `public/` folder so the 2 new interactives are available at `/example-interactives/testbed` and `/example-interactives/linked-state` ?

scytacki:
As a note for the future: We could now pretty easily deploy the example interactives to the model-resources bucket. So then it would be easier for developers to make changes or new interactives and have their branches be auto deployed without waiting for a LARA deploy. 

The pattern to follow is what I did here for deploying storybook: 
https://github.com/concord-consortium/lara/pull/772/files

That PR was merged into the `new-sections` branch instead of master that is why you don't see the .github folder on master.

dougmartin:
@scytacki Thanks for catching the lack of /public update.  I've fixed that in this commit: https://github.com/concord-consortium/lara/pull/793/commits/feeef17bc5743696efc5e87ffc157f5ec284fa7a

PR #794, opened 2021-09-27
Add hover and active states for section header menu buttons.

emcelroy:
Add hover and active states for section header menu buttons.

PT Story: https://www.pivotaltracker.com/story/show/178648784

[#178648784]


knowuh (approved):
These hover states look great.

Thank you

PR #795, opened 2021-09-27
Fix modals do not appear.

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/179693528

New class name for modal component to avoid conflicts with existing modal. Small CSS tweaks to offset other style conflicts.


knowuh (approved):
Looks good. Thank you.

PR #796, opened 2021-09-27
Removed old page nave, info, and preview

knowuh:


emcelroy (approved):
Looks great -- thanks!

PR #800, opened 2021-10-01
use new docker-rails-base

scytacki:
this fixes an issue with the https root certificates
it also bumps the version of rails being used slightly,
and upgrades all of the linux libraries

Here are the changes to the docker-rails-base:
https://github.com/concord-consortium/docker-rails-base/commit/783eb11625433df07570a944812ed832979653f1

I put the changes in docker-rails-base instead of the LARA image itself, so that the upgrade of the linux libraries doesn't slow down every build of the LARA image.


dougmartin (approved):
Looks good :+1:

PR #804, opened 2021-10-04
Fix item order bug after drag-and-drop.

emcelroy:
[#179239933]

@knowuh This fixes the ordering problem you discovered by replacing `swapIndexes` with a new `updateItemPositions` method. The latter has some similar code to `handleMoveItem` in the AuthoringPage component. So I suspect it's something we'll want to refactor and combine with that somehow, but it seems like for the time being this is good enough. Please let me know if you think otherwise, though.


knowuh (approved):
Looks good, thank you.

PR #806, opened 2021-10-07
179783183 move section item with menu

knowuh:
@emcelroy -- This is a pretty big change that is underway.

The main thing I want to point out is that transient UI state is being centralized in `lara-typescript/src/section-authoring/api/use-user-interface-context.ts`

You can see the UI action being used here: 
``` 
    ...
    const { actions: {setMovingSectionId}} = React.useContext(UserInterfaceContext);
    ...
    const handleMoveSection = () => {
       setMovingSectionId(id);
    };

```

This is the same approach I think we should use for marking an item for editing, maybe using a `setEditingItemId` and checking that value to bring up the editing dialog box.  We will also have to set that after an item has been created. (TBD).

https://github.com/concord-consortium/lara/compare/new-sections...179783183-move-section-item-with-menu?expand=1#diff-034158d87db6f3522b91b12e3854d85d5700c60e4eabfdd611868a00b87b6a47R80

### DIFF
--- a/lara-typescript/src/section-authoring/api/use-user-interface-context.ts
+++ b/lara-typescript/src/section-authoring/api/use-user-interface-context.ts
@@ -1,0 +1,61 @@
+import * as React from "react";
+import { useImmer } from "use-immer";
+
+interface IUserInterface {
+  currentPageIndex: number;
+  movingItemId: string | false;
+  movingSectionId: string | false;
+  editingItemId: string | false;
+}
+
+const defaultUI: IUserInterface = {
+  currentPageIndex: 0,
+  movingItemId: false,
+  movingSectionId: false,
+  editingItemId: false
+};
+
+interface IUIActions {
+  setMovingItemId: (id: false|string) => void;
+  setMovingSectionId: (id: false|string) => void;
+}
+
+interface IUIContext {
+  userInterface: IUserInterface;
+  actions: IUIActions;
+}
+
+const defaultContext: IUIContext = {
+  userInterface: defaultUI,
+  actions: {
+    // tslint:disable-next-line
+    setMovingItemId: (id) => console.log(id),
+    // tslint:disable-next-line
+    setMovingSectionId: (id) => console.log(id)
+  }
+};
+
+const UserInterfaceContext = React.createContext<IUIContext>(defaultContext);
+
+// 1: We define a context to be consumed from by subcomponents.
+// 2: We provide a hook for the wrapping component in api-container component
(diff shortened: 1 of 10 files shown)


lara-typescript/src/section-authoring/api/use-user-interface-context.ts
```diff
@@ -0,0 +1,61 @@
...
+    // tslint:disable-next-line
+    setMovingItemId: (id) => console.log(id),
+    // tslint:disable-next-line
+    setMovingSectionId: (id) => console.log(id)
+  }
+};
+
+const UserInterfaceContext = React.createContext<IUIContext>(defaultContext);
+
+// 1: We define a context to be consumed from by subcomponents.
```
kswenson, line 40:
```suggestion
// 1: We define a context to be consumed by subcomponents.
```

lara-typescript/src/section-authoring/api/use-user-interface-context.ts
```diff
@@ -0,0 +1,61 @@
...
+// 2: We provide a hook for the wrapping component in api-container component
+// 3: This api-container component provides UI features via context.
+// Sample from consuming component:
+// `const { userInterface, actions }  = React.useContext(UserInterfaceContext);`
```
kswenson, line 44:
```suggestion
// `const { userInterface, actions } = React.useContext(UserInterfaceContext);`
```

lara-typescript/src/section-authoring/api/use-user-interface-context.ts
```diff
@@ -0,0 +1,61 @@
...
+
+const useUserInterface = (): IUIContext => {
+  const [userInterface, setUserInterface] = useImmer<IUserInterface>(defaultUI);
```
kswenson, line 47: Ooh, cool -- `immer`, too!

lara-typescript/src/section-authoring/components/section-item-move-dialog.tsx
```diff
@@ -6,43 +6,35 @@ import { Close } from "../../shared/components/icons/close-icon";
 import { Move } from "../../shared/components/icons/move-icon";
 
 import "./section-item-move-dialog.scss";
+import { usePageAPI } from "../api/use-api-provider";
+import { UserInterfaceContext } from "../api/use-user-interface-context";
```
kswenson, line 10: In CLUE we've adopted the convention of keeping the style import(s) last.

PR #807, opened 2021-10-08
Move section item with menu part 2

knowuh:
Utilities and tests for traversing and moving sections / sectionItems

[#179720480] [#179720483]
https://www.pivotaltracker.com/story/show/179720480

### DIFF
--- a/app/controllers/api/v1/interactive_pages_controller.rb
+++ b/app/controllers/api/v1/interactive_pages_controller.rb
@@ -90,28 +90,22 @@ class Api::V1::InteractivePagesController < API::APIController
 
     section_id = section_params.delete('id')
     section = Section.find(section_id)
-    old_page_items = section.page_items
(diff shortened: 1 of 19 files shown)


lara-typescript/src/section-authoring/api/util/finding-utils.ts
```diff
@@ -0,0 +1,54 @@
...
+        return section;
+      }
+    }
+  }
+  return null;
+};
+
+export const findSectionByAddress = (pages: IPage[], address: ISectionAddress) => {
+  const { pageIndex, sectionIndex } = address;
+  if (pageIndex !== null && sectionIndex !== null) {
```
kswenson, line 27:
```suggestion
  if (pageIndex != null && sectionIndex != null) {
```
I generally prefer this form because it handles `undefined` as well, which seems like it should be handled similarly if it were ever to occur.

PR #809, opened 2021-10-13
Add and edit text block via item picker

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/179643499

[#179643499]


lara-typescript/src/section-authoring/api/lara-api-provider.ts
```diff
@@ -88,6 +92,19 @@ export const getLaraAuthoringAPI =
     return sendToLara({url: createPageItemUrl(args.pageId), method: "POST", body});
   };
 
+  const updatePageItem: APIPageItemUpdateF = (args: {pageId: string, sectionItem: ISectionItem}) => {
+    const pageItem = args.sectionItem;
+    const pageItemData = pageItem.data;
+    const translatedData: any = {};
+    Object.keys(pageItemData).forEach((key: string) => {
```
knowuh, line 99: OK, I was thinking that this entire itterator block that creates the new object would be the function extracted to a method we can reuse. Currently `cameToSnakeCase` just changes a string, but if we could have it rename all the keys in an object, that would be great.

lara-typescript/src/section-authoring/api/lara-api-provider.ts
```diff
@@ -96,20 +113,32 @@ export const getLaraAuthoringAPI =
 
   const getAllEmbeddables = () => {
     return sendToLara({url: libraryInteractivesUrl})
-      .then( (json: ILibraryInteractiveResponse) => {
+      // tslint:disable-next-line
+      .then( (json: any) => {
         const result = {
-          allEmbeddables: json.library_interactives.map(li => ({
+          allEmbeddables: json.library_interactives.map((li: any) => ({
```
knowuh, line 119: We should probably type this `li` variable.

lara-typescript/src/section-authoring/api/mock-api-provider.ts
```diff
@@ -137,23 +161,86 @@ const deletePageItem: APIPageItemDeleteF = (args: {pageId: PageId, pageItemId: I
...
   return Promise.resolve({allEmbeddables});
 };
 
 export const API: IAuthoringAPIProvider = {
   getPages, getPage, createPage, deletePage,
   createSection, updateSections, updateSection,
-  createPageItem, deletePageItem,
-  getAllEmbeddables
+  createPageItem, updatePageItem, deletePageItem,
+  getAllEmbeddables, pathToTinyMCE: null, pathToTinyMCECSS: undefined
```
knowuh, line 245: I wonder what we could do for the mock-api-provider as far as loading TinyMCE, is there a CDN URL we could use?  We don't have to do anything now if its hard, just wondering if there is something easy to do here.

lara-typescript/src/section-authoring/components/section-item-move-dialog.tsx
```diff
@@ -95,7 +95,8 @@ export const SectionItemMoveDialog: React.FC<ISectionItemMoveDialogProps> = ({
 
     return itemsList?.map((i) => (
       <option key={i.id} value={`${i.id}`}>
-        {i.title}
+        {// i.title
```
knowuh, line 98: I am not sure about this.  Should we remove the comment?

lara-typescript/src/section-authoring/api/mock-api-provider.ts
emcelroy: Good idea. I added a CDN URL in the latest commit. I'm not sure if that may need to be changed frequently, but I doubt it. And I think we'll eventually drop support for TinyMCE anyway, so it may not be an issue for long.

PR #810, opened 2021-10-13
179922672-add-section-related-indexes

knowuh:
Adds indexes on page_items and sections.

Removes an unused index.


knowuh:
@scytacki -- just a quick look if you could.  Adding indexes to speed up React queries.

PR #814, opened 2021-10-19
Switching section layouts post items added

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/178595258

[#178595258]

### DIFF
--- a/lara-typescript/src/section-authoring/util/change-layout-utils.spec.ts
+++ b/lara-typescript/src/section-authoring/util/change-layout-utils.spec.ts
@@ -23,13 +23,13 @@ const setUpFullWidthPage = () => {
             id: "item-3",
             column: SectionColumns.PRIMARY,
             position: 3
-          }          
+          }
         ]
       }
     ]
-  }
+  };
   return page;
-}
+};
 
 const setUp4060Page = () => {
   const page = {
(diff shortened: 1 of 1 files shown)


lara-typescript/src/section-authoring/api/use-api-provider.ts
```diff
@@ -187,11 +220,15 @@ export const usePageAPI = () => {
     let sectionItemsCount = 0;
     updatedItems?.forEach((i, index) => {
       sectionItemsCount++;
-      if (index > newIndex) {
-        updatedItems[index].position = sectionItemsCount;
-      }
+      const itemPosition = updatedItems[index].position;
```
knowuh, line 223: These changes to moveSection are going to conflict with a PR I have up.  I would revert them for now. 

The move function was extracted into a utility method that has a bunch of tests. We can add a test for the ordering of items within columns.  See: lara-typescript/src/section-authoring/util/move-utils.spec.ts

I would encourage you to extract the change-layout function in a similar way, and create a test suite for the different expected switch behavior documented in the PT story.  I think it will make it easier to debug going forward, and we can modify the tests if the design changes.

knowuh (approved):
Looks good, thanks

PR #817, opened 2021-10-21
Added external report url to initInteractiveMessage

dougmartin:
This is the first of 2 PRS - the first adds the external report url to the init interactive message.  The second will update the version of the interactive-api-host package along with its use in the main package.json.


scytacki (approved):
This looks fine to me.

PR #823, opened 2021-11-01
180150052 - Remove rails-assets gem source.

knowuh:
- add JS libraries directly to app/assets
- remove gem source
- tested locally with drawing tool

[#180150052]
https://www.pivotaltracker.com/story/show/180150052


knowuh:
This wasn't as simple as I first expected.  Dependencies had to be tracked down, and we had to add stylesheets for drawing-tool too.  But I tested the drawing tool and it seemed to work.

scytacki (approved):
This seems fine to me. 
We only need to support the LARA runtime for a few more months, so we should be able to remove all of this code from LARA soon.

tejal-shah (approved):

PR #824, opened 2021-11-01
Preview library interactives in authoring.

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/179745641

[#179745641]


lara-typescript/src/section-authoring/components/managed-interactive-preview.tsx
```diff
@@ -0,0 +1,51 @@
...
+}
+
+export const ManagedInteractivePreview: React.FC<IManagedInteractivePreviewProps> = ({
+  pageItem
+  }: IManagedInteractivePreviewProps) => {
+  const {
+    aspectRatio, aspectRatioMethod, authoredState, customAspectRatioMethod,
+    id, interactiveItemId, isFullWidth, name, libraryInteractiveId,
+    linkedInteractives, urlFragment
+   } = pageItem.data as any;
```
knowuh, line 20: It would b good to add a little typing here if we could.

This `data` thing is probably typed somewhere already, if there is some problem with type field overlap, we could use a `Partial<IDataType>` in which all the fields are considered optional, that would still be an improvement.

knowuh (changes requested):
This looks great.

I am wondering if we can do a better job typing`pageItem.data` it should be one of a few kinds of embeddable authoring data types. It should already be defined somewhere else in the code I would think.  LMK if this is a pain, I wouldn't spend more than an hour trying to work it out.

emcelroy:
After looking into Noah's suggestion about adding more typing, I found it was trickier than expected. So we decided to punt on that for now.

PR #828, opened 2021-11-03
179594368 copy page item

knowuh:
- React component API additions (storybook / rails api)
- Rails routes, controllers and tests


emcelroy (approved):
This looks good, thanks. I see several lines of the `duplicate` method in app/models/page_item.rb that aren't covered by tests, but that doesn't need to hold this back if coverage would be difficult to add?

emcelroy (approved):
Looks good!

PR #829, opened 2021-11-04
Add and edit iframe interactives via Item Picker

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/179594556

[#179594556]


lara-typescript/src/page-item-authoring/mw-interactives/customize.tsx
```diff
@@ -40,34 +59,73 @@ export const CustomizeMWInteractive: React.FC<Props> = (props) => {
...
+    native_width,
+  });
+
+  useEffect(() => {
+    setState({
+      ...state,
+      aspect_ratio_method: aspectRatioValues.mode,
+      native_height: aspectRatioValues.height,
+      native_width: aspectRatioValues.width
+    });
```
kswenson, line 80:
```suggestion
    setState(state => ({
      ...state,
      aspect_ratio_method: aspectRatioValues.mode,
      native_height: aspectRatioValues.height,
      native_width: aspectRatioValues.width
    }));
```
Any time the new state depends on the previous state (like here and several places below) you should use the functional form of `setState()`, otherwise you run the risk of getting stale values.

emcelroy: Thanks, @kswenson 👍 That probably explains why one aspect of the edit form wasn't quite working right. We've decided to approach handling the form data differently anyway, but that's definitely still good to know generally.

lara-typescript/src/page-item-authoring/mw-interactives/customize.tsx
```diff
@@ -40,34 +41,67 @@ export const CustomizeMWInteractive: React.FC<Props> = (props) => {
...
+    setState(currentState => ({
+      ...currentState,
+      aspect_ratio_method: aspectRatioValues.mode,
+      native_height: aspectRatioValues.height,
+      native_width: aspectRatioValues.width
+    }));
+  }, [aspectRatioValues]);
+
+  const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setState({
```
knowuh, line 63: I think Kirk's statement from 10 days ago is pertinent here.  We should probably use the functional form to update the state:

``` 
 setState(state => ({ ...state, ... etc })
```

I wasn't aware of this issue myself, so I am going to have to do a small code audit on the work I have been doing.

Here is a blog post that explains more: https://blog.logrocket.com/a-guide-to-usestate-in-react-ecb9952e406c/
> If you use the previous value to update state, you must pass a function that receives the previous value and returns the new value:

lara-typescript/src/page-item-authoring/mw-interactives/customize.tsx
```diff
@@ -40,34 +41,67 @@ export const CustomizeMWInteractive: React.FC<Props> = (props) => {
...
+      ...state,
+      [e.target.name]: e.target.value,
+    });
+  };
+
+  const handleCheckboxChange = (name: string, checked: boolean) => {
+    setState({
+      ...state,
```
knowuh, line 71: as above

knowuh (changes requested):
This looks great to me @emcelroy .

There were two calls to `setState` that probably should use the functional invocation where the previous state is the first argument.

PR #831, opened 2021-11-05
Integration: Switching section layouts (post-items added)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/179594634

[#179594634]

The only change here is to ensure that items in the primary column are placed first when changing from a multi-column layout to full-width (single column). Otherwise, the layout switching functionality works as expected.


knowuh (approved):

knowuh:
Looks good. Thank you.

PR #832, opened 2021-11-08
Allow read of other questions attachments [#179906617]

dougmartin:
- Added interative.questionId to initInteractive
- Added questionId to lara client readAttachment()

TODO:

- update to final package versions


scytacki (changes requested):
Overall looks good. 

My requested change is to remove the `questionId` field and update the `id` field to be a string with a "typed" value. This is part of the plan I created a while ago here: https://docs.google.com/document/d/1vhRfcrxPRMtJRkHDqYbDPXKPDCqZV5JjjKoSmqB5zMg/edit#

dougmartin:
@scytacki I've made your requested changes here: https://github.com/concord-consortium/lara/pull/832/commits/962a28017a68d4985dc54b358a70c4d079562414

I also made the new attachment option interfaces public as I think I'll need those in the QR code image spike.

I've also updated the CFM and AP PRs to match this and tested them locally.

I'm going to set this as ready for review but it is still a WIP in that I need to set the final package versions for the two exported modules.

scytacki (approved):
looks great to me
👍

PR #834, opened 2021-11-12
179376124 Page Navigation and Copy Part 3

knowuh:
This PR fixes several bugs with the initial Navigation and Page copy work.

- Fixes bugs with Move Item when specifying a page
- Fixes bugs with Move Section when specifying a page
- Consolidates duplicate concerns of moving things 
- Copy Page will now honor destination (where to copy to)

### DIFF
--- a/app/controllers/api/v1/interactive_pages_controller.rb
+++ b/app/controllers/api/v1/interactive_pages_controller.rb
@@ -76,7 +76,7 @@ class Api::V1::InteractivePagesController < API::APIController
     # Remove deleted sections:
     old_sections.each do |section|
       unless (new_section_ids.include?(section.id.to_s))
-        section.delete
+        section.update_attribute(:interactive_page_id, nil)
       end
     end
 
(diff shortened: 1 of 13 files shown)


knowuh (commented):
I can't delete my comment.

lara-typescript/src/section-authoring/api/api-types.ts
```diff
@@ -160,14 +160,19 @@ export interface IPage {
    */
   sections: ISection[];
 
+  /**
+   * Sections on this page:
```
emcelroy, line 164: Should this comment be changed?

app/controllers/api/v1/interactive_pages_controller.rb
```diff
@@ -76,7 +86,7 @@ def set_sections
     # Remove deleted sections:
     old_sections.each do |section|
       unless (new_section_ids.include?(section.id.to_s))
-        section.delete
+        section.update_attribute(:interactive_page_id, nil)
```
emcelroy, line 89: Sorry but I don't quite understand this change. I understand that it's setting the section's interactive_page_id value to nil and that that must negate its association with the page. But I don't understand why that's an improvement over `section.delete`. Do the deleted sections need to kept around for some reason even though they're no longer associated with the page? Is this related to the handling of moving a section between pages? I'm guessing so, but it's not quite clear to me.

emcelroy (approved):
I think this looks good. There's one part I had a question about, but I'm probably just not fully understanding what's being done rather than it being a problem.

app/controllers/api/v1/interactive_pages_controller.rb
knowuh: When we 'move' a section, we just remove its association from the page where it is now, and add a new association to its destination. We update each page separately. 

Before this change, we would lose sections when moving them in LARA sometimes.

knowuh:
Thanks @emcelroy

PR #836, opened 2021-11-16
180250615 pages have unique urls

knowuh:
When editing pages in the new-sections React SPA, update the location to reflect the page currently being edited.

Also: When launching the editor from LARA start on the correct page based on the url of the edit link clicked from the activity page.

**Note**  Publication status polling was disabled in this PR too. It's not clear whether we are going to need portal publication. For the moment I have taken it out completely because it was slowing things down and cluttering up the console log.

One other unrelated fix: The wrong interface was being used for `QueryBoundPage` in `renderAuthoringPage`.

### DIFF
--- a/app/views/interactive_pages/edit.html.haml
+++ b/app/views/interactive_pages/edit.html.haml
@@ -62,7 +62,11 @@
     });
 
     $(document).ready(function() {
-      var params = {host: window.location.origin, activityId: "#{@activity.id}"};
+      var params = {
+        host: window.location.origin,
+        activityId: "#{@activity.id}",
+        id: "#{@page.id}"
+      };
       var element;
       console.log(params);
       var container = $("#sections-container")[0];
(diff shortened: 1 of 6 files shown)


knowuh:
@emcelroy a smaller PR finally.

emcelroy (changes requested):
This looks good to me, but when I test it out locally it doesn't quite work right. 

The problem is when I click on a specific page's edit link from an activity's "home page" I go to that specific page's edit view at first, but then after a second or so I'm automatically sent to the first page's edit view.

The same thing happens if I copy the URL for a specific page's edit view. If I paste that URL into a new tab, I'm taken to the correct page's edit view, but almost immediately it switches to the page one edit view.

knowuh:
I wonder if the javascript being served by rails didn't pick up the changes for some reason in your docker container?

Here is the same setup working for me locally:

PR #838, opened 2021-11-17
180351146 ping pong view changes

knowuh:
## Reduce visual ping-pong effect through fewer requests to the server.

We already explicitly invalidate the cache when we make changes. By default ReactQuery also
invalidates the cache whenever the page is focused or components re-mount. We would often request the list of pages several times in rapid succession.

### Some other general cleanup:
- Fix console error related to setting state in a render loop (page-provider -> user-interface-provider)
- remove unused var

PT Story: [🐞 180351146 ping pong view changes](https://www.pivotaltracker.com/story/show/180351146)

### DIFF
--- a/app/controllers/api/v1/interactive_pages_controller.rb
+++ b/app/controllers/api/v1/interactive_pages_controller.rb
@@ -37,7 +37,7 @@ class Api::V1::InteractivePagesController < API::APIController
 
     return error("Can't find activity for page") unless activity
 
-    position = params[:dest_index];
+    position = params[:dest_index]
 
     next_page = @interactive_page.duplicate
     next_page.lightweight_activity = activity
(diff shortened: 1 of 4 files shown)


app/controllers/api/v1/interactive_pages_controller.rb
```diff
@@ -37,7 +37,7 @@ def copy_page
 
     return error("Can't find activity for page") unless activity
 
-    position = params[:dest_index];
+    position = params[:dest_index]
```
emcelroy, line 40: Did you mean to delete this semicolon?

emcelroy (approved):
Looks good and definitely fixes the problem for me when working with the site in my local instance of LARA. Thanks! I just left one comment about a semicolon that may have been unintentionally jettisoned.

app/controllers/api/v1/interactive_pages_controller.rb
knowuh: Yeah, I was fixing some  Ruby ⟷ Javascript whiplash I was suffering with.  Ruby doesn't typically use semicolons like this.  :D

emcelroy: Ah, right! This is a Ruby file. Duh :)

PR #839, opened 2021-11-19
Header log in via home page

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/180171323

[#180171323]


app/controllers/api/v1/interactive_pages_controller.rb
```diff
@@ -278,6 +279,20 @@ def delete_page_item
     render_page_sections_json
   end
 
+  def get_portal_list
```
emcelroy, line 282: I tried adding test coverage for `get_portal_list` but just could not get it to work. I didn't want to hold this up any longer, so I'm going to skip it for now unless you think it's important, @knowuh?

In case it's useful, as far as I can tell, the main problem was that line 53 of spec/spec_helper.rb 

`add_fake_authorization('concord.portal')`

was ultimately causing this error when the test tried `get_portal_list`

`ActionController::RoutingError:
       No route matches {:controller=>"users/omniauth_callbacks", :action=>"passthru", :provider=>"portal"}`
       
Seems like that has to do with the call to `user_omniauth_authorize_path` in `get_portal_list`. I found that any value except "portal" passed to `add_fake_authorization` in spec/spec-helper.rb would cause the same error.

PR #840, opened 2021-11-19
179739155 section numbering

eireland:
Adds Section label to section authoring header.
Changes number shown in header to be position number from section id. Number updates to current position when sections are moved to current position.
Section move dialog shows position number in the dropdown instead of section ids. Fixes a bug where section list does not update when a section is added or deleted.
Item move dialog shows section position number in the section dropdown to move item to.

### DIFF
--- a/app/controllers/api/v1/interactive_pages_controller.rb
+++ b/app/controllers/api/v1/interactive_pages_controller.rb
@@ -240,7 +240,7 @@ class Api::V1::InteractivePagesController < API::APIController
         if linked_interactives.present?
           page_item.set_linked_interactives(JSON.parse(linked_interactives))
         end
-      end 
+      end
       if embeddable
         embeddable.update_attributes(data)
       end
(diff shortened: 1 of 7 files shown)


knowuh (approved):
Looks great, thank you!

emcelroy (approved):
Looks good 👍

One little issue I noticed when testing in a browser and which very possibly isn't related to these changes is that the Page option in the section move dialog gets stuck on the option you last selected. So if you move a section from Page 1 to Page 2 and then you navigate to Page 3 and go to move a section from there to Page 1, the section move dialog has Page 2 selected by default when it first opens.

Unless you think that's something related to these changes and/or easily fixed, I can make a separate bug story for that in PT. Just let me know. Thanks.

knowuh:
@emcelroy I am going to merge this.  Do you have  a separate PT story for the bug you mentioned?

emcelroy:
@knowuh Evangeline was going to look into fixing it and said she'd create a PT story for it if there wasn't a straightforward fix.

eireland:
Bug story for preselected page in the move dialog here https://www.pivotaltracker.com/story/show/180404389
Also found another bug story re move dialog box not showing the correct section to move when page is changed https://www.pivotaltracker.com/story/show/180404461

PR #842, opened 2021-11-22
Header: Add logged in username and icon

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/180170000

[#180170000]


knowuh:
This looks good @emcelroy  I am not sure how / when we are going to get first and last names into LARA, but that is a problem for a different day I guess.

👍

knowuh (approved):
Looks great, thank you.

PR #843, opened 2021-11-22
180087466 -- Add preview links to activity player

knowuh:
## 180087466 -- Add preview links to activity player

[PT Story 180087466](https://www.pivotaltracker.com/story/show/180087466)

### NOTES: 
- In order to test this completely, it will have to be deployed to https://lara-demo.concordqa.org/
- Any testing LARA host will also have to set an `.env` var: `ACTIVITY_PLAYER_URL=https://activity-player.concord.org/branch/new-sections/` 
- In development mode (and story book) the links will be populated, but the report service won't likely be able to talk with the developers LARA instance. 
- There is a small bug in storybook that the first page is not set, and so the preview links wont work until some navigation has happened. This should be a simple thing to fix.
- More design work and CSS work is needed.
- Rails controller test needed.

### DIFF
--- a/lara-typescript/src/section-authoring/components/preview-links.tsx
+++ b/lara-typescript/src/section-authoring/components/preview-links.tsx
@@ -1,0 +1,61 @@
+import * as React from "react";
+import { useState } from "react";
+import {ExternalLink } from "../../shared/components/icons/external-link-icon";
+
+import "./preview-links.scss";
+
+export interface IPreviewLinksProps {
+  previewLinks: Record<string, string> | null;
+}
+
+export const PreviewLinks: React.FC<IPreviewLinksProps> =
+  (props: IPreviewLinksProps) => {
+
+  const { previewLinks } = props;
+  const [ previewLink, setPreviewLink] = useState<string|false>(false);
+
+  const handleLinkSelect = (evt: React.ChangeEvent<HTMLSelectElement>) => {
+    const value = evt.currentTarget.value;
+    if (value && value.length > 0) {
+      setPreviewLink(value);
+    }
+    else {
+      setPreviewLink(false);
+    }
+  };
+
+  const renderPreviewOptions = () => {
+    if (previewLinks) {
+      return (
+        <span className="links-list" >
+          <select onChange={handleLinkSelect}>
+            { Object.keys(previewLinks).map(label => {
+              return (
+                  <option key={label} value={previewLinks[label]} >
+                    {label}
+                  </option>
+                );
+              })
+            }
+          </select>
+          { previewLink
+            ? <a href={previewLink} target="_blank"><ExternalLink/></a>
+            : <span className="disabled"><ExternalLink/></span>
+          }
+        </span>
+      );
+    }
+    return (
+        <span className="not-available">No preview link available</span>
+    );
+  };
+
+  return (
+    <>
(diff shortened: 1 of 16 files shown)


knowuh:
@emcelroy -- I was hoping to get your eyes on this, and then split up some of the remaining work tomorrow morning.

Just two items I think:

- The styling could use some of your attention.
- I need to write a test for the controller.

emcelroy:
@knowuh I have to step away from my computer for a while, but will look at this more carefully later this evening. FWIW: My first impression is that it looks pretty good.

emcelroy:
@knowuh I tested this out in my local LARA and it works pretty well. There are two issues I noticed which I'm not sure fall under styling.

1) The menu only appears on an activity's edit "home page". It doesn't appear when you edit an individual page.

2) There's no launch button to click to trigger the opening of the preview in a new window. Is that not one of the acceptance criteria anymore?

knowuh:
Hmn, @emcelroy that doesn't sound right. There should be a SVG link for opening the preview in a new window as per the mock, and the menu shows on every page. I will double check to make sure everything is in this commit, and that things didn't get clobbered by a merge. I will be in touch after I investigate.

knowuh:
Hello @emcelroy - I just verified that what is in this PR is what is in my local checkout.  Here is what I am seeing:



Maybe we could meet about this and see what is happening.

emcelroy:
@knowuh Sorry, looks like this is at least partially due to my local LARA not fully updating when I switch branches. I should've known to check that first. I'm still seeing something not quite right, though.

I just did a git pull to make sure I had the latest changes, brought my local LARA down and back up, and made sure to rebuild lara-typescript.js. Now when I edit an activity, I do see the select menu on an activity's individual pages. I also see the button icon there. However, that button still doesn't appear on the activity's edit home page. The select menu is there, but there is no button.

lara-typescript/src/section-authoring/components/preview-links.tsx
```diff
@@ -0,0 +1,61 @@
...
+    );
+  };
+
+  return (
+    <>
+      <div className="preview-links">
+        <span>Preview in:</span>
+        {renderPreviewOptions()}
+      </div>
+    </>
```
kswenson, line 59:
```suggestion
    <div className="preview-links">
      <span>Preview in:</span>
      {renderPreviewOptions()}
    </div>
```
I don't think the fragment is useful in this case.

knowuh:
> activity's edit home page … The select menu is there, but there is no button.

@emcelroy I have not been changing anything about the activity editing page until now. It sounds like you think that is part of this story, so I will take a stab at that.

emcelroy:
Thanks, @knowuh. This is looking good. The preview menu now looks as expected on the activity editing page. Sorry I didn't think about the fact that that was separate from the preview menu on the other pages. I do think it's best that they match, though.

You mentioned that the styling could use some attention. There are some small tweaks I'd like to make. What would be the best process for that?

knowuh:
@emcelroy:  Maybe we can merge this so the PR is manageable.

At that point you could open another branch from new-sections that cleans things up.
I will be working on the Rails side in another branch to send the data off to AP in the correct format.

emcelroy (approved):
Looks good, thanks 👍

PR #844, opened 2021-11-22
Header: Click CC logo to get to homepage

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/180171535

[#180171535]


knowuh (approved):
This looks good to me. 

I wasn't aware of this accessibility pattern, and it makes sense to me. 

I am not unqualified to say whether or not this is the right pattern to use here, but I have seen examples of adding an aria role of `role="button"` as well, but it would require more research.

This is a good reminder for me. I always assume screen readers will just do the right thing.

PR #845, opened 2021-11-23
Fixes bug that shows undefined section in section move dialog header

eireland:


knowuh (approved):
Looks good to me, thanks

PR #846, opened 2021-11-23
Redirects home button to activity settings page in authoring

eireland:
When user clicks on the Home button in the page nav, they will be shown the activity settings page.


knowuh (approved):
Looks good to me, thanks

PR #847, opened 2021-11-23
Makes page nav behave similar to Activity Player

eireland:
Limits the number of pages shown in page nav to 11 and scroll the numbers if there are more pages.
Removes the outline around nav arrows when disabled.
Still has the problem of home button not highlighting when user navigates to home button via previous arrow.


emcelroy (approved):
Looks great--thanks!

knowuh (approved):
Looks good, thank you.

PR #850, opened 2021-11-24
Header: Move activity name to top and restyle page title

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/180171557

[#180171557]


knowuh (approved):
looks good, thanks!

PR #853, opened 2021-11-30
Fixed teacher preview report [#180446625]

dougmartin:
Removed check for user in run when generating Report Service report url.


dougmartin:
Deleted as fix not needed.

PR #854, opened 2021-11-30
Header: Page Nav Styling

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/180250519

[#180250519]

Most of the nav menu styling work has already been done elsewhere. The only change here is a fix for completion page buttons which I seem to have not set up quite right originally.


knowuh (approved):
Looks good, thank you.

PR #855, opened 2021-11-30
Add test for get_portal_list endpoint.

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/180398652

[#180398652]

### DIFF
--- a/spec/controllers/api/v1/interactive_pages_controller_spec.rb
+++ b/spec/controllers/api/v1/interactive_pages_controller_spec.rb
@@ -232,80 +232,80 @@ describe Api::V1::InteractivePagesController do
   end
 
   describe "#get_library_interactives_list" do
-  let (:managed_interactive1) { FactoryGirl.create(:managed_interactive,
-    :library_interactive_id => library_interactive1.id
-  )}
+    let (:managed_interactive1) { FactoryGirl.create(:managed_interactive,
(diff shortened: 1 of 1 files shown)


spec/controllers/api/v1/interactive_pages_controller_spec.rb
```diff
@@ -734,5 +734,30 @@ def add_interactive_to_section(page, interactive, section)
...
+              path: "/auth/foo"
+            },
+            {
+              name: "Concord.Portal",
+              path: "/auth/foo"
+            },
+            {
+              name: "Portal",
+              path: "/auth/foo"
+            }
```
emcelroy, line 756: This duplication of the data in lines 745-748 is partially the result of [line 54 in spec/spec_helper.rb](https://github.com/concord-consortium/lara/blob/79ec90befaaf1d4b2c6ddc190bfefdbeaebb398d/spec/spec_helper.rb#L54). I'm not sure, but it seems like that line should be changed to use something other than "portal" as the strategy name value. There's already a portal named Portal which seems to exist by default in the testing environment. 

I tried both removing line 54 in spec/spec_helper.rb and then changing the strategy name instead, but both options caused other problems without a simple solution. So I've left it as is even though it makes the response here look a little odd.

knowuh (approved):
Thanks, it looks good to me.

PR #857, opened 2021-11-30
180087466 preview activities in activity player (3)

knowuh:
## Enable previews in AP:

For [This PT story](https://www.pivotaltracker.com/story/show/180087466)

## Changes:
- Modifies the export format so that the AP player can read it.
- Changes the layout names for the sections to match what the AP expects.
- Updated the export test expectations

NB: 
- We will need to update all the sections on the lara-demo server to use the new section identifiers.
- We will need to change the Activity Player URL used on lara-demo

[#180087466]

### DIFF
--- a/app/models/section.rb
+++ b/app/models/section.rb
@@ -11,12 +11,12 @@ class Section < ActiveRecord::Base
   INTERACTIVE_BOX = 'interactive_box'
   ASSESSMENT_BLOCK = "assessment_block"
   DEFAULT_SECTION_TITLE = ASSESSMENT_BLOCK
-  LAYOUT_FULL_WIDTH="Full Width"
+  LAYOUT_FULL_WIDTH="full-width"
   LAYOUT_60_40="60-40"
   LAYOUT_40_60="40-60"
   LAYOUT_70_30="70-30"
-  LAYOUT_30_70="30_70"
-  LAYOUT_RESPONSIVE="Responsive"
+  LAYOUT_30_70="30-70"
+  LAYOUT_RESPONSIVE="responsive"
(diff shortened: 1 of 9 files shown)


spec/models/interactive_page_spec.rb
```diff
@@ -201,7 +201,8 @@ def find_labbook(i_page)
   describe '#export' do
     it 'returns json of an interactive page' do
       page_json = page.export.as_json
-      expect(page_json['embeddables'].length).to eq(page.embeddables.count)
+      puts page_json
```
emcelroy, line 204: Should this be removed?

app/models/section.rb
```diff
@@ -60,6 +60,23 @@ def to_hash
...
+    helper = LaraSerializationHelper.new
+    {
+      title: title,
+      is_hidden: !show,
+      layout: layout,
+      secondary_column_collapsible: can_collapse_small,
+      secondary_column_display_mode: "stacked", # TODO: FIXME
+      # embeddables: embeddables.map do |embed|
+      #   helper.export(embed)
+      # end
```
emcelroy, line 73: Should this be removed?

emcelroy (approved):
Looks good to me 👍 I just noted a couple places where there is what looks like unnecessary code.

knowuh:
Thanks @emcelroy I cleaned things up in https://github.com/concord-consortium/lara/pull/857/commits/82d8bcaa9778922fa8ca9dc6ad23a7dfc982c051 -- I also did some safety checking on the layout properties (I was hitting an error when the value for a section layout were something we didn't spec out)

PR #858, opened 2021-11-30
Centers the section header elements

eireland:
Centers the section header elements

### DIFF
--- a/lara-typescript/src/section-authoring/components/authoring-section.scss
+++ b/lara-typescript/src/section-authoring/components/authoring-section.scss
@@ -140,14 +140,14 @@ select {
 }
 
 .activity-name {
-  color: var(--orange); 
+  color: var(--orange);
   font-size: 18pt;
-  grid-column: 2/8; 
+  grid-column: 2/8;
   justify-self: start;
 }
 
 .page-name {
-  color: black;  
+  color: black;
   font-size: 18pt;
   grid-column: 2/8;
   justify-self: start;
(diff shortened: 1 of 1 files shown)


knowuh (approved):
Well, this change is tiny.

I am not following the conversation in slack closely enough to know if this issue is resolved or not.

But the code change here is minimal and controversial so I am approving :D

emcelroy (changes requested):
Sorry but this doesn't look right to me. The elements aren't quite centered ~~and they're still a little out of alignment with each other~~. Here's a screenshot I took in Chrome at 100% zoom that I added some guidelines to.

image

I do think something will need to be done about the padding for `.sectionMenu`. Even though it's set to have the same amount of padding all around, there's less space at the bottom than at the top if you inspect it carefully.

image

I think increasing the bottom padding by one pixel will help even though that seems counterintuitive. I don't see any other option for correcting the inconsistent spacing.
 
I also think `.sectionDragHandle` needs less top margin, more like 2px instead of 5px.
 
If what you're seeing is different than what appears in the screenshots above, maybe we should share screens in real time to see if we can figure out the discrepancy?

emcelroy:
Looking at this again I think I was a _little_ wrong in that the elements _are_ actually fairly well aligned with each other. The different font sizes for the text threw me I think.

I still think they're not properly centered in the container, though. It may be that adding that pixel of padding to the bottom of the container is all that's needed.

emcelroy (changes requested):
Definitely better but there's still one tiny issue that I think can be easily fixed. Sorry. 

The "Section" H3 is ever so slightly out of line with the "Layout" label text next to it.

image

If you add `line-height: 1` to the H3's style rules, that brings it up inline.

image

So I think if you make that one change, this will be done.

emcelroy (approved):
Looks good -- thank you!

PR #859, opened 2021-12-01
Show/hide secondary column setting

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/180171403

[#180171403]


knowuh (approved):
Looks good, and tested locally. 

Thanks

PR #861, opened 2021-12-03
179597784 completion page

eireland:
When user sets page to be a completion page in page settings:
- Shows an authoring Completion Page (using existing authoring Completion Page text) 
- Moves that page to the end of the pages array so that it is now the last page of the activity.
- When a new page is added, and there is a completion page, the new page is added to the page prior to the completion page
Note that if the author changes the setting so it is no longer the completion page, that page remains at the end.

### DIFF
--- a/app/controllers/api/v1/interactive_pages_controller.rb
+++ b/app/controllers/api/v1/interactive_pages_controller.rb
@@ -91,8 +91,11 @@ class Api::V1::InteractivePagesController < API::APIController
         name: page_params['name'],
         is_completion: page_params['isCompletion'],
         is_hidden: page_params['isHidden'],
-        position: page_params['position']
       })
+
+      if page_params['isCompletion']
+        @interactive_page.move_to_bottom
+      end
     end
     @interactive_page.save!
     pages = activity.reload.pages.map do |page|
--- a/app/controllers/api/v1/interactive_pages_controller.rb
+++ b/app/controllers/api/v1/interactive_pages_controller.rb
@@ -59,7 +59,15 @@ class Api::V1::InteractivePagesController < API::APIController
     authorize! :update, activity
     return error("Can't find activity #{params[:activity_id]}") unless activity
     activity.pages.create()
+    last_idx = activity.pages.length - 1
+    prev_last_idx = activity.pages.length - 2
+    if activity.pages[prev_last_idx].is_completion
+      activity.pages[prev_last_idx].position = last_idx
+      activity.pages[prev_last_idx].move_lower
+    end
+
     pages = activity.reload.pages.map do |page|
+      page.reload
       generate_page_json page
     end
     render :json => pages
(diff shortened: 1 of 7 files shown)


emcelroy (commented):
I think this looks good and it works perfectly well when I test it locally. There are two things I'm not sure of, though, and think we should wait till @knowuh takes a look before merging.

I'll make separate comments inline with the parts I'm not sure about.

app/controllers/api/v1/interactive_pages_controller.rb
```diff
@@ -59,7 +59,15 @@ def create_page
     authorize! :update, activity
     return error("Can't find activity #{params[:activity_id]}") unless activity
     activity.pages.create()
+    last_idx = activity.pages.length - 1
+    prev_last_idx = activity.pages.length - 2
+    if activity.pages[prev_last_idx].is_completion
+      activity.pages[prev_last_idx].position = last_idx
+      activity.pages[prev_last_idx].move_lower
```
emcelroy, line 66: As I understand it `move_lower` was deprecated as of Rails 2. I'm not sure it matters, but it may be better to find a different way to do this.

lara-typescript/src/section-authoring/api/lara-api-provider.ts
```diff
@@ -90,14 +90,27 @@ export const getLaraAuthoringAPI =
...
-    return sendToLara({url: updatePageUrl(pageId), method: "PUT", body: data});
+    const pages = await getPages();
+    const indx = pages.findIndex(p => p.id === pageId);
+    if (indx > -1) {
+      if (indx < pages.length - 1 && changes.isCompletion) {
+        pages.push(pages.splice(indx, 1)[0]); // put the page at the end of the pages list
+      }
+      const newIndx = pages.length;
+      const nextPage = {... pages[newIndx], ...changes };
+      pages[newIndx] = nextPage;
```
emcelroy, line 103: I'm not sure this matters, either, but might it be better to have the Rails side handle all of this?

app/controllers/api/v1/interactive_pages_controller.rb
eireland: I got it from here? https://github.com/brendon/acts_as_list

emcelroy: Hm. Maybe it's not the same thing I was looking at then which is [this](https://apidock.com/rails/ActiveRecord/Acts/List/InstanceMethods/move_lower). I see you're discussing acts_as_list with Noah in Slack, so I'll defer to the outcome of that conversation about this.

app/controllers/api/v1/interactive_pages_controller.rb
```diff
@@ -82,8 +90,12 @@ def update_page
       @interactive_page.update_attributes({
         name: page_params['name'],
         is_completion: page_params['isCompletion'],
-        is_hidden: page_params['isHidden']
+        is_hidden: page_params['isHidden'],
```
emcelroy, line 93: Seems like the comma at the end of this line isn't necessary. Should it be removed?

emcelroy (approved):
I left a few comments about possibly cleaning up some minor bits of unused or unnecessary code, but otherwise this looks good and works as expected when tested locally. Thanks.

PR #862, opened 2021-12-06
Clean up "Edit Activity" form/page

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/180415940

[#180415940]

These changes include the addition of a Rails/Haml clone of the section-authoring activity page nav menu to the Rails views templates. I ran into issues with alternative options that I wouldn't have been able to work out quickly enough.


app/assets/stylesheets/publication.scss
```diff
@@ -1,6 +1,6 @@
 #modal {
   .publication {
-    padding: 0px;
+    padding: 0;
```
knowuh, line 3: I noticed you changed this from `0px` to `0` 

I am just wondering why for my own education.

app/views/lightweight_activities/edit.html.haml
```diff
@@ -18,77 +18,79 @@
...
-      .field
-        = f.label :time_to_complete,  'Estimated time to complete, in minutes'
-        = f.text_field :time_to_complete
       .field
         = f.label :thumbnail_url, "Thumbnail URL"
         = f.text_field :thumbnail_url, :class=>"thumbnail_source"
-        %br
         .thumbnail_info.hint
-          Your thumbnail image should be 300px x 250px
+          Thumbnail image should be 1460 pixels wide by 800 pixels high.
```
knowuh, line 65: random observation: maybe we should change the name from `thumbnail` if this is the preferred size.

app/views/lightweight_activities/new.html.haml
```diff
@@ -13,43 +13,60 @@
...
-    = field_set_tag "Authoring Mode" do
-      = f.select :editor_mode, options_for_select(LightweightActivity::EDITOR_MODE_OPTIONS, @activity.editor_mode)
-  = field_set_tag 'Notes' do
-    The notes area is for storing information for other authors, e.g. source material or usage restrictions, and is not intended to be displayed in the runtime.
-    = f.text_area :notes
+%h1 New Activity
+.sequence_form
+  = form_for @activity, :url => '/activities' do |f|
+    = f.error_messages
+    =# f.hidden_field 'investigation_id', :value =>@activity.investigation.id # LightweightActivities currently don't belong_to Investigations
```
knowuh, line 20: We could remove the above line that is commented out, obviously this isn't your edit, just legacy stuff in here.

knowuh (approved):
This looks fantastic, thank you.

I made comments but you can safely ignore all of them. Nothing to do.  👍

app/assets/stylesheets/publication.scss
emcelroy: @knowuh That's a code style habit I picked up a long time ago. I forget where exactly, but I read it's best to leave off units when a value is zero in CSS. Since you could be using any one of several units, it could be zero _anythings_, so it doesn't make sense to specify a particular unit type. It's not necessary and doesn't result in anything looking or behaving differently, but I think the idea makes good sense and it helps minimize the code.

PR #869, opened 2021-12-14
Fixed build after addition of the reportItem mode

dougmartin:
The addition of the reportItem mode caused a build break of the lara-typescript example interactives in the top-level switch statement.


scytacki (approved):
looks good to me

PR #870, opened 2021-12-15
Page Settings: Hidden Page

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/179598005

[#179598005]

- Adds an icon to indicate hidden pages in the authoring page nav menu. The icon is used in place of each hidden page's page number.
- When a page is hidden, page numbers for subsequent pages are changed accordingly in nav menu (e.g., if an activity's second page is hidden, its third page will be listed as "2" in the page nav menu, the fourth page as "3", etc.).
- Adds an alert banner to the top of hidden pages in authoring view.
- Completion pages cannot be hidden.


app/views/lightweight_activities/_edit_header.html.haml
```diff
@@ -4,9 +4,18 @@
...
     %button{ :id => "page-nav-home", :class => "page-button current", :onclick => "window.location='#{edit_activity_path(@activity)}'" }
       = svg_icon(asset_path "/assets/images/icons/page-nav/home.svg")
+    - hidden_pages_before = 0
     - @activity.pages.each.with_index(1) do |p, index|
-      %button{ :id => dom_id_for(p, :item), :class => "page-button #{p.is_completion ? 'completion-page-button' : ''}", :onclick => "window.location='#{edit_activity_page_path(@activity, p)}'" }
-        = p.is_completion ? svg_icon(asset_path "/assets/images/icons/page-nav/completion.svg") : index
+      - special_class = p.is_hidden ? 'hidden-page-button' : ''
+      - special_class = p.is_completion ? 'completion-page-button' : ''
+      %button{ :id => dom_id_for(p, :item), :class => "page-button #{special_class}", :onclick => "window.location='#{edit_activity_page_path(@activity, p)}'" }
+        - if p.is_hidden
```
knowuh, line 12: Not sure if we need to do this.  I think we could use this method too:
on the page model there is a `page_number` method...
https://github.com/concord-consortium/lara/blob/master/app/models/interactive_page.rb#L232

knowuh (approved):
This looks good to me Ethan.

I think we could have taken a short-cut and used a consistent page numbering scheme if we used the method called `page_number` on the page model, but I don't think its critical, and this is easy enough to understand.

app/views/lightweight_activities/_edit_header.html.haml
emcelroy: Oh, thanks. I'll use `page_number`. Seems like the best thing to do.

PR #871, opened 2021-12-18
Fixes page preview to show current page

eireland:
Preview links were being generated by the preview selector dropdown, and were getting updated correctly when pages were changed. But the actual preview link was not getting updated because it only gets updated if the dropdown selection is changed. This gets preview link from the current preview links list instead of waiting for dropdown selection to change.

### DIFF
--- a/lara-typescript/src/section-authoring/components/preview-links.tsx
+++ b/lara-typescript/src/section-authoring/components/preview-links.tsx
@@ -13,7 +13,7 @@ export const PreviewLinks: React.FC<IPreviewLinksProps> =
 
   const { previewLinks } = props;
   const [ previewLink, setPreviewLink ] = useState<string|false>(false);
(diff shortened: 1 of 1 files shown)


emcelroy (approved):
This generally looks good and fixes the bug when I test it locally. So I think it's good to go.

One little thing I'm not sure about, though, is the use of `previewer` for the variable name. Am I right in understanding that the value of `previewer` is the Activity Player page URL you'll be previewing? If so, would a name like `previewPageUrl` be more meaningful? It may just be me, but `previewer` sounds like something more substantial than a simple URL string. I was a little confused by the code at first and think that was part of it. I may be misunderstanding or unaware of a convention you're using, though.

emcelroy:
@eireland Thinking about what I said about `previewer` some more, I guess it doesn't quite hold water. Seems like the AP is the Previewer and the variable describes the full URL to activate the preview in the previewer... I think you can safely disregard my last comment.

PR #874, opened 2021-12-21
Item Setting for Half-Width Items in Full Width Sections

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/178595430

[#178595430]

There are two bugs I'm aware of that we're going to leave in for now. 

1) When you add a new item or edit an existing one to a section, any existing items that are set to be half width will switch to full width. If you reload the page or make another change, the items revert to half width.

2) The half width option does not get disabled for items in a section that doesn't have a full-width layout.


dougmartin (approved):
Looks good :+1:

PR #876, opened 2022-01-04
Page Settings: Hidden Page Fixes

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/179598005

[#179598005]

These changes address two issues identified by QA review of the original work done in [this PR](https://github.com/concord-consortium/lara/pull/870):

1) Hidden pages should be identified by adding "(hidden)" to their listing in the move dialog. Also, completion pages should not be included in the list of pages in the move dialog since completion pages do not display assessment items.

2) It should not be possible to set a page to be hidden and then also make it a completion page. Enabling the hidden option should automatically disable the completion page option, and enabling the completion page option should automatically disable the hidden option.


knowuh (approved):
nice, thank you

PR #877, opened 2022-01-04
180626537 support example report item

knowuh:
This is a small change to support report item interactives.

When we publish an activities structure we must also include the URL to the report-item-interactive for any interactives that have one. This is so the report can lookup the report-item-interactive when generating a report for it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,3 +61,4 @@ config/environments/local-development.rb
 # simplecov output
 coverage/
 .DS_Store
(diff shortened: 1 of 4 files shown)


dougmartin (approved):
Looks good :+1:

dougmartin (approved):
Looks even better :+1:

emcelroy:
Thanks for fixing that, Noah.

PR #879, opened 2022-01-04
180788915 copy page position

eireland:
The current implementation of the copy page uses `insert_at` from the `acts-as-list` library. This was causing the subsequent page positions to act strangely. 
ie. 
`
[{page: 1, position: 1}, {page: 2, position: 2}, {page: 3, position: 3}, {page: 4, position: 4}]`
Copy page 4 to after page 2. The result was: 
`[{page: 1, position: 1}, {page: 2, position: 2}, , {page: 4, position: 3}, {page: 3, position: 5}, {page: 4, position: 5}]
`
This was causing issues when trying to preview the correct page. When you try to preview the last page 4, sometimes you will get page 3.
There were also issue previewing items that were copied to the before page 1, as position would be set to -1, which would preview to the activity page setting. If another page is copied to after page 1, the position would be set to 0 which would also preview to activity page setting.
Changes this PR makes:
1. The method used for inserting the copied page to `set_list_position`. This inserts the copied page to the specified position and renumbers subsequent positions correctly.
2. The logic for figuring out what position to insert the page in. It now uses the position number instead of by page ids.

### DIFF
--- a/app/controllers/api/v1/interactive_pages_controller.rb
+++ b/app/controllers/api/v1/interactive_pages_controller.rb
@@ -45,10 +45,9 @@ class Api::V1::InteractivePagesController < API::APIController
     return error("Can't find activity for page") unless activity
 
     position = params[:dest_index]
-
     next_page = @interactive_page.duplicate
     next_page.lightweight_activity = activity
-    next_page.insert_at(position)
+    next_page.set_list_position(position)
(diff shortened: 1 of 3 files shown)


emcelroy (approved):
This looks great to me 👍

Works as expected when tested locally, too.

PR #881, opened 2022-01-05
Fix Report Item URL field is incorrectly styled and placed

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/180809859

[#180809859]


dougmartin (changes requested):
The indention is a little off from lines 127 to 139 and makes it hard to read - could you fix that?

emcelroy:
@dougmartin Thanks for catching that. I've fixed it in the latest commit.

PR #884, opened 2022-01-06
Adds a query param to the preview url so hidden pages can be previewed.

eireland:
This goes with work done on Activity Player. 
The preview link has an added query param `author-preview=true` when previewing a page.
On the AP side, if the query-param is true, then all pages will be rendered including hidden pages. But hidden pages will have a banner informing author that the page is hidden at runtime.


emcelroy (approved):
Looks good!

PR #886, opened 2022-01-06
Add Interactive ID information to Iframe interactive form

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/180760275

[#180760275]

These changes reinstate the Interactive ID listing at the top of the iFrame and Library Interactive edit forms.

I also snuck in some minor, unrelated style fixes for an issue introduced in a separate PR that's already been merged into new-sections.

### DIFF
--- a/lara-typescript/src/section-authoring/components/item-edit-dialog.tsx
+++ b/lara-typescript/src/section-authoring/components/item-edit-dialog.tsx
@@ -123,6 +123,12 @@ export const ItemEditDialog: React.FC<IItemEditDialogProps> = ({
     };
   };
 
+  const interactiveFromItemToEdit = (itemToEdit: ISectionItem) => {
+    const interactive = camelToSnakeCaseKeys(itemToEdit.data);
+    interactive.interactive_item_id = `interactive_${itemToEdit.id}`;
+    return interactive;
+  };
+
   const modalButtons = [
     {
       classes: "cancel",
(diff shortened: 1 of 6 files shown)


lara-typescript/src/section-authoring/components/item-edit-dialog.tsx
```diff
@@ -156,6 +156,7 @@ export const ItemEditDialog: React.FC<IItemEditDialogProps> = ({
         return <ManagedInteractiveAuthoring
                 managedInteractive={managedInteractive}
                 libraryInteractive={libraryInteractive}
+                interactive_item_id={`interactive_${itemToEdit.id}`}
```
scytacki, line 159: looking at this again, I suggest making a method like 
``` 
const interactiveFromItemToEdit = (itemToEdit: ISectionItem) => {
  const interactive = camelToSnakeCaseKeys(itemToEdit.data);
  interactive.interactive_item_id = `interactive_${itemToEdit.id}`;
  return interactive;
};
```

And then use that to make the `managedInteractive` here and the `interactive` below like:
`const managedInteractive = interactiveFromItemToEdit(itemToEdit);`

With that change you don't need to add the new property to the ManageInteractiveAuthoring and MwInteractiveAuthoring components. 
And the existing code which was setting the `interactive_item_id` like on line 114 of the old managed-interactives/index.tsx  could remain the same.

scytacki (changes requested):
on a second review I found a way to reduce the number of changes to implement this, and I think it would make it a bit more consistent.

lara-typescript/src/section-authoring/components/item-edit-dialog.tsx
emcelroy: Nice. That's definitely better. I've made those changes in the latest commit.

scytacki (approved):
Looks good to me.

PR #887, opened 2022-01-06
Fixes updating browser tab title on page nav authoring mode

eireland:
Took the solution from [here](https://stackoverflow.com/questions/46160461/how-do-you-set-the-document-title-in-react)
Tried all the other simpler ways and couldn't get it to update on page change.

### DIFF
--- a/lara-typescript/src/section-authoring/hooks/use-title.ts
+++ b/lara-typescript/src/section-authoring/hooks/use-title.ts
@@ -1,0 +1,11 @@
+import { useEffect } from "react";
+
+export function useTitle(title: string) {
+  useEffect(() => {
+    const prevTitle = document.title;
+    document.title = title;
+    return () => {
+      document.title = prevTitle;
+    };
+  });
+}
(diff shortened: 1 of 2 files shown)


lara-typescript/src/section-authoring/hooks/use-title.ts
```diff
@@ -0,0 +1,11 @@
+import { useEffect } from "react";
+
+export function useTitle(title: string) {
+  useEffect(() => {
+    const prevTitle = document.title;
+    document.title = title;
+    return () => {
+      document.title = prevTitle;
+    };
+  });
```
kswenson, line 10: Without a dependency array, this will run on every render, which means the original title will be lost by the time of the second render. You probably want to list `[title]` as the dependencies and if you really want to be able to restore the original title from before the hook ran you should store it in a ref (via `useRef`).

emcelroy (approved):
LGTM and works as expected when tested locally. Thanks!

PR #888, opened 2022-01-07
Fix adding additional item in primary column always adds at the top

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/180760254

[#180760254]

Items' position values do not always correspond to the number of items in a column. For example, if there are two items in a column, their respective position values could be 2 and 3 instead of 1 and 2. This change makes sure to increment the actual position value of the last item in the column to determine the position value for a newly added item.


knowuh (approved):
looks good to me, thank you.

PR #889, opened 2022-01-07
Fix text block default half width setting

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/180837836

[#180837836]

This migration changes the default half-width setting for text blocks to false.


knowuh (approved):
looks good, thank you.

PR #890, opened 2022-01-10
Fixes copying pages

eireland:
Some of the copy page code got stomp over. This refixes the page copying.


emcelroy (changes requested):
I think the `pageOptions` method needs to be updated to match what was happening in the commented-out lines 83-86. It should not include a completion page as an option, and it should add " (hidden)" to options for any pages that are hidden.

emcelroy (approved):
I was going to say you missed the completion page piece, but after looking at it again, it seems to me an author may want to copy the page to be immediately before the completion page. 

It'd be nice if we could disable the After position option if a completion page is selected, or force the position to be Before in that case, but I don't think that's absolutely necessary.

eireland:
I'll merge this one just so it's not completely broken, and work on the completion page in another PR.

PR #891, opened 2022-01-11
Fix Link Saved Work From field value not saving correctly.

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/180760275

[#180760275]

These changes add `linked_interactive_item_id` to the MWInteractive's `to_hash` method. This makes it possible for the edit form to populate the Link Saved Work From field with the saved value.

They also update some URLs in the Gemfile that were no longer working.


knowuh (approved):
Looks great, thank you.

PR #895, opened 2022-01-24
Shows special characters in authoring activity title

eireland:


emcelroy (commented):
This looks like it'll work fine, but I wonder about importing that file from the example-interactives folder. You may know better, but it seems to me that it'd be better to move the render-html.ts file to lara-typescript/src/shared and then import it from there in both lara-typescript/src/section-authoring/components/page-header/components/page-header.tsx and in lara-typescript/src/example-interactives/src/use-glossary-decoration.ts. Does that make sense?

If so, it could probably also be used in lara-typescript/src/section-authoring/components/text-block-preview.tsx to replace its `processContent` method.

eireland:
Good catch!

emcelroy (approved):
Looks great -- thanks 👍

PR #896, opened 2022-01-25
Fix author cannot upload image files

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/180963196

[#180963196]

Corrects how `firebase_app` value is sent during Firebase JWT request. Previously the request body was being set to `{opts: {firebase_app: "token-service"}}`. Now it will be set to `{firebase_app: "token-service"}`.


scytacki (approved):
Looks good to me.

PR #897, opened 2022-01-25
Fix adding new page from activity settings adds it after completion page

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/180980347

[#180980347]

The lines added to the interactive pages controller's `create` method were taken from the `create_page` method in app/controller/api/v1/interactive_pages_controller.rb.


knowuh (approved):
Code looks good to me.

PR #898, opened 2022-01-25
Fix change page layout doesn’t work as expected -- image outside column

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/180980286

[#180980286]

This change adds a limit on how wide an image within a text block can be so it won't extend outside the boundary of its containing column. (The problem identified by the PT story doesn't actually require a layout change to manifest. Simply adding a largish image to a text block that's within a secondary column will reveal the bug.)

There's an additional CSS tweak that ensures an image in a text block also doesn't extend beyond the bottom boundary of the text block.


eireland (approved):
Looks good

PR #899, opened 2022-01-31
Fix copy section not working

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/180979850

[#180979850]

These changes reinstate authoring API URL values that were not included in previous work on the `new-sections` branch. The missing authoring API URL information was preventing Image Questions' snapshot targets from being correctly set. That was in turn causing an error preventing sections that contained snapshot image questions from being duplicated.

### DIFF
--- a/app/models/managed_interactive.rb
+++ b/app/models/managed_interactive.rb
@@ -110,6 +110,8 @@ class ManagedInteractive < ActiveRecord::Base
     "iframe interactive"
   end
 
+  # to_hash is used by the export and duplicate methods here, and the create_page_item, update_page_item, 
+  # and generate_item_json methods in controllers/api/v1/interactive_pages_controller.rb
   def to_hash
     # Deliberately ignoring user (will be set in duplicate)
     {
@@ -144,6 +146,20 @@ class ManagedInteractive < ActiveRecord::Base
     hash[:linked_interactives] = linked_interactives_list
     hash
   end
+
+  # to_interactive is used by the to_authoring_preview_hash method above. 
+  # 
+  # It differs from to_hash by including:
+  # id, url, native_width, native_height, enable_learner_state, show_delete_data_button, has_report_url, 
+  # click_to_play, click_to_play_prompt, full_window, image_url, aspect_ratio, aspect_ratio_method, 
+  # no_snapshots, linked_interactive_id, linked_interactive_type
+  # 
+  # It also differs from to_hash by not including: 
+  # library_interactive_id, url_fragment, inherit_aspect_ratio_method, custom_aspect_ratio_method, 
+  # inherit_native_width, custom_native_width, inherit_native_height, custom_native_height, 
+  # inherit_click_to_play, custom_click_to_play, inherit_full_window, custom_full_window, 
+  # inherit_click_to_play_prompt, custom_click_to_play_prompt, inherit_image_url, custom_image_url, 
+  # linked_interactives
 
   def to_interactive
     # NOTE: model_library_url is missing as there is no analog
@@ -180,7 +196,7 @@ class ManagedInteractive < ActiveRecord::Base
   end
 
   def export
-    hash = to_hash()
+    hash = to_hash().except!(:linked_interactives)
     hash.delete(:library_interactive_id)
     hash[:library_interactive] = library_interactive ? {
       data: library_interactive.to_hash()
--- a/app/models/managed_interactive.rb
+++ b/app/models/managed_interactive.rb
@@ -133,7 +133,8 @@ class ManagedInteractive < ActiveRecord::Base
       inherit_click_to_play_prompt: inherit_click_to_play_prompt,
       custom_click_to_play_prompt: custom_click_to_play_prompt,
       inherit_image_url: inherit_image_url,
-      custom_image_url: custom_image_url
+      custom_image_url: custom_image_url,
+      linked_interactives: linked_interactives_list
     }
   end
 
@@ -172,8 +173,9 @@ class ManagedInteractive < ActiveRecord::Base
   end
 
   def duplicate
+    new_interactive_hash = self.to_hash.except!(:linked_interactives)
     # Generate a new object with those values
-    ManagedInteractive.new(self.to_hash)
+    ManagedInteractive.new(new_interactive_hash)
     # N.B. the duplicate hasn't been saved yet
   end
 
--- a/app/models/managed_interactive.rb
+++ b/app/models/managed_interactive.rb
@@ -199,8 +199,8 @@ class ManagedInteractive < ActiveRecord::Base
   end
 
   def export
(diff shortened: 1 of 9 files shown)


emcelroy:
@scytacki Before I try updating the failing tests, I wanted to get your thoughts on these changes. Do you see any problems that would require a different approach?

knowuh:
@emcelroy -- Not sure if you saw, but we broke a few tests related to import / export hash:

`./spec/models/managed_interactive_spec.rb:198`
They seem to all be "Can't mass-assign protected attributes: linked_interactives"

https://app.travis-ci.com/github/concord-consortium/lara/builds/245693382

app/models/managed_interactive.rb
```diff
@@ -133,7 +133,8 @@ def to_hash
       inherit_click_to_play_prompt: inherit_click_to_play_prompt,
       custom_click_to_play_prompt: custom_click_to_play_prompt,
       inherit_image_url: inherit_image_url,
-      custom_image_url: custom_image_url
+      custom_image_url: custom_image_url,
+      linked_interactives: linked_interactives_list
```
scytacki, line 137: If you need to remove the linked_interactives on the duplicate call, then you probably need to remove them from the export too. This is because export and import is basically the same as duplicate.

But I also wonder about this approach. The AP must know about the linked_interactive and it is based on the export, which I would assume uses to_hash. Perhaps it uses to_interactive?  

Could you add a comment above `to_hash` and `to_interactive` which describes what uses each of them, and how they are different?  It seems we keep running into problems about which one to use where. I no longer remember why we have both of them.

scytacki (changes requested):
I'm a little worried about the changes to `to_hash` because I don't remember it is used and the name of the function doesn't make that clear.

Also I'm surprised that the fix for copying sections includes changes to support linked interactives as well as the authoring api changes.   Is this PR both for copying sections and fixing linked interactives?

emcelroy:
@scytacki Sorry it wasn't clear, but these changes address a problem preventing sections containing snapshot image questions from being duplicated. In investigating the copy section bug, I found that only sections containing snapshot image questions could not be duplicated. The problem there seemed to be related to linked interactives being broken -- the snapshotTarget value wasn't being set properly. I wasn't aware of the separate story about linked interactives at the time.

I've added some notes about `to_hash` and `to_interactive` to the code.

The only thing currently using `to_interactive` in new-sections is the ManagedInteractive's `to_authoring_preview_hash` method. I don't think `to_authoring_preview_hash` is actually used anywhere now. There are just a few instances of it in the old HAML files.

At first we used it in a few places in app/controllers/api/v1/interactive_pages_controller.rb, but dropped it. The lines used to look like something like this:

`data: pi.embeddable.respond_to?(:to_interactive) ? pi.embeddable.to_interactive : pi.embeddable.to_hash`

We changed them to something more like this:

`data: pi.embeddable.to_hash`

Using `to_interactive` would break saving changes to items. It seemed like that was due to the updated data including some properties/values that `to_interactive` includes but which ManagedInteractive wouldn't allow to be updated because they're dynamically generated, not updatable fields.

app/models/managed_interactive.rb
emcelroy: Thanks. Removing `linked_interactives` from the export got the breaking tests to pass.

I added comments about `to_hash` and `to_interactive` though I'm not sure we want to keep them exactly as is.

Would it be worth investigating using `to_authoring_hash` (in app/models/base_interactive.rb) instead of `to_hash` for managed and MW interactives? I see it includes `linked_interactives` too. I think we may have briefly noted that, but then didn't try using it, but I can't recall why.

scytacki (changes requested):
The code seems good to me now. I don't think you should switch to `to_authoring_hash` at this point. But it is something to keep in mind if there are more updates. It isn't great that we now have `to_hash().except!(:linked_interactives)` in 2 places in the code. 

My two requests are about comments:
- add more to the `to_interactive` comment. From your comment in the PR it sounds like it is not used in used in new-sections. So it would be useful to indicate that in the code. Something like: "This is deprecated. It is not used by active code in LARA 2. We plan to remove it when LARA 2 is moved to production"
- document the linked_interactives property with a comment. How is it handled by the duplicate and export? Why does including it break duplicate and export? How is it handled by the AP (which uses the export)?

app/models/managed_interactive.rb
```diff
@@ -189,13 +190,17 @@ def to_interactive
...
+    # properties in the hash can, and so causes an error when we try to create the duplicate interactive.
     new_interactive_hash = self.to_hash.except!(:linked_interactives)
     # Generate a new object with those values
     ManagedInteractive.new(new_interactive_hash)
     # N.B. the duplicate hasn't been saved yet
   end
 
   def export
+    # Remove linked_interactives from the hash since it can't be mapped to a database column like the other 
+    # properties in the hash can, and so would cause an error on import.
```
scytacki, line 203: I don't think this comment is correct. As are as I can tell from the code the the export already includes linked_interactives. It seems to be added here: https://github.com/concord-consortium/lara/blob/b4bf12d4411cdbd7eacd09117640b658a3e94fac/app/services/lara_serialization_helper.rb#L25-L35 

So the problem is not the import. I haven't checked, but I'd guess the format the linked_interactives array added by the lara serialization helper is different than the format being added by to_hash

app/models/section.rb
```diff
@@ -118,8 +118,9 @@ def duplicate(helper=nil)
           embed.primary_linked_items.each do |pli|
             primary = helper.get_copy(embed)
             secondary = helper.get_copy(pli.secondary.embeddable)
-            if primary && secondary
-              lpi = LinkedPageItem.new(primary_id: primary.page_item.id, secondary_id: secondary.page_item.id, label: pli.label)
+            secondary_page_item = secondary.page_item ? secondary.page_item : pli.secondary.embeddable.page_item
```
scytacki, line 121: A comment here would be useful. Something like:
"If the secondary embeddable is in a different section, it won't be copied. So in this case the new primary embeddable references the original secondary embeddable."

scytacki (changes requested):
I found one issue with the comments, and suggested an additional comment for the new code to fix the case when a secondary interactive is in a different section than the one being copied.

app/models/managed_interactive.rb
emcelroy: @scytacki Thanks. I've updated the comment. 

It seems like importing is simply broken for an unrelated reason and I somehow missed that. I checked on lara-demo.concordqa.org and confirmed importing is broken there. I'll make a separate story about that.

emcelroy: (There's actually [a story about importing already](https://www.pivotaltracker.com/story/show/179237412).)

scytacki (approved):
Looks good. 👍

PR #900, opened 2022-02-07
Update Project form to support AP footer styling

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181059664

Adds several fields to Projects for use in Activity Player footer. Removes unused Help field from Projects.


scytacki (approved):
This looks good to me.
I see that this doesn't include the react form. I assume that will be in a separate PR. 
This makes sense to me.

My only concern is that authors would add some invalid html to the new fields. But I think it is very rare that an author edits those fields themselves, and it should be easy enough to fix later even if they do.

emcelroy:
@scytacki Thanks. Yes, there's a new, separate [PT story about the React conversion](https://www.pivotaltracker.com/story/show/181174622).

I think when we do the React conversion and set the new fields to use the RTE, that will help prevent invalid HTML being added.

PR #901, opened 2022-02-10
Update Project form to support AP footer

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181059664

[#181059664]

Adds several fields to Projects for use in Activity Player footer. Removes unused Help field from Projects.


db/migrate/20220209211536_add_ap_footer_fields_to_projects.rb
```diff
@@ -0,0 +1,21 @@
+class AddApFooterFieldsToProjects < ActiveRecord::Migration
```
scytacki, line 1: If you give this migration the same file name as the one in the new-sections (LARA 2) branch then the migration should not run again when new-sections is merged in.

scytacki (changes requested):
It would better to use the same migration file name. Then the migrations won't be run twice when switching between branches.

After this is PR is merged to master it would be good to merge master into new-sections and deal with the conflicts while they are fresh.

PR #903, opened 2022-02-11
Convert updated Project form to React

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181174622

[#181174622]

In addition to converting the Project form, these changes also convert the Projects index page to React. That seemed to make sense since in the current setup, you're redirected there when you create a new project, plus it was pretty simple and it seemed like it'd be good to convert the whole Project section in one go.


scytacki (approved):
This looks really good to me.

Perhaps @knowuh should take a look too. I'm not familiar with the new authoring patterns. For example perhaps it should be using react query? Or perhaps that was only added in Lara 2, so you have to go without it to make this easy to apply to master.

knowuh (approved):
This looks good to me too.  I had one small comment.

emcelroy:
@knowuh @scytacki Thanks. I've removed that duplicate export. I also made a small change that Scott suggested offline to how we skip verify auth token in the project controller to make it more consistent with what we do in the interactive pages controller.

I didn't think of this till now, but do you think I should use this branch as an opportunity to merge `master` into `new-sections`?

scytacki:
I was thinking you would put this code in master and then when you merge master into new-sections, then new-sections would have it too. But I see that you did the work on new-sections.

I'd guess we could get by with out the slate editor fields on master until June when we start using new-sections.  However I'd guess there will be other changes we are going to have to do on master in the meantime, so if this code is only on new-sections then it is like to cause future merge conflicts.  

So I'd recommend trying to redo the work on master.  But if it is hard to redo on master we should re-evaluate. We should try to balance: 
- how hard is it to add to master and then merge to new-sections
- vs - 
- how likely are we to have merge conflicts with future master work, and then possible more merge issues when putting finally merging new-sections into master later on.

emcelroy:
@scytacki I copied all the changes to master and things [look pretty OK](https://github.com/concord-consortium/lara/pull/907). Everything worked when I tested it out in the browser. There's a single failing controller spec test, but I don't think it's anything significant. There's also a bunch of CodeClimate suggestions, but I think we can ignore those for now?

scytacki:
@emcelroy sounds good to me. So if you can fix that failing test then we can merge that PR.  And we can close this PR.

PR #906, opened 2022-02-18
update documentation about addLinkedInteractiveStateListener

scytacki:
@emcelroy was looking at an issue in the side-by-side interactive and we decided that the graph interactive was making an assumption about how the the AP implemented `addLinkedInteractiveStateListener`.

The graph is assuming that the host will tell it the state of the linked interactive as soon as it adds a listener even if this state isn't changed. This way the graph will get the data from the linked interactive to display even if the linked interactive isn't manipulated by the user. 

This updated documentation attempts to describe how it is supposed to behave. 

Looking at the code it seems like this is how the AP host already behaves because it is using a FireStore query with an `onSnapshot` handler. 
https://github.com/concord-consortium/activity-player/blob/1d3ecb360c44592eb9751b0604513966d1051b29/src/firebase-db.ts#L178
I would assume that the `onSnapshot` handler is called immediately when it is registered since it is basically how the query is first "executed".

### DIFF
--- a/docs/interactive-api-client/globals.md
+++ b/docs/interactive-api-client/globals.md
@@ -511,6 +511,11 @@ ___
 
 ▸ **addLinkedInteractiveStateListener**‹**LinkedInteractiveState**›(`listener`: function, `options`: [IAddLinkedInteractiveStateListenerOptions](interfaces/iaddlinkedinteractivestatelisteneroptions.md)): *void*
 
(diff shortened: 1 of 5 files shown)


pjanik (approved):
Looks good to me 👍

emcelroy (approved):
Looks good to me -- thanks 👍

PR #907, opened 2022-02-18
Convert updated Project form to React (on master)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181174622

[#181174622]

These changes convert the Projects form to React on LARA's master branch. In addition to converting the Project form, these changes also convert the Projects index page to React. That seemed to make sense since in the current setup, you're redirected there when you create a new project, plus it was pretty simple and it seemed like it'd be good to convert the whole Project section in one go.


scytacki:
I'm punting this over to @knowuh. I looked at the code when it was based on `new-sections` and it looked good there:
https://github.com/concord-consortium/lara/pull/903

app/controllers/api/v1/projects_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+      render json: {project: @project}, status: 200
+    rescue ActiveRecord::RecordNotFound
+      render json: {error: "Project not found"}, status: 404
+    end
+  end
+
+  # POST /api/v1/projects
+  def create
+    @project = Project.new(params[:project])
+    authorize! :update, @project
```
knowuh, line 26: This should work, but I am wondering if we should use authorize! :create, @project instead?

knowuh (approved):
Looks good to me Ethan, thank you.

PR #913, opened 2022-03-28
WIP: Added attachments example interactive [#181434678]

dougmartin:
This example interactive shows how to use the attachments api to upload a file.


scytacki:
@dougmartin should I review this? I see it is still in draft.

dougmartin:
@scytacki - yes, I meant to move it out of draft when I assigned it to you.  After talking with Ethan last night I'm going to go ahead and add the rest of the attachments support to Lara API and the example interactive today so he is set for the open response work.  I'll base that work off of this PR.

lara-typescript/src/example-interactives/src/attachments/runtime.tsx
```diff
@@ -0,0 +1,119 @@
...
+  }, [initMessage]);
+
+  useEffect(() => {
+    const getUrls = async () => {
+      const names = Object.keys(attachments);
+      for await (const name of names) {
+        const attachment = attachments[name];
+        if (!attachment.url) {
+          try {
+            const url = await getAttachmentUrl({name});
```
scytacki, line 37: One problem with this, is that the URLs expire. So if the user loads the page and waites past the experation date the file won't load unless the audio tag does pre caching.

Even if the audio tag does pre caching, it would be good to make that clear somehow in this example interactive since developer writing attachment interactives might refer to this example.

lara-typescript/src/interactive-api-host/attachments-api/attachments-manager.ts
```diff
@@ -79,13 +79,17 @@ export class AttachmentsManager {
...
 
   public getSignedReadUrl(attachmentInfo: IReadableAttachmentInfo, options?: ISignedReadUrlOptions) {
-    const { publicPath, folder } = attachmentInfo;
+    const { publicPath, folder, contentType } = attachmentInfo;
     const { expiresIn = kDefaultReadExpirationSec } = options || {};
-    return this.getSignedUrl(folder, "getObject", { Key: publicPath, expiresIn });
+    return this.getSignedUrl(folder, "getObject", {
+      Key: publicPath,
+      expiresIn,
+      ContentType: contentType
```
scytacki, line 91: I'm not sure, but I don't see how the ContentType would actually be used here.
During a post it is needed to allow the client to send the content type header. But during a get I'd assume the client isn't sending a header, and I don't think S3 will do something to enforce the file type matches the content type of the signature.

scytacki (changes requested):
Mostly looks good. My only concern is sending the `ContentType` to the getObject signed url request.

lara-typescript/src/example-interactives/src/attachments/runtime.tsx
dougmartin: I've updated the code to load the audio on a user click now - https://github.com/concord-consortium/lara/pull/913/commits/3a458ace5a4f39a0448502c5c95ac6ab1ffe5714

lara-typescript/src/interactive-api-host/attachments-api/attachments-manager.ts
dougmartin: I've removed the content type from the signed read url request.  One note - the lower `getSignedUrl` request is defaulting the content type to `text/plain`, but again this doesn't seem to matter on read urls.

scytacki (approved):
Looks good to me.

PR #914, opened 2022-04-04
Changed getReportItemAnswer to return array of items instead of a single response type [#181516278]

dougmartin:


scytacki (approved):
looks good to me

scytacki (approved):
looks good to me

PR #918, opened 2022-04-06
Updated example attachments interactive report item

dougmartin:
This updates the runtime to set interactive state so the question shows the answer icon in the portal report and adds a simple visible boolean if there are attachments.


dougmartin:
@scytacki these are some changes I added to the example attachments interactive when I was testing the update to the portal report to view attachments.

scytacki (approved):
Seems fine to me.

PR #919, opened 2022-04-06
Added glossary model [#181798596]

dougmartin:
NOTE: This also includes a gem file lock update and schema update that must have been missed in a previous commit


scytacki (approved):
Seems good to me.

PR #921, opened 2022-04-07
Added fake glossary plugin to activity export if a glossary model is assigned [#181798706]

dougmartin:
This also removes any existing glossary plugin is it is already assigned to the activity.

The first plugin named "glossary" found is used as the values for the fake glossary plugin,
with the s3Url changed to point to the glossary api json endpoint.


app/models/lightweight_activity.rb
```diff
@@ -154,7 +154,28 @@ def export
...
-    # TODO: add special code to push glossary plugin onto plugin list if glossary is set
+
+    # if the activity has a glossary model assiged to it, add the fake glossary plugin to the list of plugins
+    # replacing any existing glossary plugins and using the existing glossary plugin's approved script
+    if self.glossary_id && approved_glossary_script = ApprovedScript.find_by_label("glossary")
+      # remove any existing glossary script
+      activity_json[:plugins].delete_if { |plugin| plugin[:component_label] == "glossary" }
+
+      fake_glossary_plugin = {
+        id: 0,
```
scytacki, line 165: A sequence could have multiple activities each with their own glossary.  
If this is loaded in the AP I suspect if they all have the same id, it might cause problems.

Also I'm not sure, but perhaps this plugin id is used when the glossary plugin stores learner data. If the id has to be a number, I suggest using `glossary_id*1_000_000_000+activity_id`. That would work until we have more than a 1 billion activities and 9 billion glossaries. I'm assuming we can use int64 for these numbers. It is hacky, string ids would be better, but that might require changes to the AP.  

I don't remember what we do with glossaries shared across sequences. If we support that (or need to) that would affect this. I think they currently don't share the student data. There might be a story about this scheduled for later.

app/models/lightweight_activity.rb
```diff
@@ -154,7 +154,28 @@ def export
...
+        description: nil,
+        author_data: JSON.generate({
+          version:"1.0",
+          glossaryResourceId: "this-is-a-fake-glossary-resource-id",
```
scytacki, line 169: If this is used in the student data saving or in some other way by the AP to manage plugins, this might need to be unique. If you know more about the requirements of it, adding a comment would be helpful.

scytacki (changes requested):
I think there might be problems with the ids you are using in the generated glossary plugin.

app/models/lightweight_activity.rb
dougmartin: The `glossaryResourceId` is not used except by the glossary plugin to generate the s3Url.  This could have been left blank but we didn't.

app/models/lightweight_activity.rb
dougmartin: We fixed this in a commit in a newer PR, as it also changes the spec to test the ids whereas the ids are not tested in this commit.  You can see the changes here:

https://github.com/concord-consortium/lara/pull/922/commits/dea6094845a1de48165972c38a33b008fd64360a

app/models/lightweight_activity.rb
scytacki: Can you add a comment in the code describing this? Given that you've built up a train a commits you could add it to another PR.

app/models/lightweight_activity.rb
scytacki: sounds good

scytacki (approved):
seems good to me

PR #924, opened 2022-04-08
Authorized users can create, edit, update, delete glossaries [#181678938]

lublagg:


app/views/glossaries/_form.html.haml
```diff
@@ -0,0 +1,17 @@
+.sequence_form
+  = form_for(@glossary) do |f|
+    - if @glossary.errors.any?
+      #error_explanation
+        %h2
+          = pluralize(@glossary.errors.count, "error")
+          prohibited this project from being saved:
```
scytacki, line 7:
```suggestion
          prohibited this glossary from being saved:
```

scytacki (approved):
Looks good just found one place where it seems the project string wasn't replaced with glossary.

app/views/glossaries/_form.html.haml
dougmartin: Fixed in https://github.com/concord-consortium/lara/pull/924/commits/ea1987827b73ab776e3a3c50a8cf38a9095835e3

PR #927, opened 2022-04-11
Fixed rspec error due to missing user in glossary

dougmartin:
This crept in between two commits that added this test and another commit that added a validation of the user in the model


lublagg (approved):
Looks good

PR #929, opened 2022-04-12
Tweaks for 4/12/2022 demo

dougmartin:
- Added temporary way to edit glossary json
- Updated click handler for view/edit glossary

### DIFF
--- a/app/views/lightweight_activities/_enable_disable_runtime_fields.html.haml
+++ b/app/views/lightweight_activities/_enable_disable_runtime_fields.html.haml
@@ -30,9 +30,10 @@
     })
 
     $view_edit_glossary.on('click', function (e) {
+      var edit_url = "#{edit_glossary_path(-1)}"
       e.preventDefault()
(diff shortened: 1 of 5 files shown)


app/views/lightweight_activities/_enable_disable_runtime_fields.html.haml
```diff
@@ -30,9 +30,10 @@
     })
 
     $view_edit_glossary.on('click', function (e) {
+      var edit_url = "#{edit_glossary_path(-1)}"
```
scytacki, line 33: Add a comment about what the `-1` is for

scytacki (approved):
looks good

app/views/lightweight_activities/_enable_disable_runtime_fields.html.haml
dougmartin: Fixed in https://github.com/concord-consortium/lara/pull/931

PR #932, opened 2022-04-12
Added glossary authoring container [#181873380]

dougmartin:
This adds a container div for the glossary authoring along with setting a new GlossaryModelAuthoring namespace in the global LARA variable and then loads the glossary plugin script which will use that info to render into the container.

This also:

1. Moves the lookup of the glossary approved script into the glossary model so that it can be updated in the future to use a setting instead of just returning the first glossary approved script found.
2. Adds an api route to update the name and/or json of a glossary that the glossary authoring plugin will use.
3. Fixes the name of the glossary api spec file


scytacki (approved):
Looks good to me.

PR #933, opened 2022-04-13
Fixed sticky header z-index bug [#181679060]

dougmartin:
Before this fix the modals were displayed under the sticky header instead of over it.


lublagg (approved):
looks good!

PR #935, opened 2022-04-19
Snapshot content is blank

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181808855

[#181808855]

Add `linkedInteractives` to `IReportInitInteractive` to support linking within the side-by-side interactive in reports/dashboard. This fixes a problem discovered by QA while testing the fix for snapshots. The changes here are required to make the changes in [this question-interactives pull request](https://github.com/concord-consortium/question-interactives/pull/182) work.

### DIFF
--- a/docs/interactive-api-client/interfaces/ireportinitinteractive.md
+++ b/docs/interactive-api-client/interfaces/ireportinitinteractive.md
@@ -20,6 +20,7 @@
 * [authoredState](ireportinitinteractive.md#authoredstate)
 * [hostFeatures](ireportinitinteractive.md#hostfeatures)
 * [interactiveState](ireportinitinteractive.md#interactivestate)
+* [linkedInteractives](ireportinitinteractive.md#linkedinteractives)
 * [mode](ireportinitinteractive.md#mode)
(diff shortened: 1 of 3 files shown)


dougmartin (approved):
Looks good - but I think there also needs to be an update, at some point, to the portal-report code to include this when it creates the init interactive message.

kswenson (approved):
👍 LGTM

emcelroy:
> Looks good - but I think there also needs to be an update, at some point, to the portal-report code to include this when it creates the init interactive message.

@dougmartin It sounds like that could probably wait? FWIW, I've tested this change with [a branch of question-interactives that depends on the change](https://github.com/concord-consortium/question-interactives/pull/182). Viewing a side-by-side interactive -- the interactive that needed this change -- in a report works and doesn't throw any errors. So I think it may not be necessary to update portal-report, at least not right away? It's not immediately clear to me how to update portal-report in light of this change, so I'm hoping to avoid it for the time being. If you're not sure, though, I can look further into it.

dougmartin:
@emcelroy It should be fine - the next time someone updates the version in the portal report they will get a compile error and need to add the linked interactives to the init message.  It can wait.

PR #938, opened 2022-04-20
Updated glossary authoring to closer match new specs [#181679164]

dougmartin:
- Updated used in activities display
- Updated glossary delete confirmation
- Added new and import buttons in glossary index like in activities list
- Updated all import checks to also check for the ability to import a glossary


lublagg (approved):
Looks good!

PR #939, opened 2022-04-20
Changed glossary visibility so that all authors and admins can see all glossaries [#181798850]

dougmartin:
This also updates the glossary selection UI in the activity editor to split glossaries into two lists and appends the email of the author to each glossary.


lublagg (approved):
Looks great!

PR #941, opened 2022-04-22
Fix a couple of issues with the glossary model authoring integration

dougmartin:
- There was a double title
- The json exported into the template was in Ruby hash format not json format


lublagg (approved):
Looks good! 👍

PR #945, opened 2022-05-03
Moved glossary dropdown down in new activity form [#182051450]

dougmartin:
This matches the edit form as the glossary option is only enabled for AP actvities.  This fix also changes to use the glossary_options_for_select helper.


lublagg (approved):
Looks good to me!

PR #946, opened 2022-05-06
Added library interactive url to list display

dougmartin:


dougmartin:
@scytacki I added this locally while I was testing the library interactive SQL import and thought it would be good to add permanently.

scytacki (approved):
Looks good to me.

PR #947, opened 2022-05-09
Added id and getFirebaseJwtUrl function to glossary model authoring init object [#181946122]

dougmartin:
This is needed so the glossary model authoring can upload files via the token service.


scytacki (approved):
seems fine to me

PR #948, opened 2022-05-09
Added admin settings interface to set current glossary [#181884477]

dougmartin:
This adds a generic key/value settings model with a concrete setting controller that currently only sets the approved glossary script id.  It is designed to be extensible to other settings.

To make id mapping more visible the ids have been added to both the plugin list and library interactive list.

### DIFF
--- a/app/assets/stylesheets/style.scss
+++ b/app/assets/stylesheets/style.scss
@@ -1221,6 +1221,12 @@ body.admin.edit {
   }
 }
 
+.settings-table {
+  td {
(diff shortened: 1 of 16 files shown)


scytacki (approved):
Seems fine to me.

I was a little confused that `view.html.haml` is actually the place to edit it, but that is fine with me.

dougmartin:
Yeah - since this isn't a standard resource controller/view in that one page is both index, show and edit I went with view.

PR #949, opened 2022-05-11
Update glossary json column to store more than 64kb [#182141060]

dougmartin:
The json column now maxes out at 16777215 (MySQL's mediumtext)


scytacki (approved):
looks good to me

PR #950, opened 2022-05-12
Changed display of glossary option value and added missing slash in settings breadcrumbs

dougmartin:


dougmartin:
I'm going to merge this without review as it is two small text tweaks

PR #951, opened 2022-05-18
180951937 Item Picker wont lock choice

knowuh:
- Item picker now allows you to chose a new item without unchosing the first.

[#180951937]
https://www.pivotaltracker.com/story/show/180951937

### DIFF
--- a/lara-typescript/src/section-authoring/components/section-item-picker.tsx
+++ b/lara-typescript/src/section-authoring/components/section-item-picker.tsx
@@ -30,11 +30,26 @@ export const SectionItemPicker: React.FC<IProps> = (props) => {
   const quickAddItems = api.getAllEmbeddables.data?.allEmbeddables.filter(e => e.isQuickAddItem);
   const { onClose, onAdd } = props;
   const modalIsVisible = true;
-  const [itemSelected, setItemSelected] = useState(false);
-  const [currentSelectedItem, setCurrentSelectedItem] = useState<ISectionItemType|undefined>();
+  const [currentSelectedItem, setCurrentSelectedItem]
+    = useState<ISectionItemType|null>(null);
   const [allItemsList, setAllItemsList] = useState(allItems?.allEmbeddables || []);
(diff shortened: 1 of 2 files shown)


lara-typescript/src/section-authoring/components/section-item-picker.tsx
```diff
@@ -71,19 +86,18 @@ export const SectionItemPicker: React.FC<IProps> = (props) => {
   const setItemClasses = (isSelectedItem: boolean) => {
     const classes = classNames("assessmentItemOption", {
       selected: isSelectedItem,
-      disabled: !isSelectedItem && currentSelectedItem !== undefined
+      disabled: false
```
emcelroy, line 89: If this is always going to be false, maybe we don't need that class name for item buttons?

lara-typescript/src/section-authoring/components/section-item-picker.tsx
```diff
@@ -134,12 +148,11 @@ export const SectionItemPicker: React.FC<IProps> = (props) => {
         {allItemsList.map((item, index) => {
           const isSelectedItem = currentSelectedItem === item;
           const itemClass = setItemClasses(isSelectedItem);
-          const itemDisabled = itemSelected && !isSelectedItem ? true : false;
           return (
             <li key={`ai-${index}`}>
               <SectionItemButton
                 item={item}
-                disabled={itemDisabled}
+                disabled={false}
```
emcelroy, line 155: Seems like the `disabled` property of SectionItemButton is also always going to be false. So maybe we can do away with that property?

emcelroy (approved):
Looks good and works as expected when tested locally. The addition of the Escape keypress option is especially nice 👍

I left a couple non-blocking comments/questions.

lara-typescript/src/section-authoring/components/section-item-picker.tsx
knowuh: For now, I am just going to leave the components defined as they are because I cant rule out a use case later.

PR #952, opened 2022-05-18
All authors can export and import. [#182202035]

lublagg:


app/views/home/home.html.haml
```diff
@@ -16,7 +16,7 @@
       - if can? :create, Glossary
         %a{ :href => new_glossary_path }
           %button=t("NEW_GLOSSARY_BUTTON")
-      - if can? :import, [LightweightActivity, Sequence, Glossary]
+      - if can? :import, LightweightActivity, Sequence, Glossary
```
scytacki, line 19: So this was broken before? Interesting...

db/migrate/20190621195718_add_can_export_to_user.rb
```diff
@@ -1,5 +0,0 @@
-class AddCanExportToUser < ActiveRecord::Migration
-  def change
-    add_column :users, :can_export, :boolean, default: false
-  end
-end
```
scytacki, line 5: When changing the database, you shouldn't delete migration(s). Instead you should add a new migration that changes it. So in this case it would be a new migration that deletes the field. When migrations are run on an existing database, only new migrations are run.  So deleing a migration won't cause any change to the existing database.   

Also in many cases there will be migrations that build off of the one that you might want to delete, so in that case running the migrations on a fresh DB would fail unless you checked and updated all of the migrations after the deleted one.

scytacki (changes requested):
Mostly looks good. Just a few requested changes.

PR #954, opened 2022-05-25
Preview interactives in edit form

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/182075235

[#182075235]

These changes modify the assessment item edit form to include a preview of the item being edited. Changes made to the item's settings appear in the preview as they are made. Some settings/features are not yet supported by the preview (these include aspect ratio,  click to play options, and hints).

Also included is a fix for the issue of assessment items on the main page not updating after they are edited until you reload the page. The fix for that is similar to the method used for updating the edit form previews (passing a `resetCount` value down to the iframe component).


lara-typescript/src/section-authoring/components/text-block-edit-form.tsx
```diff
@@ -1,34 +1,60 @@
...
   const pathToTinyMCECSS = api.pathToTinyMCECSS;
 
+  const _updatePreview = (update: any) => {
+    if (handleUpdateItemPreview) {
+      updates.current = {...updates.current, ...update};
+      handleUpdateItemPreview(updates.current);
+    }
+  };
+
+  const updatePreview = useCallback(debounce(_updatePreview, 500), []);
```
knowuh, line 36: Hmn, I haven't used `useCallback` much myself, so I went and read the documentation, but I guess I still don't understand this.

https://reactjs.org/docs/hooks-reference.html#usecallback

knowuh (commented):
Looks good so far.

Usually when I see something like the `resetCount` property being used trigger component rendering, I see it as a sign that we aren't being careful enough about how we are passing properties.

For example, I think if we have a component for displaying a person, its better to pass each of the persons properties as strings & ints then to pass a single property to the person. If we change the persons name, the person object doesn't change, so we don't get a re-render of the component. There are other approaches too. A hack I do a lot when I a property of an object changes, is to create a new object merging the old object with the changes. Because the object is new, it will re-render the component.  eg:  `setPerson({...person,...updates})`; 


Any way, that might not be what is going on here. I know when we preview iFrames in certain cases the only thing that we have is a URL that might not change, in which case we need to track something like resetCount specifically.

kswenson:
>Also not sure if you saw this debounce-hook: https://usehooks-ts.com/react-hook/use-debounce

Also potentially useful: https://betterprogramming.pub/how-to-use-debounce-and-throttle-the-right-way-with-react-hooks-bf2c174728e

knowuh:
Thanks @kswenson that article helped me understand what @emcelroy was doing here.  I sometimes worry functional components aren't super easy to grok.

emcelroy:
Noah and I discussed his comments offline and I've made some adjustments, though I've kept the reliance on `resetCount` as that did seem to be the only way to get the preview iframes to re-render. That may be something to revisit in the future, though.

scytacki (approved):
Looks good to me.

PR #955, opened 2022-05-27
Updated linked-state example interactive to add a onOnload timeout [#182307246]

dougmartin:
This optional timeout authored setting allows the page change notifications in AP to be observed by setting a timeout > 500ms for the page change and >20000ms for the page save error message.


scytacki (approved):
Looks good to me.

PR #956, opened 2022-05-28
179644118 add te windowshade to picker

knowuh:
@emcelroy -- just tagging you on this for now.

I rebased this from your latest item previewing work.  I had to make the preview side (right) optionally hide if the item type didn't support previews, so that the TE authoring interface could show itself.

There are a few failing tests that I haven't updated yet.  But it would be good to have your 👀 early.


app/views/layouts/minimal.erb
```diff
@@ -0,0 +1,52 @@
+<!DOCTYPE html>
```
knowuh, line 1: @emcelroy -- This can probably be removed it was from the previous approach where the plugin was rendered through the ERB templates.  I will see if I can back out these chaangse.

app/controllers/api/v1/interactive_pages_controller.rb
```diff
@@ -220,13 +220,35 @@ def create_page_item
...
+      embeddable = Embeddable::EmbeddablePlugin.create!
+      embeddable.approved_script_id = script_id
+      embeddable.author_data = author_data
+      embeddable.component_label = tip_type
+      embeddable.is_half_width = false
+      embeddable.save!
+
     else
-      return error("Only library interactive embeddables, iFrame interactives, and text blocks are currently supported")
+      return error("Unknown embbeddable_type: #{embeddable_type}\nOnly library interactive embeddables, iFrame interactives, and text blocks are currently supported")
```
emcelroy, line 251: Not of critical importance, but should teacher edition windowshades be added to the list of supported embeddables? Or maybe at this point, we don't need to list what's supported since everything we plan on supporting is now supported?

emcelroy (approved):
I think this looks good, thanks 👍 I left one comment/question, but it's not critically important.

PR #957, opened 2022-06-07
Added explicit bold font for bold and strong tags under .textBlock

dougmartin:


emcelroy (commented):
Does removing this cause the text block to render with a font other than Lato? If so, I don't think it'll work since we're trying to be consistent with fonts used on the page.

There's also a `--font-family-bold` in vars.scss. Would it work to add something like: 

`strong { font-family: var(--font-family-bold); }` under `.textBlock`?

emcelroy (approved):
Looks good, thanks! One little thing that probably won't matter much -- and maybe you know this already but it just slipped your mind -- is that there isn't a `<bold>` tag in HTML. There is a `<b>` tag but it's deprecated in favor of `<strong>`. So it should be `b, strong` if you want to support text blocks that may still be using the `<b>` tag for bold text.

PR #958, opened 2022-06-08
Add & Edit TE Wrappers

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/182276802

[#182276802]

### DIFF
--- a/lara-typescript/src/section-authoring/components/section-item-plugin-list.tsx
+++ b/lara-typescript/src/section-authoring/components/section-item-plugin-list.tsx
@@ -1,0 +1,121 @@
+import * as React from "react";
+import { ICreatePageItem, SectionColumns } from "../api/api-types";
+import { UserInterfaceContext } from "../containers/user-interface-provider";
+import { useGetAvailablePlugins, useGetPageItemEmbeddableMetaData,
+         useGetPageItemPlugins, usePageAPI } from "../hooks/use-api-provider";
+import { SectionItemPlugin } from "./section-item-plugin";
+
+import "./section-item-plugin-list.scss";
+
+interface ISectionItemPluginListProps {
+  sectionColumn: SectionColumns;
+  sectionId: string;
+  sectionItemId: string;
+}
+
+export const SectionItemPluginList: React.FC<ISectionItemPluginListProps> = ({
+  sectionColumn,
+  sectionId,
+  sectionItemId
+  }: ISectionItemPluginListProps) => {
+  const availablePlugins = useGetAvailablePlugins();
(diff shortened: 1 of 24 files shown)


lara-typescript/src/section-authoring/components/section-item-plugin-list.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+export const SectionItemPluginList: React.FC<ISectionItemPluginListProps> = ({
+  sectionColumn,
+  sectionId,
+  sectionItemId
+  }: ISectionItemPluginListProps) => {
+  const availablePlugins = useGetAvailablePlugins();
+  const pageItemPlugins = useGetPageItemPlugins(sectionItemId);
+  const pageItemEmbeddableMetaData = useGetPageItemEmbeddableMetaData(sectionItemId);
+  const { userInterface: {editingItemId}, actions: {setEditingItemId}} = React.useContext(UserInterfaceContext);
```
knowuh, line 24: If you wanted to you could extract `editingItemId` ,`wrappedItemId`, `setEditingItemId`, and `setWrappedItemId` into the same destructured assignment here...

lara-typescript/src/section-authoring/components/section-item-plugin-list.tsx
```diff
@@ -0,0 +1,121 @@
...
+      wrapped_embeddable_type: pageItemEmbeddableMetaData.data?.embeddableType
+    };
+    api.addPageItem(newItem);
+  };
+
+  const renderPluginsMenu = () => {
+    const plugins = availablePlugins.isSuccess ? availablePlugins.data.plugins.map((p: any, i: number) => {
+      const alreadyAdded = pageItemPlugins?.find(pi => pi.name === p.name);
+      if (alreadyAdded) {
+        return null;
```
knowuh, line 55: This is going to return an array with null objects and components, which is fine, but you could also `.filter` your list, and then `.map` your plugin components, roughly like this:
```
vailablePlugins.data.plugins.filter(p => pageItemPlugins?.find(pi => pi.name === p.name))
.map((p: any, i: number) => ... etc

```

knowuh (approved):
Looks good to me.

I had some inconsequential suggestions that can be skipped.
👍

emcelroy:
@knowuh Thanks for the suggestions. I made the change regarding the destructuring assignments. I couldn't get the other one about using `.filter` to work. I get the general idea, but couldn't get it to fully work for some unknown reason. Since I'm a bit behind on this work, I'm going to leave things as is in that regard.

PR #960, opened 2022-06-09
Updated lara-typescript.css with missing bold css [#180588632]

dougmartin:


dougmartin:
Merging without review - just a build update of the lara-typescript .css file.

PR #961, opened 2022-06-15
Add and edit Teacher Edition side tips

emcelroy:
PT Stories: 

https://www.pivotaltracker.com/story/show/182276917
https://www.pivotaltracker.com/story/show/182463494
https://www.pivotaltracker.com/story/show/182468721

[#182276917]
[#182463494]
[#182468721]

These changes add support for Teacher Edition side tips to authoring. TE side tips can be added via the item picker. They will appear as a page item on the page just like embeddables, but will have a different color. Only one TE side tip can be added per page. TE side tips are not copyable, but they can be moved.

Also included is a fix for making the TE authoring forms' Cancel button close the item edit dialog, and some minor UI tweaks including changing the "Authored State (Admin Only)" tab to simply say "Admin".


knowuh (approved):
Looks Great, thank you!

PR #962, opened 2022-06-15
Add import of glossaries when importing activities [#182476610]

dougmartin:
This matches the new glossary model based on the id in the new glossary model api endpoint.  This ignores the domain of the api endpoint so that we can import across domains during the LARA 2 testing and cutover.


dougmartin:
@scytacki I've added an environment variable (`ENABLE_DANGEROUS_GLOSSARY_LINKING_ON_IMPORT`) that when enabled allows the glossary linking based solely on `id`.  I think this is a better solution that whitelisting domains.  Here are the commits:

- https://github.com/concord-consortium/lara/pull/962/commits/b1a8c96433f1c9ae7202e3188acbd84f016ae53b
- https://github.com/concord-consortium/lara/pull/962/commits/9a6a27aad020a04985b1a65c5b3efa84268d1518

If this looks good I'll update the LARA CloudFormation template to include the parameter, defaulted to "false".

scytacki (approved):
Nice idea. Looks good to me.

PR #967, opened 2022-06-22
Bug fixes for Teacher Edition in LARA2

emcelroy:
PT Stories:

https://www.pivotaltracker.com/story/show/182276802
https://www.pivotaltracker.com/story/show/182276917

[#182276802]
[#182276917]

* Adds disabled state for the question wrapper plugins drop-down menu when all available options have been added to a particular question.
* Fixes issues around plugin script initialization that cause authoring form to crash or otherwise behave strangely.
* Fixes an issue where the names of properties under an item's `data` property are _sometimes_ not converted from snake case to camel case. This fix is meant to be a placeholder solution as there's obviously a problem elsewhere that needs to be fixed. I didn't have enough time to find the root cause.


knowuh (approved):
Looks good, thank you!

PR #968, opened 2022-06-23
Simplify GH Codespaces setup

pjanik:
[#182524012]

I've updated Readme and .env file to take advantage of `CODESPACE_NAME` env variable available in Codespace container. That way we don't have to start the server and make host public to guess the domain. Apparently, it's always `${CODESPACE_NAME}-{port}.githubpreview.dev`. So, the only manual part of the setup here is to go to Portal container and copy its Codespace name, and then paste into .env.

I also thought about re-defining `REPORT_SERVICE_TOOL_ID` in docker-compose-gh-codespaces.yml, so it doesn't use `USER` variable.

In practice, the report tool id / source key will look like: `${CODESPACE_NAME}-3000.githubpreview.dev.codespace`, as the username is always `codespace`. If there's no advantage of having 100% real hostname as tool id / source key, I guess this is fine. It makes this custom gh-codespaces config simpler.

UPDATE: 
> If there's no advantage of having 100% real hostname as tool id / source key, I guess this is fine. It makes this custom gh-codespaces config simpler.

Heh, I've just noticed that when Activity Player generates student report (summary), it just uses LARA host. I guess it has no better way to do it. So, apparently there's an advantage is using host / domain as report service tool ID and I guess I should change it again. I'll do it a bit later.

### DIFF
--- a/.env-gh-codespaces-sample
+++ b/.env-gh-codespaces-sample
@@ -7,14 +7,15 @@
 #   You can learn more about this at the top of docker/dev/docker-compose-portal-proxy.yml
 COMPOSE_FILE=docker-compose.yml:docker/dev/docker-compose-gh-codespaces.yml:docker/dev/docker-compose-portal-proxy.yml
 
-# To connect with the portal LARA needs to know the domain of the Portal. GitHub Codespaces will assign a random
-# host name that can be obtained after the port is made public in the Visual Studio Code "Ports" tab.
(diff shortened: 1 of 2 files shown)


scytacki (approved):
Looks good to me.
Yes the username is not necessary in the sourceKey. When it is used it has to be configured in AP url so it knows the right sourceKey to use.

PR #972, opened 2022-06-28
Updates to rewrite_lara_portal_resources.rb script

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/182587764

[#182587764]

* Adds option to omit `learner_processing_event` table from processing
* Removes `launch_url` from list of external_activities fields to update since it no longer exists
* Adds a method for updating `teacher_resources_urls` in the external_activities table since those often point to Portal pages.


scytacki (approved):
seems fine to me

PR #976, opened 2022-07-05
Migration script to convert all activities to sections schema

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/182518012

[#182518012]

This migration will update all existing activities to use the new sections schema. For each page of an activity, the migration creates new sections for any existing page items on the page, and associates those items with the new sections.

### DIFF
--- a/db/migrate/20220628163733_convert_activities_to_new_sections_schema.rb
+++ b/db/migrate/20220628163733_convert_activities_to_new_sections_schema.rb
@@ -1,0 +1,108 @@
+class ConvertActivitiesToNewSectionsSchema < ActiveRecord::Migration
+  class InteractivePage < ActiveRecord::Base
+  end
+
+  class PageItem < ActiveRecord::Base
+  end
+
+  def create_section(page_id, layout, position, show)
+    Section.create(
+      :interactive_page_id => page_id,
+      :layout => layout,
+      :position => position,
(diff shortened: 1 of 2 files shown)


knowuh (approved):
Looks good, thank you.

db/migrate/20220628163733_convert_activities_to_new_sections_schema.rb
```diff
@@ -0,0 +1,108 @@
+class ConvertActivitiesToNewSectionsSchema < ActiveRecord::Migration
+  class InteractivePage < ActiveRecord::Base
+  end
+
+  class PageItem < ActiveRecord::Base
+  end
+
+  def create_section(page_id, layout, position, show)
+    Section.create(
```
scytacki, line 9: Probably this should be added to the list of inline active record objects above like:
```
class Section < ActiveRecord::Base
end
```

scytacki (changes requested):
Because it sounds like this will take a while to run, I suggest adding some logging to this. So we can tell that it is making progress.
Perhaps first printing how many pages to process.
And then every batch of 10 or 100 printing how many pages have been processed like: `[processed]/[total]`

scytacki (approved):
Looks good to me. 👍

PR #977, opened 2022-07-05
Disables LARA as a runtime environment and defaults it Activity Player

eireland:


emcelroy (changes requested):
This looks good except you'll need to make similar changes to the sequence forms.

You don't need to do this, but also it may be better to simply disable the select menu altogether (leaving the default selection as Activity Player). You might also change the note below the select menu to say something like "Runtime selection is no longer applicable. This setting will be removed in a future release."

emcelroy (approved):
Looks good! Thanks for fixing that Defunct checkbox layout mistake I made.

PR #982, opened 2022-07-08
Interactive API: Report Item API updates + small cleanup

pjanik:
I think @scytacki would be the best reviewer as we discussed most of that in person already. @dougmartin, I'm marking you too, as you were involved in working on this API before.

Changes:
1. `notifyReportItemClientReady` function added to `api.ts`. It's not only wrapping raw post call, but also lets client provide metadata. This new metadata has only one property for now: `compactAnswerReportItemsAvailable`. When it's set to `true`, Portal Dashboard will ask report item iframe for `compactAnswer` report item types.
2. Multiple changes in `types.ts`. Metadata, new report item type ("score"). Also, report item requests and responses can specify `itemsType` which is either `fullAnswer` (or undefined, so we maintain compatibility with v2.0.0) or `compactAnswer`. The only supported compact answer item is "score" (for now). Unfortunately, it's not clearly visible in types. I've also added a "hidden" view type, so interactive has a hint that it doesn't need to render itself (compact answers case).
3. There's a new hook `useReportState` which should simply implementation a bit, as interactive doesn't have to manually notify the report that it's ready.
4. I've added a few tests related to report item API.
5. Also, the last commit removes some logic from `client.ts` and moves it to `api.ts`. I think it doesn't make sense to double most of these functions. I guess it'd be nice if the client stays as small as possible, and we keep more specific logic in separate functions in `api.ts` and/or `hooks.ts`.


scytacki (approved):
Looks good to me.

pjanik:
I've just updated version numbers and published client and host packages to NPM.

PR #984, opened 2022-07-11
Add legacy_ref_id field to managed interactives.

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/182657208

[#182657208]

When a built-in (aka LARA-native) question is converted to a managed interactive the built-in question's `ref_id` will be stored in this new `legacy_ref_id field`.

### DIFF
--- a/app/services/lara_serialization_helper.rb
+++ b/app/services/lara_serialization_helper.rb
@@ -8,6 +8,7 @@ class LaraSerializationHelper
     results = item.export
     results[:type] = item.class.name
     results[:ref_id] = key(item)
+    results[:legacy_ref_id] = item.legacy_ref_id if item.respond_to?(:legacy_ref_id)
     if item.respond_to?(:interactive) && item.interactive
       results[:interactive_ref_id] = key(item.interactive)
     end
--- a/app/services/lara_serialization_helper.rb
+++ b/app/services/lara_serialization_helper.rb
@@ -8,7 +8,6 @@ class LaraSerializationHelper
     results = item.export
     results[:type] = item.class.name
     results[:ref_id] = key(item)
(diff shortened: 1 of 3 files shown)


app/services/lara_serialization_helper.rb
```diff
@@ -8,6 +8,7 @@ def export(item)
     results = item.export
     results[:type] = item.class.name
     results[:ref_id] = key(item)
+    results[:legacy_ref_id] = item.legacy_ref_id if item.respond_to?(:legacy_ref_id)
```
scytacki, line 11: I don't think we want to export the legacy_ref_id. 
If it gets imported into the same or different lara it would create an interactive that looks like it was converted, but really it was just exported and imported.

scytacki (changes requested):
I don't think it should be exported.

app/services/lara_serialization_helper.rb
emcelroy: @scytacki OK. I'm confused though on how else to make it accessible to Activity Player. Or am I wrong about AP needing to access this value?

If it does need to stay here, maybe I should've also added the field to the list of excepted fields on line 42 of the import method?

emcelroy: After thinking about this some more, it seems like there's no need to make that value accessible to Activity Player... I must've just got mixed up somehow about that. I'll remove that line from lara_serialization_helper.rb.

scytacki: Right, it shouldn't be needed by the AP. It is there to help us when we migrate the learner answers. And it will be useful if we have to track down old data if something goes wrong with the migration.

scytacki (approved):
Looks good to me.

Only issue I see is the merge conflict with the schema.rb
When you fix that go ahead and merge it.

PR #985, opened 2022-07-11
Add migration_status field to activities and sequences

emcelroy:
PT Story: https://www.pivotaltracker.com/n/projects/2441249/stories/182666056

[#182666056]

Adds a `migration_status` field to lightweight activities and sequences. The field will be used to mark resources as changed during the migration process.


scytacki (approved):
looks good to me.

PR #988, opened 2022-07-12
Add "official" flag to library interactives

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/182702184

[#182702184]

Adds an `official` field to library interactives. Official library interactives are the most recent, stable versions and should be used in most production activities.

The note below the field's checkbox in the form is kind of a placeholder. It should be updated when we get to the work of limiting authors to use only official library interactives.

For convenience, I added an "OFFICIAL" text label to an official library interactive's listing on the main library interactives page. We should probably update the styling of that, but for now I just put something simple.


dougmartin (approved):
Looks good :+1:

PR #989, opened 2022-07-18
forces height and width for images in authoring preview

eireland:


emcelroy (commented):
Is there a .scss file that should be included in these changes?

emcelroy (approved):
Looks good, thanks.

PR #990, opened 2022-07-18
Changes the width of the authoring panel in the authoring modal

eireland:
Changes the width of the authoring panel in the autoring modal because:
1.  It is hiding some of the RTE buttons
2. image editing dialog in subsequent question area get cutoff.


emcelroy (approved):
Looks good 👍

PR #993, opened 2022-07-18
Create script to convert LARA runtime activities to AP activities.

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/179855280

[#179855280]

* New `legacy_ref_id` property for ManagedInteractive
* Rake tasks for creating and managing ManagedInteractive versions of old, built-in embeddables.

### DIFF
--- a/lib/tasks/convert_embeddables.rake
+++ b/lib/tasks/convert_embeddables.rake
@@ -22,17 +22,15 @@ def count_embeddables()
 end
 
 def multiple_choice_authored_state(item)
-  choice_id = 1
   converted_choices = []
   item.embeddable.choices.each do |choice|
     converted_choice = {
-      id: choice_id,
+      id: choice.id,
       content: choice.choice,
       correct: choice.is_correct,
       choiceFeedback: choice.prompt
     }
     converted_choices.push(converted_choice)
-    choice_id += 1
   end
 
   layout = item.embeddable.show_as_menu ? "Dropdown" : item.embeddable.layout
@@ -248,6 +246,7 @@ def create_new_embeddables(activity_id=nil)
     end
     # This requires the new migration_status field to be added to LightweightActivity
     # activity.migration_status = "in progress"
+    # activity.save!
   end
   puts "Processed #{activities_processed} of #{activities_to_process} activities."
 end
@@ -328,6 +327,7 @@ def replace_old_embeddables(activity_id=nil)
     activity.runtime = "Activity Player"
     # This requires the new migration_status field to be added to LightweightActivity
     # activity.migration_status = "migrated"
+    activity.save!
   end
   puts "Processed #{activities_processed} of #{activities_to_process} activities."
 
--- a/lib/tasks/convert_embeddables.rake
+++ b/lib/tasks/convert_embeddables.rake
@@ -192,6 +192,9 @@ def create_new_embeddables()
         authored_state = nil
         if item.embeddable
           legacy_ref_id = (LaraSerializationHelper.new).key(item.embeddable)
+          if ManagedInteractive.where(:legacy_ref_id => legacy_ref_id).count > 0
+            return
+          end
 
           case item.embeddable_type
           when "Embeddable::Answer", "Embeddable::AnswerFinder", "Embeddable::FeedbackFunctionality", "Embeddable::FeedbackItem", "Embeddable::LabbookAnswer", "Embeddable::Labbook"
--- a/lib/tasks/convert_embeddables.rake
+++ b/lib/tasks/convert_embeddables.rake
@@ -15,16 +15,14 @@
   "Embeddable::ImageQuestion"
 ]
 
-not_convertable_question_types = []
-
 def count_embeddables()
   # Get a count of all built-in embeddables that need to be converted.
   embeddables_count = PageItem.where(:embeddable_type => @convertable_embeddable_classes).length
   puts "There are #{embeddables_count} built-in embeddables to convert."
 end
 
-def convert_multiple_choice(item, library_interactive)
-  name = item.embeddable.name != "Multiple Choice Question element" ? item.embeddable.name : ""
+def multiple_choice_authored_state(item)
+  
   choice_id = 1
   converted_choices = []
   item.embeddable.choices.each do |choice|
@@ -39,7 +37,6 @@ def convert_multiple_choice(item, library_interactive)
   end
 
   layout = item.embeddable.show_as_menu ? "Dropdown" : item.embeddable.layout
-  
   authored_state = {
     version: 1,
     questionType: "multiple_choice",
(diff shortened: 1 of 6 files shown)


lib/tasks/convert_embeddables.rake
```diff
@@ -0,0 +1,326 @@
...
+        library_interactives[question_type] = library_interactive
+      end
+    end
+  end
+
+  if library_interactives.length == 0
+    raise Exception.new "No official library interactives found."
+  end
+
+  # activities_to_update = LightweightActivity.where(:runtime => "LARA")
```
scytacki, line 217: We might get a small performance benefit by iterating over the PageItems with a specific embeddable type. This way a smaller bit of code is going run again and again during the loop, so the ruby interpreter might optimize it better.  But I think it is best to worry about that later.

scytacki (changes requested):
This looks good to me.

From what I understand I think you got the image question correct.
The only thing I think you need to fix is the count_embeddables function, and maybe the custom_native_width, custom_native_height.

Also, if we really need to speed this up I think we could craft some custom SQL which uses the SQL `concat` function to generate the authored state directly in SQL instead of going out to ruby.  But I don't think it is worth doing that until we know how long it will take to run this current version.

Also before we optimize this we should:
- see what the second step looks like
- see what it would take to support running this script multiple times: once before the downtime and once during the downtime

lib/tasks/convert_embeddables.rake
```diff
@@ -0,0 +1,326 @@
...
+  "VideoInteractive",
+  "Embeddable::ImageQuestion"
+]
+
+not_convertable_question_types = []
+
+def count_embeddables()
+  # Get a count of all built-in embeddables that need to be converted.
+  embeddables_count = 0
+  PageItem.all.each do |item|
```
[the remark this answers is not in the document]
emcelroy: OK, I've made that change.

lib/tasks/convert_embeddables.rake
```diff
@@ -0,0 +1,326 @@
...
+end
+
+def new_properties(item, library_interactive)
+  new_properties = {
+    library_interactive_id: library_interactive.id,
+    url_fragment: nil,
+    inherit_aspect_ratio_method:	true,
+    custom_aspect_ratio_method: nil,
+    inherit_native_width: true,
+    custom_native_width: library_interactive.native_width,
```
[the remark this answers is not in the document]
emcelroy: Right. Looks like I made this mistake in the original JS converter and carried it over to here. 

Does the latest change make more sense?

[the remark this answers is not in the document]
scytacki: This still seems wrong to me. I'd guess none of the newly created interactives should have a custom native width. They should always inherit it from the library interactive. If that is right then you could make it simple like:
```
inherit_native_width: true,
custom_native_width: nil,
```

scytacki (changes requested):
I think the native_width is still wrong, I suggested what I think is right, but I might be missing something.

lib/tasks/convert_embeddables.rake
```diff
@@ -205,98 +176,131 @@ def convert()
...
-      # be associated with the activity/page. That will be done by a different function.
+      # embeddables. Set the new ManagedInteracative's legacy_ref_id to be the ref_id of the 
+      # built-in embeddable.
       page.page_items.each do |item|
-        if item.embeddable # not sure what page items don't have embeddables - plugins?
+        library_interactive = nil
+        authored_state = nil
+        if item.embeddable
           legacy_ref_id = (LaraSerializationHelper.new).key(item.embeddable)
+          if ManagedInteractive.where(:legacy_ref_id => legacy_ref_id).count > 0
```
scytacki, line 195: I'd guess this would result in about 1 million queries on production. I'd try replacing it with something like an outer join statement so you include the managed interactive (if there is one) for each of the items. 

But to start with I think it would be fine to just remove this. To prevent duplicates if we run this again, we could just delete all of the ManagedInteractives that have a legacy_ref_id before running the loop.

lib/tasks/convert_embeddables.rake
```diff
@@ -205,98 +176,131 @@ def convert()
...
-      # old, built-in embeddable. For now, the new ManagedInteractive embeddable will not 
-      # be associated with the activity/page. That will be done by a different function.
+      # embeddables. Set the new ManagedInteracative's legacy_ref_id to be the ref_id of the 
+      # built-in embeddable.
       page.page_items.each do |item|
-        if item.embeddable # not sure what page items don't have embeddables - plugins?
+        library_interactive = nil
+        authored_state = nil
+        if item.embeddable
           legacy_ref_id = (LaraSerializationHelper.new).key(item.embeddable)
```
scytacki, line 194: This approach to computing the legacy_ref_id seems in efficient since it is creating a new LaraSerializationHelper for each embeddable. Could you re-use the same serialization helper object?

lib/tasks/convert_embeddables.rake
```diff
@@ -205,98 +176,131 @@ def convert()
...
-        puts "page_item: #{page_item.inspect}"
-        # TODO: Update the page item's embeddable to be the ManagedInteractive embeddable.
+  # To replace the old, built-in embeddables, first we find any ManagedInteractive embeddables
+  # that have a legacy_ref_id value and are not yet associated with a PageItem. Then, we use 
+  # each of these ManagedInteractive embeddables' legacy_ref_id to find the old embeddable it 
+  # will replace and the PageItem that it's associated with. We replace the PageItem's old 
+  # embeddable with the new embeddable. We also update any wrapping plugins for the old embeddable
+  # so they point to the new embeddable. Finally, delete the old embeddable.
+  new_embeddables = ManagedInteractive.where("legacy_ref_id IS NOT NULL AND legacy_ref_id != ?", "")
+  new_embeddables.each do |new_embeddable|
```
scytacki, line 286: This should be `find_each` since I expect there to be about million new_embeddables

lib/tasks/convert_embeddables.rake
```diff
@@ -205,98 +176,131 @@ def convert()
...
+    if PageItem.where(:embeddable_id => new_embeddable.id, :embeddable_type => "ManagedInteractive").count == 0
```
scytacki, line 287: This will result in a lot of extra queries, so if there is a way to include the page_item via a join during the ManagedInteractive query above that will be a more efficient. If ManagedInteractive has a relation to PageItem, I think just adding a `.includes(:page_item)` to the query above. And then you can add another where condition on the missing  of the page_item_id. This SO answer provides more info: https://stackoverflow.com/a/5570221. If this relation doesn't exist on ManageInteractive then you could just add it to make this work.

lib/tasks/convert_embeddables.rake
```diff
@@ -205,98 +176,131 @@ def convert()
...
+      old_embeddable_id, old_embeddable_type = new_embeddable.legacy_ref_id.split("-")
+      old_embeddable = old_embeddable_type.constantize.find(old_embeddable_id)
+      page_items = PageItem.where(:embeddable_id => old_embeddable.id, :embeddable_type => old_embeddable_type)
```
scytacki, line 290: This will also result in a lot of queries, but I don't see a way to avoid that unless we stored additional fields on the ManagedInteractive that made it possible possible to do a SQL join with the legacy page item. If this is too slow it might be worth doing that.

lib/tasks/convert_embeddables.rake
```diff
@@ -205,98 +176,131 @@ def convert()
...
+      page_items.each do |page_item|
+        page_item.embeddable = new_embeddable
+        page_item.save!
+        wrapping_plugins = Embeddable::EmbeddablePlugin.where(:embeddable_id => old_embeddable.id, :embeddable_type => old_embeddable_type)
```
scytacki, line 294: This could be done with a join especially if the PageItem has a has_many relationship to EmbeddablePlugin via something like a wrapping_plugins.  Then in the PageItem query above on line 290 you could add a `.includes(wrapping_plugins)`. But I think that will only really help if the page item query on 290 was actually finding more than one PageItem at a time.

lib/tasks/convert_embeddables.rake
```diff
@@ -205,98 +176,131 @@ def convert()
...
+          wrapping_plugin.embeddable = new_embeddable
+          wrapping_plugin.save!
+        end
+        old_embeddable.destroy
       end
     end
   end
-  # TODO: Make sure TE question wrappers are updated to point at new ManagedInteractive.
-
+  # TODO: Set associated activities' runtime to "Activity Player" after all old embeddables are replaced?
```
scytacki, line 303: Yes and also update the migration status field. You might want to update this field to indicate the conversion of the activity is in progress and then update it again when the activity has been fully converted.

scytacki (changes requested):
I found several places where this could be more efficient. I don't think you need to do them all. 

Definitely the `find_each` changes should be done.

And the approach to find ManagedInteractives without page items also seems not too hard and worthwhile.

scytacki (approved):
This looks good to me.

I can see more optimizations now, but I think we should try running it like this and see how long the 2 parts take.

PR #995, opened 2022-07-19
Modify export to report so all questions are included

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/182762725

[#182762725]

Replace the use of `section.title` with `section.id` when retrieving all embeddables from a specific section. (Currently, all sections have their title set to the same value: `assessment_block`.)

### DIFF
--- a/app/models/interactive_page.rb
+++ b/app/models/interactive_page.rb
@@ -123,7 +123,7 @@ class InteractivePage < ActiveRecord::Base
     # this method, but this legacy feature will be removed in the future and 
     # passing the section itself will become the only option.
     if !section.instance_of? Section
-      section = sections.find { |s| s.id == section_id }
+      section = sections.find { |s| s.title == section_title }
     end
 
     if section
--- a/app/models/interactive_page.rb
+++ b/app/models/interactive_page.rb
@@ -123,19 +123,31 @@ class InteractivePage < ActiveRecord::Base
     # this method, but this legacy feature will be removed in the future and 
     # passing the section itself will become the only option.
     if !section.instance_of? Section
-      section = sections.find { |s| s.title == section_title }
+      section = sections.find { |s| s.title == section }
     end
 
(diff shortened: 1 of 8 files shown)


app/models/interactive_page.rb
```diff
@@ -118,24 +118,19 @@ def interactive_page_items
     page_items.select{ |pi| Embeddable::is_interactive?(pi.embeddable) }
   end
 
-  def section_embeddables(section_title)
-    section = sections.find { |s| s.title == section_title }
+  def section_embeddables(section_id)
+    section = sections.find { |s| s.id == section_id }
```
scytacki, line 122: How about just passing the section object itself?
Since we are getting rid of the title lookup, visible_embeddables knows the section object, this would prevent the extra `find` call here.

scytacki (changes requested):
Seems like it could be optimized, but maybe I'm missing some usage of this where it is necessary to pass the id.

Since this change is breaking several of the other method in this model I think you should at least put comments on them indicating they are broken. I agree it is good to not remove them since having them there will make it easier to track down what is using them by searching for their names.

app/models/interactive_page.rb
```diff
@@ -118,24 +118,37 @@ def interactive_page_items
...
+      end
+    end
+    section_embeddables
   end
 
   def visible_embeddables
     results = []
     sections.each do |s|
-      results += section_visible_embeddables(s.title)
+      puts "LKLKLKLK 8 #{s.inspect}"
```
scytacki, line 150: probably should remove this debugging statement

scytacki (changes requested):

app/models/interactive_page.rb
emcelroy: Whoops. Yeah, probably :)

scytacki (approved):
Looks good to me

PR #996, opened 2022-07-20
Fix incorrect layout strings in migration to convert activities to new sections schema.

emcelroy:
PT Stories:

https://www.pivotaltracker.com/story/show/182518012
https://www.pivotaltracker.com/story/show/182657008

[#182518012]
[#182657008]

The incorrect layout strings, l-4060 and l-3070, resulted in some sections' layouts being set to `nil` during the migration.


scytacki (approved):
looks good to me

PR #998, opened 2022-07-20
Added legacy_glossary_resource_id [#182761599]

dougmartin:


scytacki (approved):
Looks good to me.

I'm guessing you'll start a new PR for the rake task to update activities using a legacy glossary to ones using these new built in glossaries.

Maybe you want to expose this legacy glossary id in the UI of the glossary editing?  Doesn't have to be part of this PR or even done at all, but it might make it easier to confirm they've been setup correctly on lara2, and also that they were imported correct on other lara instances.

dougmartin:
@scytacki Yes, there is a follow on PR with the migration tasks - I thought it would be cleaner that way.

I thought we could just do a query after the update.  Instead I'll add to the new rake tasks a task that lists glossaries with glossary legacy ids.

PR #1000, opened 2022-07-21
Reinstate removed code to fix broken activity edit form.

emcelroy:
Removing the `editor_mode` code in this PR https://github.com/concord-consortium/lara/pull/995 broke the activity edit form in that the publish box no longer appears.


scytacki (approved):
seems fine to me

PR #971, opened 2022-06-27
Add defunct and migration status fields for deprecated items

emcelroy:
PT Stories:

https://www.pivotaltracker.com/story/show/182465001
https://www.pivotaltracker.com/story/show/182577639 

[#182465001]
[#182577639]

Adds a `defunct` boolean field to activities and sequences to use for identifying resources containing deprecated features.

Adds a `migration_status` string field to LARA-native multiple-choice, open response, image, image interactive, and video embeddables to use in identifying them as having been migrated. Default value is "not migrated".

PR #1001, opened 2022-07-21
Add legacy_ref_type field to managed interactives

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/182786659

[#182786659]

We were originally going to only add a `legacy_ref_id` field to managed interactives which would contain strings that look like `30-MultipleChoice` where "30" is the related built-in embeddable's numeric ID and "MultipleChoice" is the built-in embeddable's type. 

This change will put the two values in separate fields to provide greater flexibility.

Note: The `legacy_ref_id` migration was originally added to the `new-sections` branch of LARA (https://github.com/concord-consortium/lara/pull/984). I copied that migration file to this branch off of `master` since it turns out we are going to need to reference these values before upgrading to LARA 2. So long as the migration file name has the same ID in it (20220711213558) it won't be run again after we merge `new-sections` into `master`. That's why there are two separate migration files in this commit. In the second, we modify the type of `legacy_ref_id` from string to integer, as well as add the `legacy_ref_type` string field.

### DIFF
--- a/db/migrate/20220711213558_add_legacy_ref_id_to_managed_interactives.rb
+++ b/db/migrate/20220711213558_add_legacy_ref_id_to_managed_interactives.rb
@@ -1,0 +1,5 @@
+class AddLegacyRefIdToManagedInteractives < ActiveRecord::Migration
+  def change
+    add_column :managed_interactives, :legacy_ref_id, :string
+  end
+end
--- a/db/migrate/20220711222748_add_migration_status_to_lightweight_activities_and_sequences.rb
+++ b/db/migrate/20220711222748_add_migration_status_to_lightweight_activities_and_sequences.rb
@@ -1,0 +1,6 @@
+class AddMigrationStatusToLightweightActivitiesAndSequences < ActiveRecord::Migration
+  def change
+    add_column :lightweight_activities, :migration_status, :string, default: 'not_migrated'
+    add_column :sequences, :migration_status, :string, default: 'not_migrated'
+  end
+end
--- a/db/migrate/20220721203706_add_legacy_ref_type_to_managed_interactives.rb
+++ b/db/migrate/20220721203706_add_legacy_ref_type_to_managed_interactives.rb
@@ -1,0 +1,10 @@
+class AddLegacyRefTypeToManagedInteractives < ActiveRecord::Migration
+  def up
+    change_column :managed_interactives, :legacy_ref_id, :integer
+    add_column :managed_interactives, :legacy_ref_type, :string
+  end
+  def down
+    change_column :managed_interactives, :legacy_ref_id, :string
+    remove_column :managed_interactives, :legacy_ref_type
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended to check this file into your version control system.
 
-ActiveRecord::Schema.define(:version => 20220712165650) do
+ActiveRecord::Schema.define(:version => 20220721203706) do
 
   create_table "admin_events", :force => true do |t|
     t.string   "kind"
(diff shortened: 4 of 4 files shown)


emcelroy:
@scytacki Thinking about this some more, it seems like we'll also need to copy over the `migration_status` migrations from `new-sections` if the conversion rake task on `master` is going to be updating those fields. That is, if we're going to be setting them to "in progress" before the final pass. But it would also make testing the rake task easier.

Should I go ahead and copy those migration files to this PR?

https://github.com/concord-consortium/lara/pull/971
https://github.com/concord-consortium/lara/pull/985

pjanik (commented):
Personally, I'm not sure about using number as `legacy_ref_id`. I think multiple codebases assume that ref_id is a combination of a numeric ID and some class name. Especially when it's called `ref_id`, not just `id`. There's already a small confusion, as sometimes we call it `ref_id` (LARA), sometimes just plain `id` (activity resource json in Firestore), and sometimes we use `<ID>-CamelCaseClass` (LARA) and sometimes `snake_case_class_<ID>` (activity resource json in Firestore). :) This would add one more division/variant - if we deal with legacy `ref_id`, then it's split into separate number and type. And it won't be only in LARA, as we'll need that legacy_ref_id in Firestore too (probably converted to snake case format and named `legacy_id` to be consistent with other fields up there).

But, this is just my general impression. Probably I don't know all the details of this migration work. I bet @scytacki might have a better view.

emcelroy:
> Personally, I'm not sure about using number as `legacy_ref_id`. I think multiple codebases assume that ref_id is a combination of a numeric ID and some class name. Especially when it's called `ref_id`, not just `id`. There's already a small confusion, as sometimes we call it `ref_id` (LARA), sometimes just plain `id` (activity resource json in Firestore), and sometimes we use `<ID>-CamelCaseClass` (LARA) and sometimes `snake_case_class_<ID>` (activity resource json in Firestore). :) This would add one more division/variant - if we deal with legacy `ref_id`, then it's split into separate number and type. And it won't be only in LARA, as we'll need that legacy_ref_id in Firestore too (probably converted to snake case format and named `legacy_id` to be consistent with other fields up there).
> 

Would it make more sense to name the fields `legacy_id` and `legacy_type`? If so, would it still be better to use string for the `legacy_id` field type?

pjanik:
> Would it make more sense to name the fields legacy_id and legacy_type? If so, would it still be better to use string for the legacy_id field type?

I think it depends on where.  Published activity structure in portal-report Firestore doesn't use "ref" word. There are only IDs. So the published property should be probably "legacy_id".

LARA itself uses "ref" word, so I'd probably call it "legacy_ref_id" within LARA codebase and its Activity Export JSON.

But it's just personal opinion to keep things consistent with existing naming (hopefully, I'm not 100% sure about LARA).

Re division of the ID into ID and Type - still, personally I wouldn't do it. I don't see a benefit and it can just become more confusing to work with more rules around these properties. And I think we initially combined ID with a class name to ensure that each element in a set of mixed answers/questions types still has a unique ID. This will become less important with "legacy_" props, but I'd be consistent.

scytacki:
@pjanik the reason to split the id from the type in the database is so that we have the option to use more efficient SQL queries to join the legacy embeddables with the new ManagedInteractives in the second step of the AP conversion. I don't know if we'll have time to implement that improved SQL, but it is good to have the option.

scytacki:
I don't care about the names of the fields since they are temporary. As long as the prefix is the same and end is `_id` and `_type` then Rail's polymorphic relationship code should be able to handle it automatically if we try to use that.

I'd just stick with the `legacy_ref` prefix to save Ethan a little time.

@emcelroy yes go ahead and copy over those migrations too.
And then once this is PR is merged with master, you probably want to merge master back into `new-sections`. It might show a conflict because a few files will be duplicated, but I think if the content is identical in each file git will just go ahead with the merge.

emcelroy:
@scytacki @pjanik Sorry but I got mixed up about those other two migrations. One of them is _already_ on `master`. I added the other one. Does it all look acceptable now?

pjanik:
It looks good to me. @scytacki's point about the reasoning for division makes sense. And I can use migration_status in my queries (assuming it gets published to Firestore later). Thanks.

PR #1007, opened 2022-07-26
Script to convert all LARA runtime activities to AP runtime activities.

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/179855280

[#179855280]

Change `replace_old_embeddables` to iterate over new managed interactives instead of page items.

### DIFF
--- a/app/models/managed_interactive.rb
+++ b/app/models/managed_interactive.rb
@@ -45,6 +45,7 @@ class ManagedInteractive < ActiveRecord::Base
   has_one :labbook, :as => :interactive, :class_name => 'Embeddable::Labbook'
 
   belongs_to :linked_interactive, :polymorphic => true
+  belongs_to :converted_interactive, :polymorphic => true, :foreign_key => :legacy_ref_id, :foreign_type => :legacy_ref_type
 
   # getter for constructed url
   def url
--- a/lib/tasks/convert_embeddables.rake
+++ b/lib/tasks/convert_embeddables.rake
@@ -305,26 +305,29 @@ end
 def replace_old_embeddables(activity_id=nil)
   serialization_helper = LaraSerializationHelper.new
   where_conditions = activity_id ? {id: activity_id} : {runtime: "LARA"}
-  page_items = PageItem.includes(interactive_page: {lightweight_activity: []}).where(embeddable_type: @convertable_embeddable_classes).where(lightweight_activities: where_conditions)
-  page_items_processed = 0
-  page_items_to_process = page_items.count
-  page_items.find_each(batch_size: 100) do |page_item|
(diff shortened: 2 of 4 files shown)


lib/tasks/convert_embeddables.rake
```diff
@@ -303,28 +303,31 @@ def delete_orphaned_old_embeddables()
...
-      wrapping_plugins = Embeddable::EmbeddablePlugin.where(embeddable_id: new_embeddable.legacy_ref_id, embeddable_type: new_embeddable.legacy_ref_type)
-      wrapping_plugins.each do |wrapping_plugin|
-        wrapping_plugin.embeddable = new_embeddable
-        wrapping_plugin.save!
-      end
-      page_items_processed += 1
-      if page_items_processed % 100 == 0
-        puts "Processed #{page_items_processed} page items."
-      end
+  # TODO: Preload wrapping plugins in main query before looping through embeddables.
```
scytacki, line 306: Something you can try is to add:
`has_many :embeddable_plugins, as: :embeddable`
To each of the legacy embeddables.
Then you might be able to do:
`new_embeddables = ManagedInteractive.includes(:page_item).preload(converted_plugins: [], converted_interactive: [:embeddable_plugins, {page_items: {interactive_page: [:lightweight_activity]}}]}).where("legacy_ref_id IS NOT NULL AND legacy_ref_id != ?", "").where("page_items.id IS NULL")`

app/models/managed_interactive.rb
```diff
@@ -45,6 +45,7 @@ class ManagedInteractive < ActiveRecord::Base
   has_one :labbook, :as => :interactive, :class_name => 'Embeddable::Labbook'
 
   belongs_to :linked_interactive, :polymorphic => true
+  belongs_to :converted_interactive, :polymorphic => true, :foreign_key => :legacy_ref_id, :foreign_type => :legacy_ref_type
```
scytacki, line 48: I'd name this `legacy_embeddable`. It is the converted_interactive that has a legacy embeddable. 

You could possibly add the inversion of this of on each of the legacy embeddables like
`has_one :converted_interactive, as: :legacy_embeddable`

With that inverted relationship, it might be possible to go back to iterating over activities and use an `includes` that would load the converted_interactive of every embeddable in the activity. But I'm not entirely sure it would work.  Without going back to that I don't know an efficient way tell when all of the items of an activity have been re-wired. My best guess would be to use some grouping by activity, but I don't really know how to do that.

scytacki (changes requested):
I left a few comments, but I'm guessing they aren't quite right. I should be available tomorrow, to pair on this.

scytacki (approved):
Looks good to me.

PR #1008, opened 2022-07-26
[#179237412] Importing into LARA2 looks for embeddables within sections.

lublagg:


app/models/interactive_page.rb
```diff
@@ -375,16 +375,18 @@ def self.import(page_json_object, helper=nil)
...
-      end
-      # Now when all the objects are created, setup references (e.g. question pointing to interactive, or
-      # one embeddable pointing to another one).
-      page_json_object[:embeddables].each do |embed_hash|
-        helper.set_references(embed_hash[:embeddable])
+      page_json_object[:sections].each do |s|
+        # First, import and cache all the embeddables.
+        s[:embeddables].each do |embed_json_obj|
+          embed = helper.import(embed_json_obj.except(:column, :position))
+          import_page.add_embeddable(embed, embed_json_obj[:position], s[:title], embed_json_obj[:column])
```
scytacki, line 382: Do all sections have unique titles in LARA2?
I'd guess that if you make a page from scratch in LARA2 the sections will not have titles. So then in this case I think all embeddables will get added to the same first section.

Can you test this making a page from scratch with 2 sections and embeddables in each section, then export and import back to LARA2.  If that works I must be missing something.

scytacki (changes requested):
I don't think this will work for title-less sections.

dougmartin:
@scytacki I've added version specific import in this commit:  https://github.com/concord-consortium/lara/pull/1008/commits/9a5ce03fe0dda3d0776722277787e6cc5a8a4894

app/models/interactive_page.rb
```diff
@@ -375,19 +377,46 @@ def self.import(page_json_object, helper=nil)
...
-          helper.set_references(embed_json_obj)
+        page_json_object[:embeddables].each do |embed_hash|
+          helper.set_references(embed_hash[:embeddable])
+        end
+      elsif version == 2
+        page_json_object[:sections].each do |s|
+          # first create the section and add it to the page at the correct location
+          new_section = import_page.sections.create(Section::DEFAULT_PARAMS.merge({title: s[:title]}))
+          if s[:title] == Section::HEADER_BLOCK
+            new_section.move_to_top
```
scytacki, line 401: I'd recommend removing this. If we are importing a version 2 page, the sections should be ordered correctly in the json we are parsing. The titles shouldn't have any effect on the order in a version 2 page.

app/models/lightweight_activity.rb
```diff
@@ -210,6 +210,7 @@ def self.extract_from_hash(activity_json_object)
   end
 
   def self.import(activity_json_object,new_owner,imported_activity_url=nil,helper=nil)
+    version = activity_json_object[:version] || 1
```
scytacki, line 213: Seems like you need to do this on the sequence model too? Or is the version specified on each of the individual activities (I faintly remember there being something weird like that)

scytacki (changes requested):
I found one issue with how the header block title is handled in version 2. 
And another question about whether the sequence model needs to be updated.

app/models/lightweight_activity.rb
dougmartin: Yeah, unfortunately it is not set at the top level of the sequence but rather in each activity.

app/models/interactive_page.rb
dougmartin: I'll remove it.

dougmartin: Removed in https://github.com/concord-consortium/lara/pull/1008/commits/b788907652229ee237501d1131876db8aac31082

scytacki (approved):
Looks good.

PR #1014, opened 2022-07-29
Fix image question conversion

pjanik:
There were two issues with IQ conversion:
- `backgroundSource` was never set to "URL", even if the background URL was provided
- `backgroundSource` was sometimes set to `null`. This was working fine in runtime, but authoring was complaining. So, I'm not setting `backgroundSource` property at all when it's undefined.

I've tested all the Image Question variants and there's one combination that is not supported by the new version (as their authored state doesn't map 1:1). LARA IQ lets the author set source to "Upload" and display an authored URL background anyway. So, students would see this background as an initial prompt. Once they upload their own background, this authored background is never displayed anymore. It's also not mixed in any way with students' uploads.

The new IQ has a concept of "backgroundSource" instead of "source". So, once it's set to "Upload", there's no way to display authored background (it'd require backgroundSource = "url" instead of "upload"). 

I think this doesn't sound like a critical issue. Also, note that it just depends on Image Question runtime / authoring logic. The data will be there, so if someone is really missing this feature, we can implement this feature later in the Question Interactives repo and things will work as they used to.

### DIFF
--- a/lib/tasks/convert_embeddables.rake
+++ b/lib/tasks/convert_embeddables.rake
@@ -89,7 +89,17 @@ def image_authored_state(item)
 end
 
 def image_question_authored_state(item)
-  background_source = item.embeddable.bg_source == "Shutterbug" ? "snapshot" : item.embeddable.bg_source === "Upload" ? "upload" : nil
+  background_source =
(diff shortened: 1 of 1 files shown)


scytacki (approved):
Looks good to me

PR #1015, opened 2022-07-29
[#182432327]: Add URL redirect for AP activities and sequences in LARA

lublagg:


spec/features/run_complete_sequence_spec.rb
```diff
@@ -1,166 +1,169 @@
...
-            present: { mode:'teacher-edition' },
-            absent:  { foo:'bar' }
-          )
-        end
-      end
-
-    end
-  end
-end
+# require 'spec_helper'
```
dougmartin, line 1: @lublagg could you add a comment as to why this spec is disabled?  It will help future folks when they load this file.  Something like: 

`# NOTE: this spec is disabled because runs are no longer created when viewing activities, instead the user is redirected to the Activity Player`

dougmartin (changes requested):
Looks good :+1:   The comment just needs to be added to the spec as to why it is commented out.

spec/features/run_complete_sequence_spec.rb
lublagg: Will do!

dougmartin (approved):

dougmartin:
@lublagg I approved the update so I'm going to merge this.

PR #1020, opened 2022-07-31
Modify send resources function with option to send only portal resources

emcelroy:


lib/tasks/reporting.rake
```diff
@@ -32,15 +32,21 @@ namespace :reporting do
...
   end
 
-  def send_all_resources(resource_class, sender_class, opts={})
+  def send_resources(resource_class, sender_class, portal, opts={})
     index = 0
     error_count = 0
     success_count = 0
     start_time = Time.now
-    count = resource_class.count
+    # If portal is set, only send activities or sequences that are in the portal, otherwise send
```
dougmartin, line 40: This was a little confusing given the three places its called but I understand it now -- we only care about the portal param if it is a activity or requence so there is no change in the calls for student runs or anonymous runs.

dougmartin (approved):
Looks good :+1:

PR #1021, opened 2022-08-01
Filter library interactive list in authoring based on the official flag.

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/182702190

[#182702190]

With this change:

* Non-admin authors will only see official library interactives in the item picker
* Admins will see both official and non-official library interactives in the item picker


dougmartin (approved):
Looks good :+1:

PR #1022, opened 2022-08-02
Add UI for marking a section as hidden

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/182613081

[#182613081]

Adds a toggle option in authoring to set a section's `show` property. The `show` property is used by Activity Player to determine if a section should be shown. When section is hidden, its opacity is set to 50% in authoring.

### DIFF
--- a/lara-typescript/src/section-authoring/api/api-types.ts
+++ b/lara-typescript/src/section-authoring/api/api-types.ts
@@ -232,7 +232,7 @@ export interface ISection {
   items?: ISectionItem[];
 
   /**
-   * Items in this section
+   * Should the section be shown in runtime?
    */
   show?: boolean;
 }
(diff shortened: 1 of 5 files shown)


lara-typescript/src/section-authoring/api/api-types.ts
```diff
@@ -230,6 +230,11 @@ export interface ISection {
    * Items in this section
    */
   items?: ISectionItem[];
+
+  /**
+   * Items in this section
```
scytacki, line 235: Can you update this comment?

scytacki (approved):
Looks good to me, I just found one comment that should be updated.

lara-typescript/src/section-authoring/api/api-types.ts
emcelroy: Ah, thanks for catching that. It's fixed in the latest commit.

scytacki (approved):
Looks good.

PR #1023, opened 2022-08-03
Fixed "Edit null" showing in page editor when page name not set [#181889363]

dougmartin:
I know this isn't critical but it really bugs me so I fixed it in a couple of minutes.


scytacki (approved):
Looks good to me

PR #1024, opened 2022-08-03
Style hidden items differently than visible items in authoring

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/182896332

[#182896332]

If a page item is marked as hidden in the database, it will be displayed at 50% opacity and "HIDDEN - " will be prepended to its title.

In the screenshot below, the second item in the section is a hidden page item.

image


scytacki (approved):
looks good

PR #1026, opened 2022-08-16
Fix iframes not rendered in the authoring preview.

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/182908649

[#182908649]

Adds `iframe` to the default set of allowed HTML tags, and adds some related attributes to the set of allowed attributes. Also removes some unnecessary/unused imports from the text block preview component.


scytacki (approved):
looks good to me

PR #1027, opened 2022-08-17
Fix un-editable TE question wrappers

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/182893380

[#182893380]

Make sure we're getting the page item of type `Embeddable::EmbeddablePlugin` with the specified `embeddable_id`. Otherwise, we may be getting a page item of a different type.

### DIFF
--- a/app/controllers/api/v1/page_items_controller.rb
+++ b/app/controllers/api/v1/page_items_controller.rb
@@ -19,7 +19,7 @@ class Api::V1::PageItemsController < API::APIController
     plugins = []
     page_item_plugins = Embeddable::EmbeddablePlugin.where({embeddable_id: @page_item.embeddable_id, embeddable_type: @page_item.embeddable_type})
     page_item_plugins.each do |plugin|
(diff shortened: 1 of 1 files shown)


scytacki (changes requested):
This is great that you found it!

I was thinking you should convert both of these queries into associations on the models instead of doing them manually. But I looked up how to do the first one and it isn't easy. I think you'd have to do something like is described here: https://stackoverflow.com/a/2462397 to add a dynamic condition so you could use both the embeddable_id and embeddable_type in the association. 
And because that approach changes from Rails 3.1+ to 4+ it doesn't seem like a good idea to use it.  So I think the first query should just stay as is: 
`Embeddable::EmbeddablePlugin.where({embeddable_id: @page_item.embeddable_id, embeddable_type: @page_item.embeddable_type})`

The second query (the one you fixed) already has a relationship so it would be good to use it. You can replace the code you changed with `plugin_page_item = plugin.page_items.first`

scytacki (approved):
Looks good to me

PR #1028, opened 2022-08-19
Fix moving items.

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/182895947

[#182895947]

* Drop support for section ID in `add_embeddable` method. Use Section object instead.
* Only use `updateSection` on target section in `moveItem` function to avoid potentially conflicting changes to moved page items. 
* Update move dialog component to properly include secondary column page items in the list of items in a full-width column. (If a two column section is changed to a single, full-width column, page items previously in the secondary column still have their column set to "secondary.")
* Fix InteractivePage's `duplicate` method to properly set and reset section positions when a page is copied.

### DIFF
--- a/app/models/interactive_page.rb
+++ b/app/models/interactive_page.rb
@@ -203,15 +203,10 @@ class InteractivePage < ActiveRecord::Base
   def add_embeddable(embeddable, position = nil, section_identifier = nil, column = PageItem::COLUMN_PRIMARY)
 
     section_identifier ||= Section::DEFAULT_SECTION_TITLE
-    # Local function to test whether section_identifier is a numeric value
-    numeric = ->(x) { Float(x) != nil rescue false }
 
-    # look for a section specified by instance, title or ID
+    # look for a section specified by instance or title
     if section_identifier.is_a?(Section)
       section = section_identifier
-    elsif numeric.call(section_identifier)
-      section = sections.find { |s| s.id = section_identifier}  # <--- this seems like the = would blow things up
-      throw "Cant find section #{section_identifier}" unless section
     else
       section = sections.find { |s| s.title == section_identifier }
       unless section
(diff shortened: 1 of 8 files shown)


scytacki:
Responding to some slack comments from @emcelroy 
> I still haven’t figured out how the source section items get their positions updated after a page item is moved.
I'd guess that acts-as-list is being smart and it can tell the page item is being moved from one scope (section) to another. So then it updates the positions of all of the items in the old scope. I think this updating of positions was a reason for the slow responses before. In both cases the change was triggering a position update of all page items with a section of `nil` and I'd guess that might be a lot of page items.

> I’m also unsure of the error handling in moveItem in move-utils.tsx. What was there already seemed a little strange to me. I had to modify the function a little and now it seems a little less strange, but still a little iffy. I’m not sure if returning null when there’s an error is the best thing to do.
Yeah I don't know about this either. I'm not familiar with the rest of the code in LARA authoring so I don't know what the conventions are.

> And lastly, I wasn’t sure if it’d be safe to drop support for finding sections by title in the interactive page model’s add_embeddable method. So I left that in for now.
I think we have to leave the title lookup in there. I think it is needed to support importing  exports from LARA1.

app/models/interactive_page.rb
```diff
@@ -301,12 +296,28 @@ def page_number
   def duplicate(helper=nil)
     helper = LaraDuplicationHelper.new if helper.nil?
     new_page = InteractivePage.new(to_hash)
-    new_sections = sections.map { |s| s.duplicate(helper) }
 
     InteractivePage.transaction do
+      new_sections = sections.map { |s| s.duplicate(helper) }
```
scytacki, line 301: The fact that is first adds the copied sections to the original page will cause problems in the future. 
Can you pass the new page to the duplicate? So the new sections are created in that new page right from the beginning?
I wonder how this same type of problem is handled when duplicating the page items in a section.

scytacki (changes requested):
Having a second look.  I think the only thing that should be changed is the duplicating of sections. I haven't looked but maybe the duplicating of page items suffers the same problem so it also needs to be updated.

app/models/interactive_page.rb
emcelroy: @scytacki Do the latest changes look OK? They seem to work correctly in terms of adding the new sections directly to the new page and preventing the need to reset the original sections' positions.

As I understand it, the [`duplicate` method for page items](https://github.com/concord-consortium/lara/blob/new-sections/app/models/page_item.rb#L22-L47) already adds new page items directly to the appropriate section, instead of copying them in place and then moving them if needed.

scytacki (approved):
Looks good to me.

PR #1029, opened 2022-08-23
Fix after publishing, editing activity doesn't display content in report

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/182930937

[#182930937]

Call `update_activity_changed_by` from all API InteractivePages controller methods that modify activity content. The `update_activity_changed_by` will save the activity which will in turn trigger a re-publish as needed.

### DIFF
--- a/app/controllers/api/v1/interactive_pages_controller.rb
+++ b/app/controllers/api/v1/interactive_pages_controller.rb
@@ -39,18 +39,16 @@ class Api::V1::InteractivePagesController < API::APIController
   ## Mutations
   def copy_page
     activity = @interactive_page.lightweight_activity
-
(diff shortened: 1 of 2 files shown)


scytacki (changes requested):
These changes look good. Does it also work when an author makes a change to an interactive authored state. For example the prompt of a question?

emcelroy:
@scytacki Yeah, I tried changing question prompts as part of my testing. After these changes were in place, edits to prompts would show up in report-service almost immediately. 

Is there anything else like that I should check before merging?

scytacki (approved):
Sounds good to me. Go ahead and merge.

PR #1031, opened 2022-08-23
Fix reloading page with TE window shade causes blank page.

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/182918167

[#182918167]

These changes make it so Sharing Plugin items aren't rendered as separate page items in authoring. When they were being rendered as page items, it would cause a conflict with Teacher Edition Plugin page items that led to the page crashing.

### DIFF
--- a/lara-typescript/src/section-authoring/components/section-column.tsx
+++ b/lara-typescript/src/section-authoring/components/section-column.tsx
@@ -132,14 +132,10 @@ export const SectionColumn: React.FC<ISectionColumnProps> = ({
 
   const showItemPicker = () => setShowAddItem(true);
 
+  // TODO: Find a way to more generically identify wrapping items so this function isn't needed.
   const isWrappingItem = (itemData: any) => {
-    if (
-      itemData.componentLabel === "questionWrapper" ||
-      itemData.componentLabel === "interactives" // sharing plugin's component label
-      ) {
-      return true;
-    }
-    return false;
+    // "interactives" is the Sharing Plugin's label
+    return ["questionWrapper", "interactives"].indexOf(itemData.componentLabel) !== -1;
   };
 
   return (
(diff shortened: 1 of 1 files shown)


lara-typescript/src/section-authoring/components/section-column.tsx
```diff
@@ -151,14 +161,14 @@ export const SectionColumn: React.FC<ISectionColumnProps> = ({
                     // TODO: Figure out why item.data property names are sometimes not converted
                     // to camel case, and fix it instead of using this itemData variable.
                     const itemData = snakeToCamelCaseKeys(item.data);
+                    if (isWrappingItem(itemData)) {
```
dougmartin, line 164: Maybe add a comment here as to why we are exiting out early is it is a wrapping item?

lara-typescript/src/section-authoring/components/section-column.tsx
```diff
@@ -132,6 +132,16 @@ export const SectionColumn: React.FC<ISectionColumnProps> = ({
 
   const showItemPicker = () => setShowAddItem(true);
 
+  const isWrappingItem = (itemData: any) => {
+    if (
```
dougmartin, line 136: You could also do this:  `return ["questionWrapper", "interactives"].indexOf(itemData.componentLabel) !== -1`

However, this does give a nice breakpoint location to debug in.

dougmartin (approved):
Looks good :+1: 

I just had a couple of optional comments.

lara-typescript/src/section-authoring/components/section-column.tsx
scytacki: I agree a comment here is important. We are adding code that is depending on the componentLabels of specific plugins. So this means plugins are not generically handled. I think it is fine for now, so we can get this fixed, but you should comment about it.

Or if you could find out how to identify these wrapping items more generically that would be even better.

scytacki (commented):
It seems fine to me to merge as is. But if you wanted to spend more time on it, I provided a comment.

emcelroy:
@dougmartin @scytacki Thanks for the suggestions. I've made some changes, but stopped short of finding a way to generically identify wrapping plugins. I added a TODO comment about that.

dougmartin (approved):
Looks good :+1:

PR #1033, opened 2022-10-24
Fix list of available interactives in the interactive authoring

pjanik:
https://www.pivotaltracker.com/story/show/182437851

@emcelroy, you were looking at it before, so I'm marking you as a reviewer as you know the issue well.
Apparently, it was all caused by the `authoringApiUrls` vs `authoring_api_urls`, but debugging took me forever.

Two other commits are not related to this issue:
- updated some build JS files
- I saw a random warning in the same TS file and fixed it


emcelroy (approved):
Code changes look good to me, and it works as expected when tested in a browser.

Thanks for tracking that down!

PR #1036, opened 2022-11-09
Fix state update in the interactive authoring dialog

pjanik:
This PR fixes the root cause of the whole issue described here:
https://www.pivotaltracker.com/story/show/182710781

It also reverts the workaround from https://github.com/concord-consortium/lara/pull/1035.

The fix is a one-line change in ItemEditDialog... 🤦‍♂️ The previous version was mutating `pageItem` which should not happen. It essentially React state used to render components + this is query result returned from `react-query` `useQuery` hook: 
> Query results by default are structurally shared to detect if data has actually changed and if not, the data reference remains unchanged to better help with value stabilization with regards to useMemo and useCallback. 
(source: https://tanstack.com/query/v4/docs/guides/important-defaults)

Why was it causing issues related to the `linked_interactives` format? 
Update action / POST request expects `linked_interactibes: { linked_interactives: [], linked_state?: "int_123" }`, while GET result is `{ linked_interactives: [] }`.
Initially, I thought the problem was related to how the update is constructed (via forms). Then, I thought that it might be related to `react-query`. I thought it might automatically update the cached query result based on the update request, and since we have this mismatch, it might be breaking things. But... after reading all the react-query docs (it's actually a very cool library), I've realized that mutation and query results are fully independent. The only thing that mutation does is to invalidate query results, so they're re-fetched. So, we should be never seeing the UPDATE format of `linked_interactives` in the GET query results. So, I started looking for mutations done in place and found this one.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -452,9 +452,9 @@ ActiveRecord::Schema.define(:version => 20220808120311) do
     t.string   "runtime",                                :default => "LARA"
     t.string   "background_image"
     t.string   "fixed_width_layout",                     :default => "1100px"
+    t.integer  "glossary_id"
     t.boolean  "defunct",                                :default => false
     t.string   "migration_status",                       :default => "not_migrated"
-    t.integer  "glossary_id"
   end
 
   add_index "lightweight_activities", ["changed_by_id"], :name => "index_lightweight_activities_on_changed_by_id"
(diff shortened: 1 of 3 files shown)


scytacki (approved):
Looks good to me.

scytacki:
Yes react query is a cool library. If you've got your own state library like MST or Redux I think it gets in the way, but without a state library it is really helpful.

PR #1039, opened 2022-11-10
Make `npm run build:dev:lara-typescript` faster, restructure LARA typescript webpack config

pjanik:
There might be better solutions, but this will let me not having to constantly comment and uncomment parts of the webpack config. It makes build time faster when you work only on authoring part.


scytacki (approved):
Looks good. 👍

PR #1040, opened 2022-11-10
Fix state linking in Managed Interactives, let authors unlink interac…

pjanik:
[[#183759609]](https://www.pivotaltracker.com/story/show/183759609)

This PR fixes two issues:
1. Linked Interactive ID wasn't being displayed in the authoring UI
2. It was impossible for authors to unlink interactive. Removing the interactive ID or setting it to some random value was resulting in LARA API returning 404 error and the linked_interactive property wasn't updated.


scytacki (approved):
Looks good 👍

PR #1041, opened 2022-11-10
Fix interactive authoring preview to show updated linked interactives array

pjanik:
[[#183641151]](https://www.pivotaltracker.com/story/show/183641151)

This PR fixes interactive authoring preview. It wasn't handling updates of the linked interactives, so in some cases, authors could get confused. PT story describes that nicely using the image question example.

I've extracted the preview update functionality to `useAuthoringPreview` hook and used that in both MwInteractive and MangedInteractive classes to avoid code duplication.

### DIFF
--- a/lara-typescript/src/page-item-authoring/common/hooks/use-authoring-preview.ts
+++ b/lara-typescript/src/page-item-authoring/common/hooks/use-authoring-preview.ts
@@ -1,5 +1,5 @@
 import { ILinkedInteractive, ISetLinkedInteractives } from "@concord-consortium/interactive-api-host";
-import { RefObject, useCallback, useEffect, useState } from "react";
+import { RefObject, useCallback, useEffect, useMemo, useState } from "react";
 import { debounce } from "ts-debounce";
 
 interface IInteractiveBase {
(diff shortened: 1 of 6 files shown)


lara-typescript/src/page-item-authoring/common/hooks/use-authoring-preview.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  const _updatePreview = ({ newAuthoredState, newLinkedInteractives, newUrl }: IPreviewUpdate) => {
+    handleUpdateItemPreview?.({
+      authoredState: typeof newAuthoredState === "string" ? newAuthoredState : JSON.stringify(newAuthoredState),
+      linkedInteractives: newLinkedInteractives,
+      url: newUrl
+    });
+  };
+
+  const updatePreview = useCallback(debounce(_updatePreview, 500), []);
```
scytacki, line 40: This approach of `useCallback(debounce(...), [])` is confusing to me. If it makes sense to you, can you add a comment about how it works?

As far as I understand `debounce` will be called each time this hook is called. However, only the function returned by debounce the first time will be used used due to the dependencies of `[]`.  So updatePreview will always be a debounced version of the first `_updatePreview` function.  If the caller of `userAuthoringPreview` passes a different `handleUpdateItemPreview` function this will never be called.  I might have that all wrong though.

scytacki (approved):
This looks fine to me.

It seems like the useCallback, debounce, and useEffect could be simplified but I don't totally get what is going on.

lara-typescript/src/page-item-authoring/common/hooks/use-authoring-preview.ts
kswenson: Some suggest using `useMemo` for this purpose, e.g. https://dmitripavlutin.com/react-throttle-debounce/.

pjanik: > As far as I understand debounce will be called each time this hook is called. However, only the function returned by debounce the first time will be used used due to the dependencies of []. So updatePreview will always be a debounced version of the first _updatePreview function. If the caller of userAuthoringPreview passes a different handleUpdateItemPreview function this will never be called. 

@scytacki, I think you're right. This line was copied from the old code found in MwInteractive and ManagedInetractive React components. I saw it was working fine (the delay was noticeable and updates were merged), so frankly speaking, I didn't think about it too much while moving it to this hook.

I've just pushed one more commit that cleans it up following some advice from @kswenson's article.

On a related note - it seems that our linter + rules of hooks don't work too well. I believe it should have complained about the original code (empty dependency array) and it didn't. I've noticed that in a few other places too. I bet it's caused by the fact that we use deprecated TSLint + its rules of hooks plugin. I guess it'd be great to migrate to ESLint when we have another time for maintenance.

scytacki (approved):
Looks good to me.

PR #1050, opened 2022-12-23
Interactive API: ensure that setInteractiveState, setAuthoredState and setGlobalInteractiveState are stable

pjanik:
[#184099334]

When a developer is using React `useState` hook, the set state function is expected to be "stable". This is not the case for our setInteractiveState, setAuthoredState and setGlobalInteractiveState functions provided by custom hooks in LARA Interactive API. It might cause infinite loops of re-renders in some cases. This PR fixes this problem by adding `useCallback`.

### DIFF
--- a/docs/interactive-api-client/globals.md
+++ b/docs/interactive-api-client/globals.md
@@ -1183,7 +1183,7 @@ ___
 
 * **authoredState**: *null | AuthoredState*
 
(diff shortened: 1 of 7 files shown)


scytacki (approved):
looks good to me

kswenson (approved):
👍 Looks good

PR #1057, opened 2023-01-28
Update README.md

pjanik:
Small README update. I often create LARA Codespaces without Portal and follow this approach.


scytacki (approved):
Looks good.

PR #1058, opened 2023-01-30
Fix page settings dialog

pjanik:
[#184002921]

Goal of this PR was to ensure that author cannot add sidebar to the summary/completion page.

So, each time author selects `Page is a completion/summary page` checkbox, it'll unselect sidebar checkbox.

Also, it redesign the whole dialog handling code, so it uses controlled inputs. There were other bugs in the previous implementation. Eg. when author clicked twice on any checkbox (checked and unchecked, or unchecked and checked), it'd still toggle the setting. :wink:


emcelroy (approved):
Looks good 👍 Thanks!

PR #1059, opened 2023-01-30
Disable completion page checkbox when page is not empty

pjanik:
[[#184268545]](https://www.pivotaltracker.com/story/show/184268545)

The story description suggests not rendering the questions, but I asked on Slack and Scott confirmed that disabling the checkbox might be an easier and better solution.


emcelroy (approved):
Looks good!

PR #1062, opened 2023-02-24
feat: Added bi-directional custom messages (PT-184475161)

dougmartin:
Before custom messages were only host -> client and now they are bi-directional.  This change includes a sendCustomMessage api helper that the client can use to send a message to the host.

The same message name was used for both host and client messages to denote this change isn't the normal request/response api message set.


dougmartin:
@scytacki This PR is WIP/draft as it needs the final package.json version but it is ready for review.

scytacki (approved):
looks good

PR #1066, opened 2023-03-09
Update GH Codespaces URL host in sample env file

pjanik:
[#184536435]

This PR updates the host to a new version used by GH Codespaces. 
There are still some problems with oauth, but this is necessary update anyway.


scytacki (approved):
Looks good.

PR #1067, opened 2023-03-15
feat: Added hide read aloud setting to activities and sequences [PT-184688393]

dougmartin:
This boolean setting defaults to false and is exported, imported and duplicated.


pjanik (approved):
Looks good 👍

PR #1068, opened 2023-03-15
fix: Update CF stack template for new staging stack

433eros:
The LARA CloudFormation stack template needed to be tweaked quite a bit to add new stuff for the new staging environment stack


scytacki (changes requested):
Can you give me a link so I can see what has been changed?

There might be a better way, but here is one:
- add the original LARA cf stack template in a first commit
- then add a second commit with this file

Then I can look at the second commit to see the diff.

PR #1069, opened 2023-03-16
fix: Updated hide read aloud checkbox hints [PT-184688393]

dougmartin:
- Changed sequence hint to mention sequences
- Added a period at end of activity hint
- Removed (unrelated) double period in runtime hint


lublagg (approved):
Looks good! 👍

PR #1070, opened 2023-03-21
fix: Add auto scroll to edit preview panel [PT-182710801]

dougmartin:
This adds an overflow: auto directive to the item edit preview panel to allow scrolling to see the bottom of the labbook and image interactives.


pjanik (approved):
Looks good 👍

PR #1073, opened 2023-04-18
feat: Added font size setting to activities and sequences [PT-184910002]

dougmartin:
This adds a new module named Accessible that holds the font size settings, and possible future accessibility settings.

Note: the final placement of this setting in the edit forms is yet to be determined.


scytacki (approved):
Looks good to me. Amazing how many places have to be changed to add one field.

PR #1075, opened 2023-04-19
feat: Updated layout of activity and sequence new/edit forms [PT-184909981]

dougmartin:
This also updates the activity new/edit to use a shared partial as the new form seemed to have been neglected over the years and didn't get any new fields as they were added.


scytacki (approved):
Looks good to me.

PR #1076, opened 2023-04-21
feat: Add accessibility object to initInterative message [PT-184987026]

dougmartin:
This adds an accessibility object to the initInteractive message that currently contains both the name of the required font size (normal or large) and the font size in pixels.

The two font settings allow interactives to either directly set the font size by value or use the font name in their css files.

### DIFF
--- a/app/helpers/interactive_run_helper.rb
+++ b/app/helpers/interactive_run_helper.rb
@@ -51,6 +51,12 @@ module InteractiveRunHelper
     data['interactive-name'] = interactive.name
     data['linked-interactives'] = interactive.linked_interactives_list.to_json
 
+    # default to normal size unless set in activity
+    data['font-size'] = 'normal'
+    if run && run.activity && run.activity.font_size
+      data['font-size'] = run.activity.font_size
+    end
+
     opts = {
       :src => interactive.url,
       :data => data,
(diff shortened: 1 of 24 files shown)


app/helpers/interactive_run_helper.rb
```diff
@@ -51,6 +51,12 @@ def interactive_data_div(interactive,run)
     data['interactive-name'] = interactive.name
     data['linked-interactives'] = interactive.linked_interactives_list.to_json
 
+    # default to normal size unless set in activity
+    data['font-size'] = 'normal'
+    if run && run.activity && run.activity.font_size
+      data['font-size'] = run.activity.font_size
+    end
+
```
scytacki, line 59: Is this used anymore, without the lara runtime?

lara-typescript/src/interactive-api-client/hooks.ts
```diff
@@ -176,3 +176,39 @@ export const useReportItem = <InteractiveState, AuthoredState>({ metadata, handl
     return () => client.removeGetReportItemAnswerListener();
   }, []);
 };
+
+export const DefaultAccessibilitySettings: IAccessibilitySettings = {
+  fontSize: "normal",
+  fontSizeInPx: 16
+};
+
+export const useAccessibility = (props?: {updateHtmlFontSize?: boolean, addBodyClass?: boolean}) => {
```
scytacki, line 185: This `useAccessibility` hook is nice. As long as clients use this, it will make it possible for us to introduce the idea of dynamic accessibility settings. Where students can change the settings and have the components update. I realize that isn't supported by the current initMessage approach of passing them around, but the `useAccessibility` abstracts that from the client.

scytacki (approved):
This looks good to me.

app/helpers/interactive_run_helper.rb
dougmartin: It isn't, but I didn't want to introduce a bunch of noise into the PR with deleting old LARA runtime files - that should be another (large) PR in the future.

PR #1078, opened 2023-04-26
fix: Added default font size for authoring preview [PT-185043384]

dougmartin:
Hack for now.  The interactive.font_size is always undefined for now as the embeddable to_hash Ruby method does not include it as it is a property of the top level activity or sequence.

This may be the way we wish to keep it (defaulting to normal) so that the editing experience does not need to be tested for the large font setting.

The user can always use the AP preview to view what it looks like in the large font.


scytacki (approved):
looks good to me

PR #1079, opened 2023-05-02
feat: Updated sequence and activity form labels and hints [PT-184909981]

dougmartin:
- In Sequence Settings, Preview changed to Thumbnail
- In Sequence Settings, added "This will override the "Activity Page Fixed Width" setting for any activities in this sequence." text to Activity Page Fixed Width Layout
- In Sequence Settings, added "This will override the "Activity Page Background Image URL" setting for any activities in this sequence." text to Activity Page Background Image URL fields
- For Sequence and Activity Font Size setting, added "Note some interactives may not respond to font size settings." text


scytacki (approved):
Looks good to me.

PR #1082, opened 2023-06-01
feat: Added section name editing [PT-184813606]

dougmartin:
This adds an inline editing feature in the section authoring component to set the section name.


lublagg (approved):
Looks good 👍

PR #1083, opened 2023-06-02
fix(ProjectSettings): rm invalid references to theme model, update spec, node-sass -> sass 

pjanik:
[[PT-185317933]](https://www.pivotaltracker.com/n/projects/2441249/stories/185317933)

Long story short, the themes model and controller were removed, but the theme was still referenced by the project settings form. I guess this issue was introduced during one of [the recent LARA cleanups](https://www.pivotaltracker.com/n/projects/2441249/stories/184917932), PR: https://github.com/concord-consortium/lara/pull/1074.

Note that I also replaced `node-sass` with `sass` to fix issues with `npm install`. This never caused me any issues, so I hope it's also the case here. But I can see that the rebuilt CSS file is slightly different (mostly whitespace as far as I can tell).


scytacki (approved):
Looks good to me.

PR #1084, opened 2023-06-02
fix(ProjectSettings): rm invalid references to theme model, update spec [PT-185317933]

pjanik:
This PR is essentially the same as https://github.com/concord-consortium/lara/pull/1083, but without `node-sass` -> `sass` update.


scytacki:
We ended up cherry picking the commit from the other PR and using that in a patch build. The other PR has been merged into master so I'm going to close this one.

PR #1085, opened 2023-06-08
fix: Added max section name display length [PT-185314409]

dougmartin:
This fix changes the section name to have a constant width with hidden overflow with ellipsis for the name when it gets too long.


lublagg (approved):
Looks good 👍

PR #1087, opened 2023-06-13
fix: Added modal title truncation [PT-185369384]

dougmartin:
This truncates long modal titles so they don't overflow to the next line.


lublagg (approved):
Looks good.

PR #1088, opened 2023-06-13
feat: Added mediaLibrary to initInteractiveMessage [PT-185394064]

dougmartin:
This is not used by the old Lara iframe-saver code but is used by AP to gather up media library items and to pass them to interactives like the lab book for use as an image source.

NOTE: there is a chore in the notebook epic to do the final release, this commit uses a pre-release version.


dougmartin:
@scytacki I assigned this to you along with @lublagg so you can see the new API for the media library.

lublagg (approved):
This looks good to me. 👍

scytacki (approved):
Looks good to me. 👍

PR #1089, opened 2023-06-14
feat: Add setting for override activity layout in sequences [PT-185316649]

dougmartin:
This adds a layout override option to sequences.  The default 0 value means no override and the existing layouts option values are shifted up by 1.


lublagg (approved):
Looks good 👍

PR #1090, opened 2023-06-15
feat: Added font type to accessibility [PT-185336244]

dougmartin:
This adds the font type to the accessibility object in the initInteractive message.

This adds a new option to the useAccessibility hook options: a css selector to use to create a font-family rule.  This allows question-interactives to target the inner runtime div.

As with the original accessibility work the LARA scripts that are no longer used have also been updated.

### DIFF
--- a/app/helpers/interactive_run_helper.rb
+++ b/app/helpers/interactive_run_helper.rb
@@ -57,6 +57,12 @@ module InteractiveRunHelper
       data['font-size'] = run.activity.font_size
     end
 
+    # (for now) the font type is implied by the layout and not set directly on the interactive
+    data['font-type'] = 'normal'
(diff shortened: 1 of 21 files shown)


dougmartin:
@scytacki I'm leaving this as a WIP until I get back on Tuesday.  Take a look at the approach and see what you think.  I ended up going with a new style rule as question-interactives has a local font-family rule at the inner runtime div so I couldn't target the body tag for the dynamic font-family update.

I left this WIP as I need to figure out how to set the injected style tag.  It isn't showing up in Jest but does work for real.

scytacki (approved):
I skimmed this. It looks OK to me when looking at it quickly. Probably should wait for a more complete review with Lucy before merging.

lublagg (approved):
Looks good!

PR #1091, opened 2023-07-06
feat: Check if multiple copies of the lib are imported and throw an error when it happens [PT-185546697]

pjanik:
[#185546697](https://www.pivotaltracker.com/story/show/185546697)

This PR is related to debugging and issues we had with question-interactives that for some time were importing two copies of lara-interactive-lib. A detailed description of this problem: https://concord-consortium.slack.com/archives/C0M5CM1RA/p1688581569887089

Generally, lara-interactives-api cannot be imported twice, as it uses some module-level variables as a cache/copy of the interactive state. This copy is used to automatically respond to the parent window when it requests an interactive state (so developers don't need to handle that manually) or to debounce state updates.  When two copies are defined, these state copies might be out of sync and lots of unexpected things might happen.

The code I added is very brutal - it'll display an alert and throw an error to fully break the execution. I did that, as it's very likely that a console warning would be overlooked or ignored. And leaving interactive in this state might lead to painful and time consuming debugging.

I've tested this code in the problematic version of question-interactives and it works as expected:
Screenshot 2023-07-06 at 13 51 04

### DIFF
--- a/lara-typescript/src/interactive-api-client/index.ts
+++ b/lara-typescript/src/interactive-api-client/index.ts
@@ -4,3 +4,25 @@ export * from "./in-frame";
 export * from "./api";
 export * from "./hooks";
 export * from "./client";
+
+import * as packageJson from "./package.json";
+
+const version = packageJson.version;
+
+const warningMsg = (loadedVersion: string, newVersion: string) => `
+LARA Interactive API is loaded multiple times. This will lead to unexpected behavior and might break multiple features
+of the API (especially interactive state saving). Please ensure that the library is loaded only once by the main app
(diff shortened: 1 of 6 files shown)


dougmartin (approved):
Looks good :+1: 

I think @scytacki should also review this too though.

pjanik:
I released it as a minor version bump to make it easier to update in other projects / hope they'll pick it up automatically. But we could also release it as `v1.10.0`, so it's easier to remember when this feature was added and that it might have these implications (btw, probably media library features should also be released as `v1.10.0`). But not sure if it makes a significant difference in practice.

scytacki (approved):
This looks good to me. A minor version bump seems good to me.

PR #1092, opened 2023-10-27
feat: Added starting z-index to plugin api [PT-185637498]

dougmartin:
**WIP**: need to update version and publish final version after PR approved.


dougmartin:
This is no longer needed.  The issue with the AP z-index (https://www.pivotaltracker.com/story/show/185637498) is due to the grid layout and can't be fixed with z-index tweaks.

PR #1094, opened 2023-11-13
feat: Change project list ordering to order by title [PT-186428022]

dougmartin:
This changes the project list api endpoint to return the projects in order by title.  This is only used by the ProjectList component shown on the /projects page.


lublagg (approved):
Looks good 👍

PR #1101, opened 2023-11-28
feat: Refactored ability.rb and spec test [PT-186125994] [PT-186126000]

dougmartin:
- Removed unneeded extra admin "cans" (`can :manage, :all` is sufficient)
- Reorganized author "cans" by model and alphabetized them
- Added missing can update own sequence
- Added more tests to user_spec to handle all existing and the one new sequence can


lublagg (approved):
Looks good 👍

PR #1107, opened 2023-12-05
fix: Missing edit link on glossaries for project admins [PT-186530973]

dougmartin:
The glossary used a method outside of cancan to set the editing ability.


lublagg (approved):
👍

PR #1108, opened 2023-12-06
fix: user.project_admin_of? with nil project

dougmartin:
This was returning nil instead of false which then caused nothing to be output in the glossary edit html javascript section breaking the glossary view link when the user was not a an admin or project admin.


pjanik (approved):
Looks good 👍

PR #1109, opened 2023-12-31
ci: Replace Travis with GitHub Actions (PT-186702504)

433eros:
See PT-186702504 for more information. Currently `npm ci` doesn't work for the `lara-typescript` folder due to a dependency conflict: https://github.com/concord-consortium/lara/actions/runs/7373046594/job/20061898636#step:10:15

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,13 +38,13 @@ jobs:
       - run: nix develop -c bundle exec rake
       - name: Test LARA plugin API v2
         run: |-
-          npm ci
-          npm run test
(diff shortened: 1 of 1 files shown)


dougmartin (commented):
Looks good :+1: 

I'm just commenting instead of approving so that @scytacki has a chance to also review it.  

I see we are adding `nix` to the mix - is that going to cause other devs issues (such as needing to install anything) if they need to update this in the future?

dougmartin (approved):
Looks good after review during meeting :+1:

PR #1120, opened 2024-01-26
ci(deploy): Create composite Action for deployment

433eros:
Since rigse has an identical deployment process I converted the deployment workflow to a 'composite Action' so its reusable across repos.


scytacki (approved):
looks good to me

PR #1122, opened 2024-02-05
ci: Make deployment workflow fully parameterized

433eros:
There's some changes that need to be made so this can be used in Portal:
- Add the `workflow_call` block to declare the workflow as "reusable" in other repositories
- Parameterize the value for the `ParameterKey` itself since the CloudFormation templates for LARA & Portal use different parameter names (`LaraDockerImage` vs. `PortalDockerImage`)


scytacki (approved):
looks good to me

PR #1123, opened 2024-02-07
fix:  Allow non-admins to edit existing activities with non-official library interactives [PT-186993789]

dougmartin:
With this change the server no longer filters the non-official library interactives and instead of the React UI does the filtering.

This fix was made so that project admins could edit activities with non-official library interactives included in them.

There were also some misc lint issues fixed from existing code that were highlighted when the TypeScript was rebuilt.


lublagg (approved):
LGTM 👍

PR #1124, opened 2024-02-15
fix: Plugins not showing in add item UI for non-admins [PT-187056636]

dougmartin:
This fix adds the official flag to all embeddable plugins so they shown to non-admins in the add item UI.  There is no official flag in the plugin server side model so it needs to be set on the client side.


scytacki (approved):
seems fine. 👍

PR #1126, opened 2024-03-18
Update readme to include information about GH personal access tokens

pjanik:
It adds this information to both the initial section and the Codespaces section (I only read this one while setting up a new Portal/LARA environment).

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,11 +7,19 @@ This is a Rails application intended to provide a platform for authoring and usi
 
 ## Getting started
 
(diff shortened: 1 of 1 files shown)


scytacki:
Oh I do see that "CI / Build Docker image (push)" failed. I don't know if that is important, but since this is a just a readme change it can't be caused by this PR.

pjanik:
> Oh I do see that "CI / Build Docker image (push)" failed. I don't know if that is important, but since this is a just a readme change it can't be caused by this PR.

I restarted it and it worked now. Hopefully some isolated glitch, and yes, it can't be related to readme updates.

PR #1129, opened 2024-04-08
Update GH Codespaces sample .env file

pjanik:
Recently, my GH Codespaces instances started to use `.app.github.dev` domain instead of `preview.app.github.dev`. This obviously breaks quite a few things in the Portal and LARA connection. It's the second time this has happened to me (although the previous update was a long time ago). It may be some global update in Codespaces or specific to my instances.

I was looking if there's some way to improve that and found an env variable that automatically returns the port forwarding domain: `GITHUB_CODESPACES_PORT_FORWARDING_DOMAIN`. It's documented here:

https://docs.github.com/en/codespaces/developing-in-a-codespace/default-environment-variables-for-your-codespace


dougmartin (approved):
Looks good :+1:

PR #1130, opened 2024-05-06
Cleanup potentially misleading references to Shutterbug server

pjanik:
https://www.pivotaltracker.com/story/show/187531185

SHUTTERBUG_URI was used by the older versions of Question Interactive, Labbook, etc., which were implemented within the LARA JS code. I've removed all that since it could potentially confuse someone trying to debug Shutterbug issues.

Currently, the Shutterbug Server URL is embedded within Shutterbug.js and it can be overridden using a URL parameter. However, this approach will be tricky for iframe questions within the Activity Player. But that, in fact, is a topic for another discussion.

I've also removed some JS code, as I assume we no longer support these old questions in any way.


dougmartin (approved):
Looks good :+1: Too bad Noah isn't still around - he would have loved to see all this deleted code. :)

PR #1131, opened 2024-07-18
feat: Add rubric model [PT-187946683]

dougmartin:
This is a mirror of the glossary model with the glossary specific attributes (like legacy glossary id, removed).


bacalj (approved):
New to rails, but looks like a solid setup for the new rubric setup

pjanik (approved):
Looks good 👍

PR #1132, opened 2024-07-19
feat: Add create rubric [PT-187944087]

dougmartin:
This also adds the foundation for future stories by copying the functionality of the glossary views and controllers.


pjanik (approved):
Looks good 👍 
I have a question about HAML: Is that all we'll add? 
I'm seeing comments suggesting that the rest of the authoring will be rendered in JS, right? 
So, the current HAML was added because it was significantly easier than creating some React components for that? In general, my preference would be not to add any classic forms and HAML anymore, but if it fits better here, I understand.

dougmartin:
@pjanik This works like the glossary authoring where there is a haml page that just defines a div with an id and then there will be a React component (to be done in a later story) that renders at that id.

PR #1133, opened 2024-07-19
feat: List rubrics on homepage [PT-187944098]

dougmartin:
This follows the other resource patterns and lists the 10 most recently modified rubrics on the homepage.

This also adds a missing translation for "GLOSSARIES" which wasn't a problem but was showing a missing translation error in the html.


pjanik (approved):
Looks good 👍

PR #1134, opened 2024-07-19
feat: Add rubric import [PT-187982777]

dougmartin:


bacalj (approved):
Thanks again, approving as it seems all straightforward adding according to new rubric resource

pjanik (approved):
Looks good 👍

PR #1135, opened 2024-07-19
fix: Missing GON window variable [PT-187984369]

dougmartin:
This fixes a bug introduced when the shutterbug uri variable GON variable was removed.  This caused no GON variables to exist which caused the script tag that initialized the window.gon variable to not be emitted, which other scripts rely on to exist.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -41,6 +41,7 @@ class ApplicationController < ActionController::Base
   before_filter :reject_old_browsers, :except => [:bad_browser]
   before_filter :set_locale
   before_filter :store_auto_publish_url # to enable auto publishing to build an url from the request object
+  before_filter :intialize_gon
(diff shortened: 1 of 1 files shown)


pjanik (approved):
Looks good 👍 
Just one quesiton - if there are no gon variables, can't we get rid of it completely? 
I wonder what other scripts are looking for if it's empty now.

app/controllers/application_controller.rb
```diff
@@ -41,6 +41,7 @@ class ApplicationController < ActionController::Base
   before_filter :reject_old_browsers, :except => [:bad_browser]
   before_filter :set_locale
   before_filter :store_auto_publish_url # to enable auto publishing to build an url from the request object
+  before_filter :intialize_gon
```
kswenson, line 44: `intialize_gon` => `initialize_gon`

PR #1138, opened 2024-07-24
feat: Add rubric edit settings form [PT-187984395]

dougmartin:


lara-typescript/src/rubric-authoring/rubric-settings.tsx
```diff
@@ -0,0 +1,70 @@
...
+            {projects.map(({id, title}) => <option key={id} value={id}>{title}</option>)}
+          </select>
+        </div>
+        <div className="rubric-settings-buttons">
+          <input type="submit" value="Save Changes" disabled={formDisabled} />
+          <button onClick={cancel}>Cancel</button>
+        </div>
+      </form>
+    </div>
+  );
```
pjanik, line 69: You have a mix of controlled and uncontrolled inputs in this component. It doesn't seem like it's for a reason.

Wouldn't it be better to use just one approach here, and probably in the subsequent code as well?

pjanik (approved):

PR #1139, opened 2024-07-25
feat: Added rubric general options and TBD placeholders for other sections [PT-187942504]

dougmartin:
NOTE: value gathering and saving will be implemented in a future story.


pjanik (approved):
Looks good 👍

PR #1140, opened 2024-07-25
feat: Change edit settings toggle to a button [PT-187984395]

dougmartin:
This updates the UI to match the spec that was updated since the initial edit settings PR was created.


pjanik (approved):
Looks good 👍

PR #1141, opened 2024-07-25
feat: Add rubric ratings [PT-187942540]

dougmartin:
NOTE: the svg icons are placeholders and will be updated when the icons are created and exported in the spec.


pjanik (approved):
Looks good 👍

PR #1142, opened 2024-07-26
feat: Added rubric criteria [PT-187942542]

dougmartin:
Also added a useRubric hook and context to track the rubric values.


pjanik (approved):
Looks good 👍

PR #1147, opened 2024-07-29
feat: Add criteria groups [PT-187942542]

dougmartin:
This adds a group element above the criteria with optional labels that are displayed in the table.


pjanik (approved):
Looks good 👍

PR #1148, opened 2024-07-29
feat: Added optional criteria icon url [PT-188024114]

dougmartin:
This url, if provided, is displayed next to the criteria description in the rubric table.


pjanik (approved):
Look good 👍

PR #1150, opened 2024-07-31
fix: Various rubric QA issues [PT-187942542]

dougmartin:
- Added disable of clicks on scoring guide when the url is invalid
- Added back monospace font for code tag in rubric table markdown
- Fixed broken table layout when description text length is long


pjanik (approved):
Looks good 👍

PR #1151, opened 2024-08-01
feat: Added authored contents api [PT-187967104]

dougmartin:


pjanik (approved):
Looks good 👍

app/controllers/api/v1/authored_contents_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+      # not an environment variable - we use the source in the key to split production, staging, and development content
+      bucket_name: "cc-project-resources",
+      # production and staging can set the source, in development we don't set it and fall back to the (probably Docker) hostname
+      source: ENV["AUTHORED_CONTENT_SOURCE"] || "hostname_#{Socket.gethostname()}"
+    }
+  end
+
+  def s3_put(s3_config, key, contents)
+    s3 = AWS::S3.new(:access_key_id => s3_config[:access_key_id], :secret_access_key => s3_config[:secret_access_key])
+    s3.buckets[s3_config[:bucket_name]].objects.create(key, contents)
```
scytacki, line 42: I guess this `create` function will work even if the key already has a value in S3?

scytacki (approved):
Looks good to me.

app/controllers/api/v1/authored_contents_controller.rb
dougmartin: Yes, it will overwrite the contents.

PR #1152, opened 2024-08-01
feat: Added rubric saving using authored content api [PT-187967207]

dougmartin:


app/models/rubric.rb
```diff
@@ -1,12 +1,15 @@
 class Rubric < ActiveRecord::Base
-  attr_accessible :name, :user_id, :project_id, :project
+  attr_accessible :name, :user_id, :project_id, :project, :authored_content_id, :authored_content
   validates :name, presence: true
   validates :user_id, presence: true
 
   belongs_to :user
   belongs_to :project
+  belongs_to :authored_content
```
scytacki, line 8: Is it necessary to add the authored_content field to the rubric table? Since the authored_content points back to the rubric, can you just use a `has_one :authored_content` instead? Perhaps that'd need some polymorphic flag on it so Rails will handle it right.

scytacki (changes requested):
This looks good, my only concern is the two way reference of rubric pointing to authored_content and then authored_content pointing back to rubric.

app/models/rubric.rb
dougmartin: Fixed in https://github.com/concord-consortium/lara/pull/1152/commits/03271dde6aac7ba1d4001bcf955b1092a24d9375

scytacki (approved):
Looks good. 👍

PR #1156, opened 2024-08-05
fix: Remove scoreUsingPoints option and made "Show rating descriptions" conditional [PT-188034293]

dougmartin:
The scoreUsingPoints option was move to the portal dashboard.  The "Show rating descriptions" option is now disabled when the "Hide rubric from students in Student Report" option is checked.

### DIFF
--- a/lara-typescript/src/rubric-authoring/rubric-general-options.tsx
+++ b/lara-typescript/src/rubric-authoring/rubric-general-options.tsx
@@ -1,10 +1,11 @@
 import * as React from "react";
 import { useRubric } from "./use-rubric";
+import classNames from "classnames";
 
 import "./rubric-general-options.scss";
 
 type IRubricOptionKey = "referenceURL" | "criteriaLabel" | "criteriaLabelForStudent" | "feedbackLabelForStudent";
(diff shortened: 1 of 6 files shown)


lara-typescript/src/rubric-authoring/rubric-general-options.tsx
```diff
@@ -65,33 +66,29 @@ export const RubricGeneralOptions = () => {
...
-            <td className="rubric-checkbox">
-              <input
-                type="checkbox"
-                name="scoreUsingPoints"
-                checked={rubric.scoreUsingPoints}
-                onChange={handleUpdateCheckbox("scoreUsingPoints")}
-              /> Score using points
+                name="hideRubricFromStudentsInStudentReport"
+                checked={rubric.hideRubricFromStudentsInStudentReport}
+                onChange={handleUpdateCheckbox("hideRubricFromStudentsInStudentReport")}
```
pjanik, line 71: If you're always using the `name` prop, I sometimes have handlers where the key is obtained from `name`, like `const name = e.target.name`. But then you lose type checking, as you can't enforce it for `name`.

pjanik (approved):
Looks good 👍

PR #1159, opened 2024-08-06
feat: Add set rubric select to activity [PT-187942488]

dougmartin:
This adds a new select dropdown to the activity to select the rubric.  This is modeled after the activity glossary select.


pjanik (approved):
Looks good 👍

PR #1161, opened 2024-08-06
feat: Publish rubric document and json urls to portal [PT-187946772]

dougmartin:
This required the referenceUrl to be moved from the rubric JSON to the rubric model so that it was available to post to the portal.


pjanik (approved):
Looks good 👍

PR #1163, opened 2024-08-19
fix: Delete criterion index bug [PT-187942542]

dougmartin:
This fixes a bug where the index of the criterion was also being used as the index of the criteria group.


bacalj (approved):
lgtm

PR #1164, opened 2024-08-19
fix: Remove 24x24 criterion icon css constraint [PT-188024114]

dougmartin:
Per request from the PI it will be up to the author to resize the icon.


bacalj (approved):
lgtm

PR #1165, opened 2024-08-19
fix: Student rubric preview crash when no student descriptions [PT-188034293]

dougmartin:
This was caused by the use of .trim() on possibly undefined string values.


bacalj (approved):
lgtm

PR #1166, opened 2024-08-21
fix: Disable deletes of glossaries and rubrics when used in an activity [PT-188116484]

dougmartin:
This removes the delete link and adds text next to the activity usage count.


pjanik (commented):
Looks good 👍

pjanik (approved):

PR #1169, opened 2024-09-06
fix: Set the minimum rubric rating score to 1

dougmartin:
This avoids issues with the rubric not being considered scored when the score is checked.


pjanik (approved):
Looks good 👍

PR #1170, opened 2024-09-06
fix: Hide rubric export/import/copy links

dougmartin:
Neither export, import or copy currently contains a reference to the S3 authored content data so only the metadata is exported, imported or copied which is not really useful at all.

For now these options are hidden until we have time in the future to correctly implement the actions.


pjanik (approved):
Looks good 👍

PR #1177, opened 2024-10-11
feat: Add hide question number option to interactives [PT-187807322]

dougmartin:
This adds the hide_question_number option to interactives and library interactives.


bgoldowsky (approved):
I have not tested or checked if they are complete, but just reading through them these changes look very reasonable to add the configuration field.

PR #1183, opened 2024-10-18
feat: Add hide question number custom override [PT-187807322]

dougmartin:
This adds a custom override to hide question number authoring.


emcelroy (approved):
Looks good to me 👍

PR #1185, opened 2024-10-21
Address deprecation warning issues.

lublagg:


app/views/shared/_loading.html.haml
```diff
@@ -1,14 +1 @@
...
-  #loading-container {
-    display: none;
-    left:40%;
-    position: absolute;
-    text-align: center;
-    background: white;
-    padding: 10px;
-    font-size: large;
-    z-index: 100;
-  }
```
emcelroy, line 14: After the upgrade, this :scss instance was causing the error:

```
Failure/Error: <%= render :partial => 'shared/loading' %>
ActionView::Template::Error: wrong number of arguments (given 3, expected 1)
```

Removing it from here and adding the new _loading.scss file fixed that.

app/assets/stylesheets/application.css
```diff
@@ -33,4 +33,5 @@
  *= require lara-typescript
  *= require styled-admin-forms
  *= require project-editing
+ *= require shared/_loading
```
emcelroy, line 36: For this and the related addition of app/assets/stylesheets/shared/_loading.scss, see comment below on app/views/shared/_loading.html.haml.

config/application.rb
```diff
@@ -58,7 +58,7 @@ class Application < Rails::Application
     # This will create an empty whitelist of attributes available for mass-assignment for all models
     # in your app. As such, your models will need to explicitly whitelist or blacklist accessible
     # parameters by using an attr_accessible or attr_protected declaration.
-    config.active_record.whitelist_attributes = true
+    # config.active_record.whitelist_attributes = true
```
dougmartin, line 61: Maybe add a comment here about why you commented this out?

dougmartin (approved):
Looks great! :+1:

config/application.rb
emcelroy: I removed that line and the comment above it in the latest commit. When uncommented, it would cause this warning:

"DEPRECATION WARNING: config.active_record.whitelist_attributes is deprecated and have [sic] no effect. Remove its call from the configuration."

We originally left it in because we weren't 100% convinced that is true, but at this point I think it's safe to remove.

PR #1186, opened 2024-10-23
feat: Updated library interactive index page [PT-188467531]

dougmartin:
1. Sorted the list alphabetically
2. Made the url to the interactive a link
3. Added `CUSTOMIZABLE` on the same line as `OFFICIAL` if the library interactive has the "Allow author to change the following options" checkbox checked in the "Possibly Authorable Options" section.


emcelroy (approved):
Looks good 👍

PR #1187, opened 2024-10-23
feat: Display Library Interactive name in authoring [PT-#180951907]

dougmartin:
This adds the library interactive name and id to the authoring interface.


emcelroy (approved):
Looks good 👍

PR #1196, opened 2024-10-31
chore: upgrade from rails 4.1.16 to 4.2.11.23 (PT-186836210)

lublagg:


scytacki (commented):
I only looked at the environment change. It looks good to me.

app/models/c_rater/score_mapping.rb
```diff
@@ -60,4 +60,10 @@ def self.visible(user)
     self.all
   end
 
+  def self.create_mapping(mapping)
```
dougmartin, line 63: Is this new method needed?

dougmartin (approved):
Looks good :+1: 

Just one question about the need for the new `create_mapping` method but it doesn't block the merge.

PR #1198, opened 2024-11-05
chore: upgrade to rails 5.1.7 (PT-186836220)

emcelroy:
[#186836220](https://www.pivotaltracker.com/story/show/186836220)


Gemfile
```diff
@@ -118,8 +117,8 @@ gem "responders", "~> 2.0"
 gem "ribbons-rails", "~> 0.0", {git: "https://github.com/concord-consortium/ribbons-rails.git"}
 gem "safe_yaml", "~> 1.0.5"
 gem "spreadsheet", "~> 1.1.2", "= 1.1.2"
-# # gem "sqlite3", "~> 1.3"
-# # gem "test-unit", "~> 3.0"
+# gem "sqlite3", "~> 1.3"
```
dougmartin, line 120: Maybe since this is the last commit for a while could you just remove the commented out gems that aren't needed or group them together with a comment they are not needed?  That should help reduce confusion for the next person that touches this.

dougmartin (approved):
Looks good :+1: 

I skipped the review of all the migration files though since they all just added the 5.1 tag.

Gemfile
emcelroy: Good idea. We'll do this in the next PR for the 5.2 upgrade.

PR #1200, opened 2024-11-12
fix: Page settings checkboxes are not saving (PT-188545861)

emcelroy:
[#188545861](https://www.pivotaltracker.com/story/show/188545861)

Page settings parameters are camel case. The params whitelist needed to be modified to reflect that.


dougmartin (approved):
Looks good :+1:

PR #1201, opened 2024-11-12
fix: Unable to update/view TE Window Shades (PT-188545198)

emcelroy:
[#188545198](https://www.pivotaltracker.com/story/show/188545198)

Adds params required by the Teacher Edition window shades to the whitelist defined in `data_update_params` which is used by `update_page_item`. That allows window shade changes to be saved. It also seems to allow the window shades to appear correctly in the Activity Player Teacher Edition preview, though I'm not entirely sure why that is. Maybe the window shades don't show up until the default state/content has been edited... But, since I've had no problems viewing the window shades in Activity Player since making this change, I'm pretty sure fixing editing fixed that part of the bug as well.


dougmartin (approved):
Looks good :+1:

PR #1203, opened 2024-11-13
fix: Glossary image uploads [PT-188545704]

dougmartin:
This allows all the params to be sent in the body of the JWT request to the portal.


emcelroy (approved):
Looks good 👍

PR #1206, opened 2025-01-16
188545797 Add data-testid Attributes to Support Automated Testing

nstclair-cc:
[PT-#188775242](https://www.pivotaltracker.com/story/show/188775242) & [PT-#188545797](https://www.pivotaltracker.com/story/show/188545797)

### Description:
This PR introduces `data-testid` attributes in LARA to enhance automated testing reliability. These attributes simplify test creation and maintenance.

### Changes Made:
1. Added data-testid attributes for:
	•	Page Settings Modal:
	•	data-testid="page-title-input"
	•	data-testid="save-close-button"
	•	data-testid="cancel-button"
	•	Section-Level Actions:
	•	data-testid="section-collapse-button"
	•	data-testid="section-edit-button"
	•	data-testid="section-move-button"
	•	data-testid="section-copy-button"
	•	data-testid="section-delete-button"
2. Updated affected components to pass the attributes properly.

### Context:
These changes were made to support new automated tests added in [Cypress Tests [PR 17](https://github.com/concord-consortium/lara-cypress-tests/pull/17)].

PR #1207, opened 2025-01-21
adds back line break

nstclair-cc:
There's a line break issue I introduced in [PR 1206](https://github.com/concord-consortium/lara/pull/1206) that I think created issues in the header of LARA Staging. You can compare them in the screenshots here:

---

Screenshot 2025-01-20 at 8 07 44 PM
LARA Staging

Screenshot 2025-01-20 at 8 07 37 PM
LARA Production (how we want it to look)

---

The goal of this PR is to get the formatting back to where we wanted it to be. My guess is it was introduced at line 36 of `home.html.haml` since in Github this line is formatted partly in gray in the original PR: https://github.com/concord-consortium/lara/pull/1206/files#diff-ea95b1e97591bcd6c5713e60f07ff1cea51d305f84e34e08ed8dd6500fef84bb

Marking @dougmartin as reviewer in case I missed something obvious.

### DIFF
--- a/app/views/home/home.html.haml
+++ b/app/views/home/home.html.haml
@@ -9,37 +9,45 @@
     .buttons-menu
       - if can? :create, LightweightActivity
         %a{ :href => new_activity_path }
-          %button{ "data-testid" => "new-activity-button" }= t("NEW_ACTIVITY_BUTTON")
+          %button{ "data-testid" => "new-activity-button" }=t("NEW_ACTIVITY_BUTTON")
+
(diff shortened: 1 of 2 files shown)


dougmartin:
@nstclair-cc The issue wasn't with extra spaces but rather because a div was accidentally indented when you added the data-testid attributes.  I've fixed it with a new PR and added you are the reviewer:

https://github.com/concord-consortium/lara/pull/1208

I'm closing this PR.

PR #1208, opened 2025-01-21
fix: Mis-indented div in sticky header [PT-188545797]

dougmartin:
This fixes a div that was accidentally indented in home.html.haml as part of work to add data-testid attributes.


emcelroy (approved):
👍

nstclair-cc:
LGTM!

PR #1209, opened 2025-01-21
fix: Clean up new activity footer [PT-188793789]

dougmartin:
- Removes the unneeded "Lightweightactivity (new)" text in the footer.  This was broken with the LARA upgrade and it is original impetus for this story.
- Updates the save/cancel button classes to match all the other "new" forms (sequences, glossaries, etc) and adds padding around the buttons to match the other forms style.

---

New footer (from local build):

image


emcelroy (approved):
Looks good 👍

PR #1210, opened 2025-01-21
fix: Lint issues with data-testid changes [PT-188545797]

dougmartin:
The linter didn't like unquoted object keys after the quoted "data-testid" key was added.

Also adds the updated built lara-typescript.js


dougmartin:
@nstclair-cc You can ignore the generated `lara-typescript.js` file when reviewing - that is what is built by the `npm run build` script.

nstclair-cc (approved):
LGTM! Thanks for making those fixes, Doug!

PR #1214, opened 2025-02-03
added data-testid tags for updating user permissions

nstclair-cc:
[PT-#188754684](https://www.pivotaltracker.com/story/show/188754684)

This PR adds `data-testid` tags to LARA admin pages so it’s easier to test and track changes to user accounts.

### DIFF
--- a/app/views/admin/users/index.html.erb
+++ b/app/views/admin/users/index.html.erb
@@ -1,12 +1,12 @@
-<h1>Listing Users</h1>
+<h1 data-testid="listing-users">Listing Users</h1>
 
 <div class="users">
   <%= will_paginate @users %>
 
   <form class="search" method="get">
     <lable>Search</lable>
-    <input type="text" id="q" name="q" placeholder="user@host"/>
-    <input type="submit" value="submit"/>
+    <input type="text" id="q" name="q" placeholder="user@host" data-testid="search-users"/>
+    <input type="submit" value="submit" data-testid="search-submit"/>
   </form>
 
   <table>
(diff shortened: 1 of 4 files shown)


dougmartin (approved):
Looks good :+1:

app/views/admin/users/index.html.erb
```diff
@@ -17,8 +17,8 @@
     <% @users.each do |user| %>
       <tr>
         <td><%= user.email %></td>
-        <td><%= link_to 'Edit', edit_admin_user_path(user) %></td>
-        <td><%= link_to 'Destroy', admin_user_path(user), method: :delete, data: { confirm: 'Are you sure?' } %></td>
+        <td><%= link_to 'Edit', edit_admin_user_path(user), data: { testid: "edit-user-#{user.id}" } %></td>
+        <td><%= link_to 'Destroy', admin_user_path(user), method: :delete, data: { confirm: 'Are you sure?', testid: "destroy-user-#{user.id}" },  %></td>
```
emcelroy, line 21: Looks like there's an errant comma here, right before the closing `%>`?

app/views/admin/users/index.html.erb
```diff
@@ -1,12 +1,12 @@
-<h1>Listing Users</h1>
+<h1 data-testid="listing-users">Listing Users</h1>
 
 <div class="users">
   <%= will_paginate @users %>
 
   <form class="search" method="get">
     <lable>Search</lable>
```
emcelroy, line 7: I know you didn't write this line, but I just noticed this misspelled "lable" tag :) Maybe fix that while you're working on this?

emcelroy (approved):
Looking good 👍 I think there's an unnecessary comma on line 21 of `app/views/admin/users/index.html.erb` that caused a test failure. And then there's an old misspelling I happened to notice which should probably be corrected while you're working on this. Otherwise I think it's all set.

app/views/admin/users/index.html.erb
nstclair-cc: Good catch! Thanks, I removed the trailing comma.

PR #1217, opened 2025-02-04
Remove code related to publishing to LARA runtime.

lublagg:
Note: I wasn't able to really figure out what else in the code was related to publishing to LARA runtime. I looked over the `publication` model and controller and didn't see any explicit references to LARA runtime.

If there are places in the code where it would make sense to take a look, let me know.

### DIFF
--- a/app/helpers/lightweight_activity_helper.rb
+++ b/app/helpers/lightweight_activity_helper.rb
@@ -75,13 +75,6 @@ module LightweightActivityHelper
 
     activity_player_url = activity.activity_player_url(base_url, page: page)
     activity_player_te_url = activity.activity_player_url(base_url, page: page, mode: "teacher-edition")
-    if page
-      lara_runtime_url = preview_activity_page_path(activity, page)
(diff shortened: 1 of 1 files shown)


dougmartin (commented):
Is this just the start of the runtime removal work?

The runtime removal task is going to be big - it means going into every controller and finding the methods that were used in the LARA runtime (but are now handled by AP) and removing them, along with their views (and the views and partials they call) and the routes to them.  

Here is an example for the old single page runtime method:

- controller action needs to be removed: https://github.com/concord-consortium/lara/blob/master/app/controllers/lightweight_activities_controller.rb#L49-L78
- route needs to be removed: https://github.com/concord-consortium/lara/blob/master/config/routes.rb#L281
- all partials used by the view (the the partials within partials, etc...) needs to be removed: https://github.com/concord-consortium/lara/blob/master/app/views/lightweight_activities/single_page.html.haml
- view needs to be removed: https://github.com/concord-consortium/lara/blob/master/app/views/lightweight_activities/single_page.html.haml

PR #1219, opened 2025-02-05
fix: Missing allowed lightweight activity params [PT-188860583]

dougmartin:
The following params were not allowed for lightweight activities so were not being saved:

- fixed_width_layout
- is_official
- publication_status


lublagg (approved):
👍

emcelroy (approved):
Looks good 👍

PR #1221, opened 2025-02-05
chore: remove show in featured question report [Lara-4]

emcelroy:
This is for Task 13  in the [LARA-4 story](https://concord-consortium.atlassian.net/browse/LARA-4).

The "Show in featured question report" option was only used by the old dashboard app which is no longer in use and not supported. There is no plan to use the option in the new dashboard. So we are removing it.


dougmartin (commented):
Why the huge change in `package-lock.json`?  Did you change node versions which changed package versions?  I think that big of a change should be in its own PR.

emcelroy:
> Why the huge change in `package-lock.json`? Did you change node versions which changed package versions? I think that big of a change should be in its own PR.

Thanks for catching that. It was not intended and I've removed those changes.

dougmartin (approved):
Looks good :+1:

PR #1222, opened 2025-02-06
chore: remove embeddable views [LARA-4]

emcelroy:
This is for Task 14  in the [LARA-4 story](https://concord-consortium.atlassian.net/browse/LARA-4).

Removes all embeddable views for question types that have been replaced by question-interactives. Note: I attempted to update the Cypress tests in light of these changes, but couldn't get Cypress to run in my Codespaces instance of LARA. I think it's not worth spending more time on that though since we'll be adding newer Cypress tests.


dougmartin (approved):
Looks great :+1: 

I'm curious why you didn't also remove the controllers?

btw, in case @knowuh still gets email notifications - this is your kind of PR Noah :)

emcelroy:
> I'm curious why you didn't also remove the controllers?

@dougmartin The task said to remove the views, but maybe I'm taking that too literally? I thought we weren't so sure the controllers aren't used at all anymore. But if it makes sense to remove those too, I can do that.

dougmartin:
@emcelroy I think this is enough for this PR.  I think we are out of time for this work anyway right?

emcelroy:
> @emcelroy I think this is enough for this PR. I think we are out of time for this work anyway right?

@dougmartin Yeah, we're out of time. I thought removing the controllers might be relatively quick, but I'll skip it and just merge this as is. Thanks.

PR #1223, opened 2025-02-06
Remove support for score mappings + CRater.

lublagg:
This PR removes models, views, controllers, routes, scripts, and stylesheets related to CRater/Score Mappings.
Activities that use auto-scoring should've been updated to use the [Scorebot Library Interactive](https://github.com/concord-consortium/question-interactives/tree/e46807862a96d7532b285cca5974cd938e70ca07/packages/score-bot).


db/migrate/20250206202956_remove_show_in_featured_question_report_from_managed_interactive.rb
```diff
@@ -0,0 +1,8 @@
+class RemoveShowInFeaturedQuestionReportFromManagedInteractive < ActiveRecord::Migration[5.2]
+  def up
+    remove_column :managed_interactives, :show_in_featured_question_report
+  end
+  def down
+    add_column :managed_interactives, :show_in_featured_question_report, :boolean, default: true
+  end
+end
```
emcelroy, line 8: This was added because I missed it in the ["remove show in featured question report" PR](https://github.com/concord-consortium/lara/pull/1221). It was causing Lucy problems with this work, so we added it here. If we really should break it out into a separate PR, though, let me know.

emcelroy (approved):
This looks good to me 👍. I would wait to see what Doug and/or Scott have to say about it before merging, though.

dougmartin (approved):
This looks good to me :+1: 

I think @scytacki should also review though as I'm not too familiar with the arg blocks/score mappings.

scytacki (approved):
Looks good to me. 👍

PR #1224, opened 2025-02-06
Remove Question Trackers code.

lublagg:
Removes Question Trackers + Tracked Questions code.


emcelroy (approved):
This looks good to me 👍 But it's another one where Doug and/or Scott should weigh in before merging, of course.

dougmartin (approved):
Looks good :+1: 

Again, @scytacki should also take a look as I'm not that familiar with question trackers.

scytacki (approved):
Looks good. Nice to get rid of so much code. 👍

PR #1226, opened 2025-02-10
build: Change Docker base image to ruby:X [LARA-5]

dougmartin:
This changes the base image to no longer use the private base image and instead lifts the RailsLTS password into the LARA configuration.

The reason for this is the number of different Ruby upgrades that are required coupled with the required updates to the gems to support that Ruby version.  This will be easier to upgrade using a direct base image.


emcelroy (approved):
Looks good -- thanks!

dougmartin:
@emcelroy / @lublagg One thing I didn't explicitly call out in the upgrade notes is that each time you change the `FROM` image in the docker files you'll need to run `docker compose build` to re-build your development image before you start any other upgrade steps.

PR #1227, opened 2025-02-10
Upgrade to Rails 6.0.6.1

lublagg:


config/environments/production.rb
```diff
@@ -34,7 +34,7 @@
   # config.action_dispatch.x_sendfile_header = 'X-Accel-Redirect' # for nginx
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  # config.force_ssl = true
+  config.force_ssl = true
```
dougmartin, line 37: I'm wondering if this might be a problem when we deploy to AWS.  We don't do it in the portal.  We'll see this morning.

dougmartin (approved):
Looks good :+1: 

When I deploy this to staging we'll see if we need to remove the `force_ssl` configuration.

PR #1229, opened 2025-02-12
fix: Disable production force_ssl setting

dougmartin:


dougmartin:
@emcelroy / @lublagg My spider-sense was right - the initial 6.1 staging deployment wouldn't finish because the health checks didn't finish with this option enabled because it couldn't route requests to the container.  This fix is already working on staging at `v2.15.0-rails.6.1.7.10-1`.

emcelroy (approved):
Looks good 👍 Thanks for fixing that.

PR #1231, opened 2025-02-12
fix: publish options missing [LARA-149]

emcelroy:
[LARA-149](https://concord-consortium.atlassian.net/browse/LARA-149)

The "Publish to Other Portals" component previously existed only in the `itsi-authoring` CoffeeScript code. So we lost that when we removed the ITSI Authoring elements. This PR reinstates the related files, although we've placed them in the main `/app/assets/javascripts/components/` directory instead of reinstating the `itsi-authoring` sub-directory. 

While fixing that, we found that the definition of `gone.publication_details` for activities needs to be made in `LightweightActivitiesController` instead of `InteractivePagesController`. Otherwise the "Publish to Other Portals" element wouldn't render for activities. We're not 100% sure, but suspect because the app was always using the ITSI Authoring component, and there is a fallback `gon.ITSIEditor.publication_details` [here in the main branch](https://github.com/concord-consortium/lara/blob/ecf75c1e89dcd23885756826b0c4fb3ec6e4de6d/app/views/publications/_publication_details.html.haml#L4), that issue was being masked.

[LARA-149]: https://concord-consortium.atlassian.net/browse/LARA-149?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/assets/javascripts/components/publication_ajax_mixin.js.coffee
+++ b/app/assets/javascripts/components/publication_ajax_mixin.js.coffee
@@ -1,0 +1,22 @@
+modulejs.define 'components/publication_ajax_mixin', ->
+
+  getInitialState: ->
+    last_publication_hash: @props.publicationDetails.last_publication_hash
+    latest_publication_portals: @props.publicationDetails.latest_publication_portals
+    polling: false
+
+  componentDidMount: ->
+    @poller = setInterval @pollForChanges, 5000
+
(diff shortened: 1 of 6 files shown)


dougmartin (approved):
Looks good :+1:

PR #1232, opened 2025-02-13
fix: plugin wrappers not saving [LARA-153]

emcelroy:
[LARA-153](https://concord-consortium.atlassian.net/browse/LARA-153)


dougmartin (approved):
Looks good :+1:

PR #1233, opened 2025-02-13
chore: Upgrade to Rails 7.1.5 [LARA-9]

lublagg:
[LARA-9](https://concord-consortium.atlassian.net/browse/LARA-9)

In addition to upgrading to Rails 7.1, there are other significant changes here due to problems with deprecated gems. We ended up moving Ruby all the way up to 3.3.7 while working through related issues. Since the app seems to be working fine, we propose to keep that change even though it may not be required.

We removed `compass-rails` and `compass-blueprint`, and removed or replaced all related code. We also replaced `uglifier` with `terser`.

[LARA-9]: https://concord-consortium.atlassian.net/browse/LARA-9?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Dockerfile
```diff
@@ -19,11 +19,13 @@ WORKDIR $APP_HOME
...
-ENV BUNDLE_GEMFILE=$APP_HOME/Gemfile
-
-# need to copy the Gemfile.lock created during build so it isn't overriden by the following add
-RUN bundle install --without development test && \
+# Determine Bundler version and install it, then copy the Gemfile.lock created during build so
+# it isn't overridden by the following add
+RUN BUNDLER_VERSION=$(grep -A1 "BUNDLED WITH" Gemfile.lock | tail -n1) && \
+    gem install bundler -v "$BUNDLER_VERSION" && \
+    bundle _"$BUNDLER_VERSION"_ install --without development test && \
     cp Gemfile.lock Gemfile.lock-docker
```
dougmartin, line 27: Do you need this copied file?  I don't see it referenced anywhere else.

Gemfile
```diff
@@ -60,10 +58,10 @@ group :test, :development do
 end
 
 gem "acts_as_list"
-gem 'aws-sdk', '= 2.0.22'
+gem 'aws-sdk', '~> 3'
```
dougmartin, line 61: I'm wondering if the major version release is going to break any editor code, like the rubric saving.  We'll see on staging.

dougmartin (approved):
Looks good :+1: 

I'll get this on staging and test out the aws s3 uploads for the rubrics.

Gemfile
dougmartin: No problems saving rubrics on staging!  Nice job.

PR #1235, opened 2025-02-14
chore: upgrade ro Rails 8.0 [LARA-101]

emcelroy:
[LARA-101](https://concord-consortium.atlassian.net/browse/LARA-101)


config/initializers/assets.rb
```diff
@@ -0,0 +1,6 @@
+Rails.application.config.assets.paths << Rails.root.join("vendor", "assets", "javascripts")
+Rails.application.config.assets.paths << Gem.loaded_specs["chosen-rails"].full_gem_path + "/vendor/assets/javascripts"
+Rails.application.config.assets.precompile += %w(chosen-jquery.js)
+Rails.application.config.assets.paths << Rails.root.join("vendor", "assets", "stylesheets")
+Rails.application.config.assets.paths << Gem.loaded_specs["chosen-rails"].full_gem_path + "/vendor/assets/stylesheets"
+Rails.application.config.assets.precompile += %w(chosen.css)
```
emcelroy, line 6: After the upgrade, Rails couldn't seem to find these `chosen` assets for reasons I could not figure out. This was the only way I was able to get them added correctly. It seems like overkill adding this file just for those assets, but I'm not sure what else to do.

dougmartin (approved):
Looks good :+1:

PR #1237, opened 2025-02-21
fix: plugin wrappers are not saving [LARA-153]

emcelroy:
[[LARA-153](https://concord-consortium.atlassian.net/browse/LARA-153)]

I'm not proud of these changes but they fix the problem, which is that Teacher Edition question wrappers were not editable. In `update_page_item` there are issues with the different types of embeddables and the fields they support, plus how we're using the `page_item` params to update both the page item and the embeddable that make things more complex than I'd like. I'm not sure how to handle it otherwise, though. Happy to hear any thoughts on how to improve these changes.

Maybe a small improvement would be a separate `update_page_item_embeddable` method that could be called by `update_page_item` and we move most of the embeddable-specific code there? That might at least make the pieces more comprehensible.

[LARA-153]: https://concord-consortium.atlassian.net/browse/LARA-153?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/api/v1/interactive_pages_controller.rb
+++ b/app/controllers/api/v1/interactive_pages_controller.rb
@@ -292,14 +292,12 @@ class Api::V1::InteractivePagesController < Api::ApiController
     render json: result.to_json
   end
 
-  def update_page_params
-    params.require(:page_item).permit(
-      :section, :position, :embeddable, :column, :data
-    )
+  def page_item_update_params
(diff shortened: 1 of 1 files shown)


dougmartin (approved):
Looks good :+1:

PR #1239, opened 2025-03-14
[LARA 175] added a log in and log out spec

nstclair-cc:
### Description:
This PR adds a working Cypress spec that successfully logs in and out of LARA in the staging environment. The code is currently in draft mode with a couple of open questions for @dougmartin and some potential enhancements:

---

#### ✅ Summary of Changes:
- Added a login/logout test for LARA Staging.
- Configured portalBaseUrl for staging environment.

---

#### 1. Commands.js:
We could move the login/logout logic into the `commands.js` file for cleaner test code.
- ➡️ The tradeoff I see is that if we switch to Playwright in the future, refactoring custom commands may be more complex.
#### 2. Secure Password Storage:
Do we need to worry about securing the password in the dev environment for LARA? Currently, it’s passed via CYPRESS_password at runtime, which seems fine for staging but might need tightening up for dev.
#### 3. Configuration in Dev Environment:
In the current setup, I added:

> `portalBaseUrl: "https://learn.portal.staging.concord.org"`

Where would this live in the dev environment? Is there an existing config pattern we should follow?

#### 4. Switching from `data-cy` to `data-testid`:
There are several data-cy selectors currently used in the tests.
✅ Playwright works more reliably with `data-testid`, and it’s also more secure for production environments.
✅ To migrate, I’d need to update the tags in staging, which is a bit annoying but necessary.
💡 Should I create a separate GitHub branch just for migrating the tags? That way, we can isolate that change from the main spec work.

---

#### ✅ Next Steps:

- [x] Feedback from Doug on the questions above.

If the direction looks good, we can:
- [x] Refactor into commands.js
- [x] Set up secure password storage if needed
- [ ] Create a new branch for the data-testid migration

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,7 +47,7 @@ jobs:
           nix develop -c npm run test
         working-directory: ./lara-typescript/
       - name: Create CI admin user
-        run: nix develop -c bundle exec rake create_ci_admin_user
+        run: nix develop -c bundle exec rake lightweight:create_ci_admin_user
       - name: Start Rails server
         run: nix develop -c bundle exec rails server -b 0.0.0.0 -p 3000 &
       - name: Run Cypress Tests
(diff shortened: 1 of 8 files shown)


dougmartin (approved):
Looks great :+1:

PR #1241, opened 2025-04-06
build: Add RESTART_TOGGLE parameter to stack_template.yml

dougmartin:
This allows us to restart the containers without changing other parameter values.  This mimics the parameter in the portal stack.


emcelroy (approved):
Looks good — thanks👍

PR #1244, opened 2025-08-28
feat: Add search by id number [LARA-187]

dougmartin:
Adds functionality to the search box on the home page that allows searching by id number.

If a positive integer is entered, the search will return any activities, sequences, glossaries, or rubrics with that id number along with name/title matches of that number.

This also adds a line above the listings that shows what the current search term is.


emcelroy (approved):
Looks great 👍

PR #1245, opened 2025-08-28
fix: Clicking on the empty title-container div goes to a broken link [LARA-182]

dougmartin:
The title is empty when the user is not editing a resource, such as on the homepage.

Now, when the title is empty, the title-container div has no background color or cursor change on hover and it prevents any click actions.


emcelroy (approved):
LGTM 👍 Thanks for fixing that.

PR #1262, opened 2026-07-14
docs(LARA-219): add host-authoring guide for the Interactive API

scytacki:
## Summary

Adds a current, host-focused guide for the LARA Interactive API at
`lara-typescript/src/interactive-api-host/interactive-host-guide.md`.

**Why:** The existing public docs site
([concord-consortium.github.io/lara-interactive-api](https://concord-consortium.github.io/lara-interactive-api/docs/))
is stale — it references CoffeeScript and is written from the *interactive-author*
point of view. We want an accurate doc we can hand to external partners (e.g. a
publisher evaluating or building a host) or point an AI at.

## What the doc covers

- The iframe-phone / `postMessage` message model (with a diagram)
- The **minimal host** — just `initInteractive` / `interactiveState` / `height` — as a complete working example
- `initInteractive` modes (runtime / authoring / report / reportItem)
- Saving & loading student work, including the unload handshake
- The full client↔host message catalog
- Optional subsystems (attachments, pub/sub, jobs, focus) and where the
  `@concord-consortium/interactive-api-host` library helps vs. doing it yourself

It is framed so a host can be built **with or without** the helper library, and
points to `interactive-api-shared/types.ts` as the authoritative protocol spec and
Activity Player's `iframe-runtime.tsx` as the reference host implementation.

Jira: [LARA-219](https://concord-consortium.atlassian.net/browse/LARA-219)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[LARA-219]: https://concord-consortium.atlassian.net/browse/LARA-219?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lara-typescript/src/interactive-api-host/interactive-host-guide.md
+++ b/lara-typescript/src/interactive-api-host/interactive-host-guide.md
@@ -1,0 +1,239 @@
+# Building a Host for LARA Interactives
+
+This document describes how to build a **host** — an environment that embeds and
+runs LARA/AP "interactives." It is written for engineers integrating interactives
+into their own platform, **whether or not they use the
+`@concord-consortium/interactive-api-host` library.** The library is an optional
+convenience; the actual contract between a host and an interactive is a small
+`postMessage` protocol, and you can implement it directly.
+
+If instead you are *building an interactive* (the thing that runs inside the
+iframe), see the client library `@concord-consortium/lara-interactive-api`.
+
+---
+
+## 1. The mental model
+
+An **interactive** is a web page (its own URL) that a host loads inside an
+`<iframe>`. The host and the interactive never share a JavaScript context — they
+communicate only by `postMessage`. On top of raw `postMessage`, both sides use
+[**iframe-phone**](https://github.com/concord-consortium/iframe-phone), a tiny
+library that provides:
+
+- a **handshake** so each side knows the other is ready, and
+- a simple **named-message** API: `phone.post(type, payload)` and
+  `phone.addListener(type, handler)`.
+
+```
+┌─────────────────────── Host (your platform) ───────────────────────┐
+│                                                                     │
+│   iframe-phone ParentEndpoint                                       │
+│        │  post("initInteractive", …)      ── student config ──▶     │
+│        │  addListener("interactiveState") ◀── student work ───      │
+│        │  addListener("height")           ◀── iframe sizing ──      │
+│        ▼                                                            │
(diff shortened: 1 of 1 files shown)


lara-typescript/src/interactive-api-host/interactive-host-guide.md
```diff
@@ -0,0 +1,245 @@
...
+    error: null,
+    hostFeatures: {},                 // advertise optional capabilities here (see §6)
+    authoredState: savedAuthoredState, // how the interactive was configured by an author
+    interactiveState: savedStudentState, // previously saved student work, or null
+    globalInteractiveState: null,
+    interactiveStateUrl: "",
+    collaboratorUrls: null,
+    classInfoUrl: "",
+    interactive: { id: "interactive-1", name: "" },
+    authInfo: { provider: "", loggedIn: true, email: "" },
```
emcelroy: Would it be slightly better to use `loggedIn: false` or leave it true and add a value for `provider`?

lara-typescript/src/interactive-api-host/interactive-host-guide.md
```diff
@@ -0,0 +1,245 @@
...
+
+The core in §2 needs no library. These richer features involve real bookkeeping
+(request/response correlation, backend calls, listener maps), and that is exactly
+what `@concord-consortium/interactive-api-host` provides — you can adopt it
+piecemeal. Activity Player uses these managers while still driving the core loop
+by hand.
+
+| Feature | Do it yourself | Or use from the library |
+|---------|----------------|-------------------------|
+| **Capability advertising** | Set fields on `initInteractive.hostFeatures` (`IHostFeatures`) so the interactive knows what you support (e.g. `modal`, `getFirebaseJwt`). | — |
```
emcelroy, line 219: At first I thought the `getFirebaseJwt` on this line was a small typo since `getFirebaseJWT` appears earlier in the document. But it looks like they're actually two separate things. Claude suggested adding this footnote to avoid similar confusion:

```
**Note the casing:** the message is `getFirebaseJWT`, but the `IHostFeatures` field advertising support for it is `getFirebaseJwt` (as are the payload types `IGetFirebaseJwtRequest`/`Response`). This is a historical quirk of the API, not a typo — the identifiers genuinely differ.
```

emcelroy (approved):
Looks good 👍 I left a couple small suggestions but I don't think they're required if you prefer to merge as-is.

PR #1263, opened 2026-08-07
feat: label which column is primary in the section editor

dougmartin:
Closes [LARA-220](https://concord-consortium.atlassian.net/browse/LARA-220).

In a two-column section one column is "primary" and the other "secondary", and that distinction already drives real behavior: which column can stick to the top while a student scrolls, which one students can collapse, which one uses carousel mode, and the order question numbering follows. The section editor never said which was which, so authors had to infer it from the layout token. This adds the labels and makes the Layout dropdown self-explanatory. Labeling only: no data model change, no migration, no Activity Player change, and no change to how any published activity renders.

The rule the labels express: primary is on the left for `60-40` and `70-30`, and on the right for `40-60`, `30-70`, `responsive-30-70`, and `responsive-50-50`. Note `responsive-50-50` is the counterintuitive one, since the columns are equal width and the primary is simply the right-hand one.

## Please test the drive-by bug fix

**This PR also fixes a pre-existing defect that has nothing to do with labeling, and it is worth testing on its own.** `id="section_layout"` and `id="toggle-secondary-column"` were hardcoded, so every section on a page emitted the same two IDs. Each `<label>` carries `htmlFor`, which wins over wrapping, so on a page with more than one section:

- clicking the words "Allow student to hide secondary column" on any section below the first toggled **section 1's** checkbox and fired **section 1's** save, and
- every section's "Layout:" label pointed at the first section's dropdown.

Clicking the checkbox square itself always worked. Only the label text misrouted, which is why this survived so long. To verify: open a page with two or more sections and click the checkbox *label text* on the second section.

No separate Jira bug was filed. This matches what the repo already does with drive-by fixes, and a ticket opened and closed inside one PR has no lifecycle. Flagging it here instead so it gets tested deliberately rather than discovered inside a labeling story.

## What changed

Four commits, each independently reviewable and each leaving the app working:

1. `refactor` extracts the layout derivations into `util/section-layout-utils.ts`. `columnValueForIndex` was a closure inside `AuthoringSection` capturing `layout` from state, so it was neither importable nor unit-testable. Pure refactor, nothing renders differently, so its behavioral equivalence can be checked on its own before the visible changes land on top.
2. `fix` makes the section header's element IDs unique per section (the defect above).
3. `feat` gives the Layout dropdown readable option text.
4. `feat` labels each column.

A fifth commit adds a regression test found during final review, and the last commit closes the spec into a single summary at `specs/LARA-220-label-which-column-is-primary-in-authoring.md`.

## Decisions worth knowing while reviewing

- **The dropdown keeps the raw token** (`60-40 (primary on left)`, not `60/40`). The tokens are the stored values and appear in the help doc and in years of support conversations, so an author searching for "60-40" should still find a match. Each option's `value` is byte-identical to before, so nothing about what gets saved changes.
- **No label is shown for a layout outside the enum**, even though two columns render. LARA's own Ruby export and the Activity Player disagree about which column is primary for such a layout, so a label there risks saying the opposite of what students see. Silence is the honest answer. A census of both deployed environments found zero such rows (production 281,699 sections, staging 5,045, both with 0 NULL and 0 unrecognized), so this is defensive rather than load-bearing.
- **The header is a `<div>`, not a `<header>` and not a heading.** A `<header>` becomes a `banner` landmark unless it descends from `article`, `aside`, `main`, `nav`, or `section`, and `role="group"` on the parent does not exempt it. `<h4>` would sit at the same outline level as the section item titles nested inside the column, and `<h3>` collides with the section name.
- **The header ID is unique per column, not per section.** A per-section ID resolves both columns' `aria-labelledby` to the first header, so no column gets announced as Secondary.
- **`grid-template-rows: min-content 1fr` is load-bearing**, scoped to columns that render a header. Without it the shorter column's header inflates and pushes its content below the other column's. The obvious alternative, `align-content: start`, realigns the headers but shrinks that column's drop target, because today the `.edit-items-container` stretch is what makes a column's whole empty area a valid drop zone.
- The two dead `data-testid` props on `SectionColumn` are dropped and nothing replaces them. They never reached the DOM, since the component destructures only its declared props. Tests identify columns by accessible role and name instead.

## Verification

- **81 tests across 5 suites** in `section-authoring`; **358 tests, 43 suites** for the whole `lara-typescript` package. `lint` and `lint:build` clean. `lint:unused` reports 91 errors, unchanged from master, so this introduces none.
- Every new test was confirmed by negative control: reverting each fix fails exactly the tests that should fail. A per-section header ID fails 10; removing the `hasColumnHeader` class fails exactly 1.
- **Verified in Chromium against the running app**, not just jsdom. Computed style is `uppercase`, 12px, weight 700, `rgb(1, 96, 130)`, which is `--dark-teal` on the `--ltr-teal` column background at 5.59:1, passing AA. The header measures 25px and stays on one line in every layout including the tightest, the 305px static column of `responsive-30-70`. Both drop zones start at the same offset with identical heights. The accessibility tree shows `group "Primary column"` and `group "Secondary column"` with `Add Item` inside, and exactly one `banner` in the section, the pre-existing `sectionMenu`, so the column headers added no landmark.
- **Drag and drop still works with the header present**, checked with react-beautiful-dnd's keyboard sensor: lift and cancel both announce correctly and the drop-zone offset holds steady during the drag. Note that Playwright's `dragTo` cannot reorder an rbd list at all, with or without this change, so use the keyboard sensor or real mouse steps for any browser-level drag check.

### Known cosmetic consequence

The wider dropdown moves the section header's wrap threshold from roughly 1140px to roughly 1210px. In that band the header grows from 40px to 57px: the checkbox label wraps and DELETE drops onto its own line. Measured at 1200px, nothing is clipped and there is no page-level horizontal scroll. Above roughly 1220px there is no difference, and below roughly 1140px the header already wraps today. Accepted rather than filed.

## Out of scope

Letting authors choose which column is primary (the other half of the original request) needs a new section field, a migration, authoring API and export changes, and Activity Player changes in about half a dozen places. Also deferred: any in-editor explanation of what primary/secondary controls, since every honest explanation is conditional and AP-129 is scheduled to change the sticky rule; the authoring help-doc note, which should cover only the AP-129-independent facts for now; and Storybook work of any kind, including the pre-existing swapped single-column stories.

[LARA-220]: https://concord-consortium.atlassian.net/browse/LARA-220?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


scytacki (approved):
looks good to me

